PR #2291, opened 2018-12-04
Add Travis CI

beutlich:
Build (of both pdf and html) takes about 6 min at Travis CI.

Took me some [iterations](https://travis-ci.org/beutlich/ModelicaSpecification/builds) to get both builds running. TexLive 2013 (which comes with trusty) is not suitable since it misses the `\textsubscript` command. Thus, at least TexLive 2015 is required (if TeX sources must not be touched).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,27 @@
+language: generic
+
+git:
+  depth: 5
+
+cache:
+  directories:
+    - /tmp/texlive
+
+before_install:
+  # Thanks to http://stackoverflow.com/a/32358866
+  # Install modules into ~/perl5 using system perl
+  - curl -L https://cpanmin.us | perl - App::cpanminus
+  - export PATH=$PATH:~/perl5/bin/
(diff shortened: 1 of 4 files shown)


sjoelund:
The one negative part I can see about this setup is it may not reproduce the same result if run on the same commit at a later date due to installing new packages from a git, etc. Better would be to prepare a docker image with a fixed version of everything installed, tag it and use that. Later on if a newer image is needed, tag another one :D

This would possibly be an issue for a maintenance branch of the spec if it no longer gives the same result when a typo is to be fixed.

beutlich:
Yes, I tried with docker images, too. But had no success for whatever reason. I'll leave that to the webmasters :wink:

sjoelund:
I added #2315, which uses a custom docker image (Dockerfile in the repo) and builds on (OpenModelica) Jenkins server and uploads to [specification.modelica.org](https://specification.modelica.org/master/MLS.html) (and the PDF as well). The docker image would need to be updated if we want a newer version of Hans latexml repository...
When something is tagged it would also be uploaded.

PR #2296, opened 2018-12-06
This fixes the ugly table from section 7.1.3

dietmarw:
The (probably auto generated) table looked awful. This PR converts it from:
image

to 

image

Tested on PDF and HTML output.


dietmarw:
I just noticed that the original LaTeX version with the ugly table was even wrong when one compares the grey cells. The new version is now in sync with the latest official PDF.

PR #2351, opened 2019-05-08
Formatting inline code

GallLeo:
As discussed in #2284, MA Backoffice went through the document, and added `\lstinline!...!`, where previously "Courier New" had been used in the Microsoft Word document.

- Some LaTeX comments `%TODO-FORMAT ` indicate a few open questions/todos which could not be dealt with during inline formatting

- One could use `\begin{nonnormative}` in more places. But this would be another lengthy walk through the document, because already from Word, the `\emph{}` sections had been interrupted in many cases. (-> Next step)

- Syntax highlighting is still inconsistent in many places. But, at least we should have the same status/looks like in version 3.4, again.

- If needed, further checking and smaller fixes could be done on master

- How can I squash my commits from my fork into one commit in master?


sjoelund:
> How can I squash my commits from my fork into one commit in master?

It seems you named your branch `master` instead of something else. That complicates things slightly for you; I assume you have the MA remote as `origin` and your fork as `leo`:
```bash
git fetch leo
git checkout leo/master
git checkout -b mysquash
git fetch origin
git merge --squash origin/master
git commit # Choose your commit message based on the other ones
git checkout master
git reset --hard mysquash
git push --force
```
You can also squash things in the github UI as far as I know. You can also do it in fewer lines, but with the above you have a smaller risk of messing things up:
```bash
git checkout master
git fetch origin
git merge --squash origin/master
git commit # Choose your commit message based on the other ones
git push --force -u leo
```

dietmarw:
There is no need to do any complicated squash merge on your side Leo. GitHub gives us the option to do a "Squash Merge" instead of a "Merge". We do that all the time (when appropriate) for MSL.

image

dietmarw:
@HansOlsson so if you approve now I can do that.

chapters/dae.tex
```diff
@@ -51,7 +51,7 @@ \chapter{Modelica DAE Representation}\doublelabel{modelica-dae-representation}
 \textbf{m}(t\textsubscript{e}) & Modelica variables of type
 \emph{discrete Real, Boolean, Integer} which are unknown. These
 variables change their value only at event instants t\textsubscript{e}.
-pre(m) are the values of m immediately before the current event
+\lstinline!pre!(m) are the values of m immediately before the current event
```
dietmarw: I think the (m) should be part of the listinline here.

HansOlsson: The old document had exactly the formatting above, i.e. pre as code - but (m) and m as normal text, so I think we can leave it as is. Either it was a pure mistake, or the intention was that m is a variable in the pseudo-code and not in Modelica.

GallLeo: Thank you for spotting it. It looked strange, so I fixed it.

There are more "taste" decisions and the LaTeX document isn't a one-to-one copy. It just should look similar.

GallLeo:
> There is no need to do any complicated squash merge on your side Leo. GitHub gives us the option to do a "Squash Merge" instead of a "Merge". We do that all the time (when appropriate) for MSL.

Thank you for the hint, that is what I had in mind. I think it's great if the receiver can decide whether to squash or not. As we sometimes went back and forth during editing, added todos and removed them again ... I think in this case it makes much sense to squash.

chapters/derivationofstream.tex
```diff
@@ -230,7 +230,7 @@ \subsection{Connection of 3 stream connectors where two mass flow rates are posi
 inStream(h_{outflow,1})=\frac{\text{max}(-\dot{m}_2,0)h_{outflow,2}+\text{max}(-\dot{m}_3,0)h_{outflow,3}}{\text{max}(-\dot{m}_2,0)+\text{max}(-\dot{m}_3,0)}=\frac{0}{0}
 \end{equation*}
 
-The inStream() operator cannot be evaluated for a connector, on which
+The \lstinline!inStream!() operator cannot be evaluated for a connector, on which
```
dietmarw: one more parenthesis :-)

chapters/derivationofstream.tex
```diff
@@ -195,7 +195,7 @@ \subsection{Connection of 3 stream connectors where one mass flow rate is identi
 properties discussed for two connected components still hold. The
 connection set equations reflect that the sensor does not any influence
 by discarding the flow rate of the latter. In several cases a non-linear
-equation system is removed by this transformation. However, inStream(..)
+equation system is removed by this transformation. However, \lstinline!inStream!(..)
```
dietmarw: and here

dietmarw: Strange with the two dots though

HansOlsson: Also issues in the original word-document. (Including the two dots). 

There are about 44 cases of (..) in Modelica 3.4. 
In some cases it may be intentional since fixed-width dots make `(...)` look too wide - however, that can be adjusted to use proper ldots.

GallLeo: I'm open to fix all these, but before doing so we would need a guideline from MAP-LANG, how the specification should look like. 

In many cases we did intentionally do a "stupid robot job", because there was no consistent way to be seen, so we "copied" the word document.

GallLeo:
It would be interesting to have a look at the generated [html](https://specification.modelica.org/master/MLS.html). Last build is from April 23. How is this triggered?

chapters/derivationofstream.tex
GallLeo: Thank you for spotting. Fixed.

chapters/derivationofstream.tex
GallLeo: Thanking you for spotting. Fixed.

Regarding consistency and general looks: I did some minor improvements, but main task was to get the LaTeX document up to the state of Word. See my initial description of the pull request: "Syntax highlighting is still inconsistent in many places. But, at least we should have the same status/looks like in version 3.4, again."

HansOlsson:
> It would be interesting to have a look at the generated html. Last build is from April 23. How is this triggered?

Automatically when main document is updated. I don't know if it is possible to do for branches/pull-requests in a simple way.

dietmarw:
It should get build now, that it is merged into master.

sjoelund:
For future reference, the generated PDF and html were available here: https://test.openmodelica.org/jenkins/job/ModelicaAssociation/job/ModelicaSpecification/view/change-requests/job/PR-2351/

dietmarw:
@sjoelund Could you add a small link to either README or Wiki pages?

PR #2354, opened 2019-05-13
Changed = to := in algorithms

HildingElmqvist:


GallLeo:
MA Backoffice now filed Hilding's MA CLA, so merging should be fine.

PR #2360, opened 2019-05-15
Find files with trailing whitespace

sjoelund:
Note: This should not be possible to merge until all whitespace has been removed (which we should not do until #2353 is fixed).


sjoelund:
I'll close this and try to reopen because Jenkins seems to think there is no .CI/Jenkinsfile in the repo after this PR is merged...

PR #2361, opened 2019-05-15
Find files with trailing whitespace

sjoelund:
The CI should fail if we find any LaTeX file with trailing whitespace.

Note: This should not be possible to merge until all whitespace has been removed (which we should not do until #2353 is fixed).

### DIFF
--- a/.CI/Jenkinsfile
+++ b/.CI/Jenkinsfile
@@ -4,6 +4,17 @@ pipeline {
     disableConcurrentBuilds()
   }
   stages {
+  stage('checks') {
+    agent {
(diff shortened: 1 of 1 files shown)


sjoelund:
[Jenkins (ephermal link; will be broken in the future)](https://test.openmodelica.org/jenkins/blue/organizations/jenkins/ModelicaAssociation%2FModelicaSpecification/detail/PR-2361/4/pipeline) now shows the files (and lines) that improperly end with a whitespace. We could possibly also use a pre-commit hook for trimming trailing whitespace automatically.

PR #2420, opened 2019-10-03
Text markup for captions

henrikt-ma:
As decided at the 100th design meeting, there should be a discussion about text markup for captions.  This is what we propose.

### DIFF
--- a/RationaleMCP/0033/predefined-plots.md
+++ b/RationaleMCP/0033/predefined-plots.md
@@ -94,3 +94,28 @@ When an axis bound isn't provided, the tool computes one automatically.
 The Modelica tool is responsible for showing the unit used for values at the axis tick marks, so the axis `label` shall not be used to convey this information.
 
 When an axis label isn't provided, the tool produces a default label.  Providing the empty string as axis label means that no label should be shown.
+
+## Variable replacements
+In most places where text is displayed (`title`, `caption`, `legend`, `label`), the final value of a result variable can be embedded by refering to the variable as `%{inertia1.w}`.  This is similar to the `Text` graphical in 18.6.5.5 _Text_,.
+
+Note that expansion to the final value means that expansion is not restricted to parameters and constants, so that values to be shown in a caption can be determined during simulation.
+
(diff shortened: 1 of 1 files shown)


henrikt-ma:
@otronarp, will you let us know when the prototype has implemented what is proposed here?

DagBruck:
Ok, let's go for it.

- Hans and Dag

otronarp:
I have implemented support for the proposed text markup in captions now. Here is an example caption that references subplots and variables. Example figure

henrikt-ma:
I'm requesting reviews both from additional tool vendors as well as persons involved in MAP-Lib, as it would be great if we could detect any disagreement about how this should be done before entering the formal evaluation of the entire MCP.  I'll give it a week or so, before going ahead and preparing specification changes.

RationaleMCP/0033/predefined-plots.md
```diff
@@ -94,3 +94,28 @@ When an axis bound isn't provided, the tool computes one automatically.
...
+Note that expansion to the final value means that expansion is not restricted to parameters and constants, so that values to be shown in a caption can be determined during simulation.
+
+The percent character is encoded `%%`.  Neither `%class` nor `%name` is supported in this context, as this information is expected to already be easily accessible (when applicable) in tool-specific ways.  (Titles making use of `%class` or `%name` would then only lead to ugly duplication of this information.)
+
+## Text markup in captions
+In addition to variable replacements, a very restricted form of text markup is used for the `caption`.
+
+### Links
+Links take the form `%[<text>](<link>)`, where the `[<text>]` part is optional.  The `<link>` can be in either of the following forms:
+- A URI, such as `https://github.com/modelica/ModelicaSpecification` or `modelica:///Modelica.Blocks`.
```
dietmarw: Just for clarification. When you say "`[<text>]` part is optional" this leaves `%(<link>)` and not `%<link>`. So what would be the correct syntax here?

otronarp: Exactly the intention is that the two valid forms should be:

- `%[text](<link>)` 
- `%(<link>)`

PR #2441, opened 2019-10-29
Removed the redundant words

adeas31:


adeas31:
True but now I guess the pdf and html versions are generated automatically.

PR #2469, opened 2020-01-07
Settle the top level structure

henrikt-ma:
Opening PR for roadmap item 'Settle the top level structure'.

Note that operator overloading is being removed in the current state.  If we want to keep them, we should add production rules specifically for these, so that we avoid reintroducing recursiveness of _class-definition_.

Before being ready for merge, we should first merge https://github.com/modelica/ModelicaSpecification/pull/2465 (and resolve the resulting conflicts).

### DIFF
--- a/RationaleMCP/0031/grammar.md
+++ b/RationaleMCP/0031/grammar.md
@@ -86,13 +86,13 @@ end _F;
 > &emsp; ~~**partial**?~~\
 > &emsp; ( ~~**class**~~\
 > &emsp; | ~~**model**~~\
-> &emsp; | **operator**? **record**\
+> &emsp; | ~~**operator**?~~ **record**\
 > &emsp; | ~~**block**~~\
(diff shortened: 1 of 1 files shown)


henrikt-ma:
Note that this PR is where we are supposed to decide whether to introduce a new keyword such as `flat_model`.  With the current proposed structure, where this keyword wouldn't appear near the beginning of the file anyway, there seems to be little reason for introducing a new keyword, which is why `model` is the one being proposed.

olivleno:
Web Meeting: Overloaded operator can bee added to future version if it turns out that they are needed. For now they shall be removed.

olivleno:
Web Meeting: 
- [x] Agreed to have version header  according to Modelica MCP.
- [x] All class declarations above the model.
- [x] Keyword for a flat model shall be: model
- [x] Agreed to not having a top level container

henrikt-ma:
Rebased.  Marking as _Ready for review_.

PR #2475, opened 2020-02-03
Clarify the consequence of using impure function in binding equation for parameter

henrikt-ma:
The old formulation sounded almost as if the value of the impure function call wasn't even respected at initialization.

### DIFF
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -319,11 +319,14 @@ from within:
 \item
   in initial equations and initial algorithms,
 \item
-  in bindings for variables declared as parameter -- which is seen as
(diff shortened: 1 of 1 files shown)


henrikt-ma:
That's right.  I was thinking this wasn't an issue because of the rules disallowing the expression to end up in the iterative solution of an equation system, but failed to see that there may also be identical expressions that are also evaluated at initialization, and cause similar discrepancy.  Still, I would suggest that we leave it as is, since it's only a non-normative clarification after all, and one would have to be very careful about how to word things in order to not cause confusion about that equation system rule at the same time.

PR #2482, opened 2020-02-12
MCP-0033 Annotations for Predefined Plots

henrikt-ma:
We are now beginning the collection of reviews in order to be able to enter the MCP state _under evaluation_.

This PR is currently in draft state, since reviewers might raise non-trivial issues that we would like to address before entering _under evaluation_.  When entering _under evaluation_, the draft state will be removed as a signal of not intending to commit anything but minor corrections.


chapters/annotations.tex
```diff
@@ -68,27 +73,185 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+\end{lstlisting}
+
+A \lstinline!Plot! can contain several \lstinline!curves!, see
+\autoref{plot-curves}, that all share a common x and y axis with properties
+described in \autoref{axis-properties}.
+
+The \lstinline!title! of a \lstinline!Figure! shall be non-empty and is
+mandatory, as every figure needs a title for presentation in contexts such as a
+list of figures, and as it is generally hard for tools to automatically generate
+a meaningful title.  On the other hand, when the \lstinline!title! of a
```
henrikt-ma: The poll at the 101th design meeting determined that the `title` mustn't be required.  Instead, it should have a default value of `""`, where `""` means that the tool must produce a non-empty title based on the figure content.

chapters/annotations.tex
```diff
@@ -68,27 +73,185 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+
+The Modelica tool is responsible for showing the unit used for values at the
+axis tick marks, so the axis \lstinline!label! shall not be used to convey this
+information.
+
+When an axis label isn't provided, the tool produces a default label. Providing
+the empty string as axis label means that no label should be shown. Variable
+replacements, as described in \autoref{variable-replacements}, can be used in
+the \lstinline!label! of \lstinline!Axis!
+
```
henrikt-ma: This could be a good place to insert a section about the inheritance of figures.  Since inheritance of annotations isn't a well defined concept, it probably needs a sentence or two of explanation.  Ideally, the way we write this should also be coordinated with how we explain the inheritance of `experiment` annotations (decided, but not yet formulated).

henrikt-ma:
@GallLeo, any idea why I can't add you to the list of suggested reviewers?  Maybe @HansOlsson knows?

HansOlsson:
> @GallLeo, any idea why I can't add you to the list of suggested reviewers? Maybe @HansOlsson knows?

Could be because he wasn't a member in the project, I have changed that.

henrikt-ma:
Funny, because to me he seemed (and still seems to be) the only member: https://github.com/orgs/modelica/teams/map-lang

HansOlsson:
> Funny, because to me he seemed (and still seems to be) the only member: https://github.com/orgs/modelica/teams/map-lang
> 
> I guess projects and teams are different things…

Yes, the map-lang team didn't have rights for ModelicaSpecification.

HansOlsson:
The automatically generated plot-labels (not to be confused with axis-labels) are not clear to me, and I'm not sure if they are needed as part of the normative text.

In Dymola plot-windows have a number 1, 2, 3... to be able to tell them apart - do I understand that the label is similar to that?

Obviously it is good that you can talk about the different plots - but I don't see how these labels are used currently.

chapters/annotations.tex
```diff
@@ -68,27 +73,185 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+menu attached with menu items for plotting the variable, setting its start
+value, or investigating the equation system from which it is solved.  On the
+other hand, \lstinline!%[angular velocity](inertia1.w)! could be formatted in
+the same style as the surrounding text, except some non-intrusive visual clue
+about it being linked.
+\end{nonnormative}
+
+A sequence of one or more newlines (encoded either literally or using the \lstinline!\n!
+escape sequence) means a paragraph break.  (A line break within a paragraph is
+not supported.)
```
henrikt-ma, line 254: It should also be stated that newlines before the first paragraph or after the last paragraph are ignored.

henrikt-ma:
> The automatically generated plot-labels (not to be confused with axis-labels) are not clear to me, and I'm not sure if they are needed as part of the normative text.

They need to be part of normative specification, since they way one writes the caption depends on how one can refer to the different plots of a figure.  If a tool would fail to generate the labels, the users of this tool would find themselves inventing workarounds for referring to the different plots, instead of using the dedicated linking mechanism; nobody would write a caption like `In %(plot:foo), the foo is seen…` if the tool doesn't support rendering the link in a useful way. For the link to be useful, the tool must then attach some sort of label to the plot, so that this label can be used when rendering `%(plot:foo)` in the caption.

> In Dymola plot-windows have a number 1, 2, 3... to be able to tell them apart - do I understand that the label is similar to that?

Yes, unless a "plot-window" shows the entire `Figure`, instead of just a `Plot`; in this case the number would be a _figure label_, but the MCP relies on the figure title instead of having such labels.  On the other hand, if a "plot-window" shows just a plot, I find it difficult to imagine how you would organize the different windows and the caption…

> Obviously it is good that you can talk about the different plots - but I don't see how these labels are used currently.

Is it more clear now?

HansOlsson:
> > The automatically generated plot-labels (not to be confused with axis-labels) are not clear to me, and I'm not sure if they are needed as part of the normative text.
> 
> They need to be part of normative specification, since they way one writes the caption depends on how one can refer to the different plots of a figure. If a tool would fail to generate the labels, the users of this tool would find themselves inventing workarounds for referring to the different plots, instead of using the dedicated linking mechanism; nobody would write a caption like `In %(plot:foo), the foo is seen…` if the tool doesn't support rendering the link in a useful way.

It's still not clear as I don't see a mechanism for getting the label. To me it seems that the real requirement is that you can use this linking mechanism to refer to the plot, and the label is just a way to implement that (and avoid having the same text for all links). Am I missing something?

henrikt-ma:
> It's still not clear as I don't see a mechanism for getting the label. To me it seems that the real requirement is that you can use this linking mechanism to refer to the plot, and the label is just a way to implement that (and avoid having the same text for all links). Am I missing something?

No, the label is just the key to the visual linking.  For example, a caption may be rendered like this:
> **Step response.** _**(a)** Reference tracking.  **(b)** Input signal._

Here, the _**a**_ and _**b**_ are the plot labels, which must also be shown in connection to the actual plots, so that the user can tell what the different parts of the caption refer to.  The corresponding `caption` source would look like this:
```
%(plot:tracking) Reference tracking.  %(plot:input) Input signal.
```

That the plot labels are shown within parentheses in the caption is just the way this imaginary tool chose to style an automatically generated plot link, and the labels could possibly be formatted differently when shown in connection to the respective plots.

The actual choice of labels is for the tool to decide, so if Dymola wants to use numbers, that would be perfectly fine.  Besides giving each tool the freedom to choose a scheme that fits the rest of the tool environment, this also opens up for user configuration, so that figures may be adapted to different company or publication guidelines.

otronarp:
@HansOlsson I think one reason for having the labels as a part of the standard is to make it clear that it is a tool responsibility to create usable (short) visual linking of plots so users don't start to write titles like this:

> **a)** A Title
> **b)** An Other Title

And then use **(a)** and **(b)** directly in captions.

chapters/annotations.tex
otronarp: I have made the conclusion that they are completely orthogonal. Here, we don't need any merging/overriding/etc but there can exist a `figures` annotation on any level in the inheritance hierarchy and they live their own life. However, maybe we should make it explicit that you are expected to access (show) Figures from inherited classes as well?

henrikt-ma: Yes.  Perhaps #2544 should be formulated more in terms of inheritance?  Perhaps something like this:
> The `figures` annotation is inherited in the sense that each class has a collection of figures comprised by the contents of the `figures` annotation in the class itself, as well the `figures` annotations from any base classes.

Then, the uniqueness requirement on the `Figure` `identifier` follows naturally.

It might also be worth noting that there could be a difference between referring to (from the `info`) an inherited figure via the present class versus via a parent class.  The difference might not be clear now, but it would probably make a difference when thinking of what simulation result to use when rendering figures inline in the class documentation.

chapters/annotations.tex
```diff
@@ -43,14 +43,19 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
 
 \begin{lstlisting}[language=grammar]
 documentation-annotation:
-   annotation "(" Documentation "(" "info" "=" STRING ["," "revisions" "=" STRING ] ")" ")"
+   annotation "(" Documentation "(" [ "info" "=" STRING ]
```
HansOlsson, line 46: This does not work, since `,` is a separator and if `info `is a missing there should be no `,` before `revisions`, and if both `info` and `revisions` are missing we should not have `, figures=…`. The more complicated question is: what works?
One variant would be:
```
documentation-annotation: annotation "(" Documentation "(" doc-item {", " doc-item} ")" ")"
doc-item: "info" "=" STRING |
                 "revisions" "=" STRING |
                 "figures" "=" figure-annotations
```
Another would be to use a record to describe it instead.

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
 \begin{lstlisting}[language=modelica]
 <a href="Modelica://MultiBody.Joints.Revolute#info">Revolute</a>
 \end{lstlisting}
-\begin{lstlisting}[language=grammar]
-preferred-view-annotation:
-   annotation "(" preferredView "=" ("info" | "diagram" | "text") ")"
+
+\subsection{Annotations for Figures\doublelabel{annotations-for-figures}}
```
HansOlsson, line 77: The labels are normally after the subsection. It could be that they should be inside the sections, but then we should redo everything (in a separate PR).

chapters/annotations.tex
```diff
@@ -1031,9 +1199,32 @@ \subsubsection{Edit input}\doublelabel{edit-input}
 end OnMouseDownEditString;
 \end{lstlisting}
 \section{Annotations for the Graphical User Interface}\doublelabel{annotations-for-the-graphical-user-interface}
+This section describes the annotations that are used to define properties of the
```
HansOlsson, line 1202: Why does it look like this unrelated annotation was cut and pasted instead of just text inserted before it?
Was there some change in the text (I didn't read that closely)? Was it moved?

henrikt-ma: I think the changes are just minor, including adding a paragraph break before the `documentation-class-annotation` code block.

henrikt-ma: OK, it is actually more than that.  Both the `documentation-class-annotation` and `preferred-view-annotation` are now described with the other _Annotations for the Graphical User Interface_.

chapters/annotations.tex
henrikt-ma: Fixed.  Resolving.

chapters/annotations.tex
henrikt-ma: The change only follows the existing style used for `revisions`, so unless `revisions` was only meant to be allowed when there is an `info` this should be fixed in a separate PR.  If `revisions` was only meant to be present when there is an `info`, your proposal should be changed accordingly (although I'm afraid one would need some extra lookahead in the parser to handle it):
```
doc-item: "info" "=" STRING [, "revisions" "=" STRING] |
                 "figures" "=" figure-annotations
```

Besides, I'd say that the `revisions` is very poorly documented:
> the `revisions` annotation giving a list of revisions and other annotations defined by a tool [_The `revisions` documentation may be omitted in printed documentation_].

It doesn't even seem to match current practice in the MSL, where it is set to a string, not a list.

chapters/annotations.tex
otronarp: With the motivation that they control how the UI should handle the class and doesn't have anything to do with the actual documentation.

otronarp:
I have created a test model with several example figures demonstrating the various features of the MCP, it's available here: https://github.com/otronarp/MCP033Examples/tree/master/MCP0033. I also added the resulting figures generated with [System Modeler](https://github.com/otronarp/MCP033Examples/blob/master/SystemModeler/README.md).

chapters/annotations.tex
otronarp: Addressed by #2580.

chapters/annotations.tex
otronarp: Addressed by #2581

chapters/annotations.tex
otronarp: @HansOlsson, I consider this resolved now. Feel free to reopen the discussion if you don't agree.

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+end Curve;
 \end{lstlisting}
 
-The \lstinline!preferredView! annotation defines the default view when selecting the
-class. \lstinline!info! means info layer, i.e., the documentation of the class,
-\lstinline!diagram! means diagram layer and \lstinline!text! means the Modelica text layer.
-\begin{lstlisting}[language=grammar]
-documentation-class-annotation:
-   annotation "(" DocumentationClass "=" true ")"
+The mandatory \lstinline!x! and \lstinline!y! expressions are restricted to be
```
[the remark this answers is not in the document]
otronarp: They have to be specified, and since `x` has a default binding it's already specified.

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+
+The \lstinline!identifier! in \lstinline!Figure! and \lstinline!Plot! is an
+optional \lstinline!String! identifier, and is intended to identify the
+\lstinline!Figure! and \lstinline!Plot! for programmatic access.  A
+\lstinline!Figure! must be uniquely identified by its \lstinline!identifier! and
+containing class.  This means that a \lstinline!Figure! \lstinline!identifier!
+must be unique among all \lstinline!Figure! annotations within the same
+\lstinline!figures! annotation as well as among all \lstinline!figures!
+annotations from inherited classes.  A \lstinline!Plot! \lstinline!identifier!
+on the otherhand is only required to be unique among the \lstinline!plots! in
```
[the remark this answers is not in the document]
otronarp: Fixed in 561af3a4d0bd05a50bc2da14d144fc76f2f801e3

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+For \lstinline!Figure!, this makes it possible to reference the plot from a
+tool-specific scripting environment. For \lstinline!Plot!, this makes it
+possible to reference the plot in the figure caption, which becomes useful when
+the \lstinline!Figure! contains more than one \lstinline!Plot!.
+\end{nonnormative}
+
+Every \lstinline!Plot! has an automatically generated \emph{label} which is
+required to be shown as soon as at least one \lstinline!Plot! in the
+\lstinline!Figure! has an \lstinline!identifier!.  A tool is free to choose both
+labeling scheme (such as \emph{a}, \emph{b}, \dots, or \emph{i}, \emph{ii}, \dots) placement in the plot,
```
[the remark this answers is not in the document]
otronarp: Fixed in 561af3a4d0bd05a50bc2da14d144fc76f2f801e3

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
-documentation-class-annotation:
-   annotation "(" DocumentationClass "=" true ")"
+The mandatory \lstinline!x! and \lstinline!y! expressions are restricted to be
+component references refering to a scalar variable or \lstinline!time!.
+
+When \lstinline!legend! isn't provided, the tool produces a default based on
+\lstinline!x! and/or \lstinline!y!.  Providing the empty string as
+\lstinline!legend! means that the curve shall be omitted from the plot legend.
+Variable replacements, as described in \autoref{variable-replacements}, can be
+used in the \lstinline!legend! of \lstinline!Curve!
```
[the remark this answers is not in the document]
otronarp: Fixed in 561af3a4d0bd05a50bc2da14d144fc76f2f801e3

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+In addition to variable replacements, a very restricted form of text markup is
+used for the \lstinline!caption!.
+
+Links take the form \lstinline!%[<text>](<link>)!, where the
+\lstinline![<text>]! part is optional. The \lstinline!<link>! can be in either
+of the following forms:
+\begin{itemize}
+\item A URI, such as
+\lstinline!https://github.com/modelica/ModelicaSpecification! or
+\lstinline!modelica:///Modelica.Blocks!.
```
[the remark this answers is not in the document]
otronarp: Sadly you are right, fixed in 561af3a4d0bd05a50bc2da14d144fc76f2f801e3

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+reference such as \lstinline!inertia1.w!.
+\item A \lstinline!plot:<id>!, where \lstinline!<id>! is the identifier of a
+\lstinline!Plot! in the current \lstinline!Figure!.
+\end{itemize}
+
+When \lstinline![<text>]! is omitted, a Modelica tool is free to derive a default based on
+the \lstinline!<link>!.
+
+The styling of the link text and the link action is left for each Modelica tool
+to decide.
```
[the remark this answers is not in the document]
otronarp: Fixed in 561af3a4d0bd05a50bc2da14d144fc76f2f801e3

chapters/annotations.tex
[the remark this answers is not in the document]
otronarp: Fixed in 561af3a4d0bd05a50bc2da14d144fc76f2f801e3

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+
+When an axis label isn't provided, the tool produces a default label. Providing
+the empty string as axis label means that no label should be shown. Variable
+replacements, as described in \autoref{variable-replacements}, can be used in
+the \lstinline!label! of \lstinline!Axis!
+
+\subsubsection{Variable Replacements}\doublelabel{variable-replacements}
+In most places where text for display is defined the final value of a result
+variable can be embedded by refering to the variable as
+\lstinline!%{inertia1.w}!.  This is similar to the \lstinline!Text! graphical in
```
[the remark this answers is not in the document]
otronarp: Maybe, or  "similar to the Text graphical _primitive_ in", since that's how they are called in that section?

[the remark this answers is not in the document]
otronarp: Fixed in f5b0e3a988c836e3d683585ce559f72b91607fba

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+A \lstinline!Plot! can contain several \lstinline!curves!, see
+\autoref{plot-curves}, that all share a common x and y axis with properties
+described in \autoref{axis-properties}.
+
+Both \lstinline!Figure! and \lstinline!Plot! can have an optional title. When
+the \lstinline!Figure! \lstinline!title! is the empty string (the default), the
+tool must produce a non-empty title based on the figure content.  On the other
+hand, the \lstinline!Plot! \lstinline!title! has a tool-dependent default, but
+the default may be the empty string.  When the Plot title is the empty string, no
+title should be shown. The plot title is not to be confused with the plot
```
henrikt-ma, line 111: I think one has to choose between "`Plot` `title`" and "plot title".  It makes sense to say "plot title" when relating to "plot label", but I think the "Plot title" should be "`Plot` `title`".

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+
+A \lstinline!Plot! can contain several \lstinline!curves!, see
+\autoref{plot-curves}, that all share a common x and y axis with properties
+described in \autoref{axis-properties}.
+
+Both \lstinline!Figure! and \lstinline!Plot! can have an optional title. When
+the \lstinline!Figure! \lstinline!title! is the empty string (the default), the
+tool must produce a non-empty title based on the figure content.  On the other
+hand, the \lstinline!Plot! \lstinline!title! has a tool-dependent default, but
+the default may be the empty string.  When the Plot title is the empty string, no
```
otronarp, line 110: @henrikt-ma do you mean like this?
```suggestion
the default may be the empty string.  When the \lstinline!Plot! \lstinline!title! is the empty string, no
```

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+\begin{nonnormative}
+For \lstinline!Figure!, this makes it possible to reference the plot from a
+tool-specific scripting environment. For \lstinline!Plot!, this makes it
+possible to reference the plot in the figure caption, which becomes useful when
+the \lstinline!Figure! contains more than one \lstinline!Plot!.
+\end{nonnormative}
+
+Every \lstinline!Plot! has an automatically generated \emph{label} which is
+required to be shown as soon as at least one \lstinline!Plot! in the
+\lstinline!Figure! has an \lstinline!identifier!.  A tool is free to choose both
```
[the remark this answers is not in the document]
henrikt-ma: It was a deliberate design choice to make the visual appearance depend in a simple fashion on the presence of `identifier`.  Two use cases supporting this choice (assuming the plot isn't referenced from the `caption`):
- One might want to refer to the plot from a different figure or the `documentation` annotation, perhaps even from a different package.  As long as the tool is aware of the plot being referenced, it should be able to provide a nicely formatted hyper-link to that plot.  Currently, there is no standardized way of doing this, but there is MCP0037 _Generalized Modelica URIs for external resources_ that should address this need.
- The figure will be referenced from a different context without the possibility of hyper-linking the plots, where one actually wants to refer to the respective plots by label.

chapters/annotations.tex
```diff
@@ -68,27 +73,190 @@ \section{Annotations for Documentation}\doublelabel{annotations-for-documentatio
...
+organize figures similar to how \lstinline!group! is used in the
+\lstinline!Dialog! annotation (see \autoref{annotations-for-the-graphical-user-interface}).  However, leaving \lstinline!group! at
+the default of an empty string does not mean that a group will be created
+automatically, but that the figure resides outside of any group. The
+\lstinline!group! is both the key used for grouping, and the name of the group
+for display purposes.
+
+If the \lstinline!preferred! attribute of \lstinline!Figure! is \lstinline!true!
+tools should should show the figure automatically when the class is simulated. A
+class may define multiple \lstinline!preferred! figures.
```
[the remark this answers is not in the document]
henrikt-ma: I think we should start by writing the normative part so that it doesn't require things that don't always make sense, and hopefully this removes the need for a non-normative elaboration.  How about this?
> The `preferred` attribute of `Figure` indicates whether the figure should be given preference when automatically determining which figures to show.  For example, a tool might choose to automatically show all preferred figures when the class is simulated.

chapters/annotations.tex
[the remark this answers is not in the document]
maltelenz: Isn't your second use case dangerous, since it would make the reference only "work" in a single tool, since each tool is free to decide the label appearance?
I can see the argument with future Modelica URIs.

chapters/annotations.tex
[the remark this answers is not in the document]
maltelenz: Sounds great to me!

chapters/annotations.tex
henrikt-ma: Just so we get somewhere here, I made a PR for this: https://github.com/modelica/ModelicaSpecification/pull/2588

chapters/annotations.tex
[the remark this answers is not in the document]
henrikt-ma: Fixed.  Resolving.

chapters/annotations.tex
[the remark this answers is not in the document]
henrikt-ma: Yes, I am aware of that argument against the second case.  It will work in a context where the tool behavior is known, or the figure appearance has been locked by rendering it to a graphics format.  I see that this could be addressed by trying to agree on a labeling scheme, but I'd like to avoid it if we can still reach agreement.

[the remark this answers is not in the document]
maltelenz: I'm personally fine with the current text given your motivation, but I would also be fine with my initial suggestion, knowing that it would limit the possibilities with Modelica URIs.

I do not want the specification to specify the labeling appearance, I think that would be to restrictive.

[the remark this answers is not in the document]
otronarp: I assume @henrikt-ma means that the user reference using the identifier, but the tool displays it using the label, so it should be safe?

chapters/annotations.tex
henrikt-ma: @HansOlsson, should we go for the pseudo code record to resolve this?

HansOlsson: That's likely the best solution if we want something formal.
Regarding the previous issues: I believe that revisions was almost always combined with info so in practice that made sense, and I think "list" mean that the html-code contain a list of revisions - not that revision-annotation itself was a list.

chapters/annotations.tex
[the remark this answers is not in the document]
otronarp: I'll interpret that as this is resolved then?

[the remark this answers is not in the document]
maltelenz: Yes. I don't seem to be able to mark it as such (or I'm just to stupid to find the button, but it's not where the github help indicates it should be).

chapters/annotations.tex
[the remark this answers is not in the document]
henrikt-ma: I'm not sure I understand the question, but I think we are talking about the potential issue resulting from a user referencing the plot by its label (because the reference is made in an external context where the identifier isn't accessible).  If the same figure is then re-rendered using a different tool (or the same tool, but with different setting for the label scheme), the hard-coded label in the reference could (would) be left dangling.  So: not safe, but not important enough to motivate sacrificing the freedom to choose scheme.

chapters/annotations.tex
henrikt-ma: To be resolved by #2594.

henrikt-ma:
Dear @GallLeo and @gkurzbach, now that @maltelenz has provided one review and @DagBruck has made a second test implementation that I think can be counted also as a review, we are down to one missing review before we can move this MCP to the next stage.  Could (one of) you please consider taking a look?  Thanks!

chapters/annotations.tex
henrikt-ma: The PR #2588 has been merged.  Resolving conversation.

henrikt-ma:
> Approve that the MCP is presented to MA for inclusion, but we will request one minor addition:
> 
> > If links or variable replacements are used, a tool may use them to substitute the references in the text.
> 
> The reason is that such substitutions cannot be assumed to be implementable with reasonable effort. This is analogous to how `connectorSizing` is specified.

To prevent this discussion starting all over here, I'd like to point out that there is an ongoing discussion on this matter in #2578.  Please let us continue the discussion there!

henrikt-ma:
Thanks!  I'll consider this a review, and make sure we round off discussing these items before entering _Under evaluation_.  I'm afraid that this will become too many parallel questions in the PR conversation, and if things take off in that direction I think we should open separate issues for the different topics, or have review comment discussions attached to changed lines of the specification text.

> In general the concept of this MCP fits well with what we have in SimulationX (except for renaming).

For what it's worth, SystemModeler is not updating its user interface to reflect the names of various annotations.  To some extent, the way things are called in the user interface isn't always for us to decide at the SystemModeler level.  Hence, a discrepancy between what things are called in Modelica and what they are called in the user interface is more or less expected anyway, and what we are proposing are names for maximum clarity of the Modelica specification.

> There is a significant difference: In SimulationX the Axis specification is related to the Curve. This means that there can be more than one axis (x or y) in a plot if there is more than one curve with different units.

SystemModeler also supports use of a second y axis, which will be handled using vendor annotations, probably something like this:
```
curves = {
  Curve(y = var1, __Wolfram_yaxis = "left"),
  Curve(y = var2, __Wolfram_yaxis = "right")
},
__Wolfram_yRight=Axis(…)
```
Since the presence of multiple axes increase the effort required for implementation, we'd like to keep this feature out of the MCP, with possibility to add it in the future.  (I also think it still makes good sense to keep the actual axis definition separate from the curves, since it is common for an axis to be shared between several curves.)

> The proposal assumes that all curves in a plot have the same units, and this is not as general as it could be. Perhaps, to fix it, there could be an optional Axis specification in the Curve that overwrites the one in the plot if present.

I'd say it's often bad style to mix curves with incompatible units, and that it is better to begin by leaving support for such plots up to tools to provide using vendor annotations.  Again, we could revisit this in the future, but then I'd like to just include a conversion factor (typically just a 1 of some `displayUnit`) in the `Curve`, so that the tool can decide how to present the conversion in legends, on axes, etc.

> One small thing: The semantics of the group in the figure is not really clear, especially why it is only present in the Figure and not in the Plot as well.

Please leave a review comment for the lines where `group` is defined, so that we can start a separate on how to clarify.  Short answer: The plots of a figure are meant to be placed in some simple arrangement (typically in a column or a grid) in the figure with no obvious use for grouping of the kind that makes sense for figures; while a model could contain lots and lots of figures calling for some way to organize them, each figure should just contain a careful selection of plots that make sense to have in the same figure.  However, in a future where we can have stylesheets for our figures, I do see a possibility to define a CSS style class for each plot, which could then be used to do grouping that could be reflected in how the different groups are styled.

> We miss also the specification of colors and sizes of the diagrams. So a tool has to be very "intelligent" when plotting the first time diagrams created with another tool. This complicates the implementation and may lead to different results in different tools.

Yes, the MCP is missing many aspects of styling that would be required for a fully featured figures mechanism — as reported by many (all?) reviewers at various stages of the MCP development.  Leaving out styling is intentional, as styling is such an important aspect that needs careful design, so that trying to include it would risk delaying standardization indefinitely.

What this MCP provides is a content model that in its simplicity has proven itself to give [good results](https://github.com/otronarp/MCP033Examples/tree/master/SystemModeler), and that will let the MAP-Lib start the work on defining figure content for MSL examples (using figures for MSL 3.2.3 provided by Wolfram as a starting point).

Personally, I'm looking forward to working on some sort of style sheet solution for figures once we are finished with this MCP (the new Modelica URIs could also be useful for referencing style sheets).  A good style sheet solution should both allow style sheets local to a figure that can control the appearance of individual curves, as well as library-wide or even company-wide style sheets for consistent styling across a library or organization.  I think there is a lot to learn from the success of CSS for HTML here!

henrikt-ma:
> One small thing: The semantics of the group in the figure is not really clear, especially why it is only present in the Figure and not in the Plot as well.

To explain the use of figure grouping a bit more, imagine having a big model with several major subcomponents, like the powertrain of a car.  Then figures could be grouped by major subcomponent (taken from https://en.wikipedia.org/wiki/Powertrain), where the different groups will have different amounts of figures depending on the purpose and focus of the powertrain model:
- Engine
- Transmission
- Drive shafts
- Differentials
- Wheels

Another example could be to group figures according to:
- Performance measures: Figures for high level evaluation of system performance.
- Disturbances: Figures for various disturbance models included in the plant.
- Control systems: Figures for investigation operation of various controllers in the system.
- Thesis: Figures to be included in the author's thesis.
- Regression testing: Figures containing additional variables that are tracked for detection of errors in the simulation result.

For plots, I think that one should first consider splitting a plot into several figures before thinking about some sort of grouped arrangement of a figure's plots.  If anyone has some nice examples of grouped plots inside a figure, please share!

DagBruck:
> For plots, I think that one should first consider splitting a plot into several figures before thinking about some sort of grouped arrangement of a figure's plots. If anyone has some nice examples of grouped plots inside a figure, please share!

In Dymola we have a Commands menu where we put all the predefined plots according to MCP-0033. We use the grouping to create a sub-menu of plot commands. Maybe that was not the intended use, but to me it seems quite natural.

henrikt-ma:
> > For plots, I think that one should first consider splitting a plot into several figures before thinking about some sort of grouped arrangement of a figure's plots. If anyone has some nice examples of grouped plots inside a figure, please share!
> 
> In Dymola we have a Commands menu where we put all the predefined plots according to MCP-0033. We use the grouping to create a sub-menu of plot commands. Maybe that was not the intended use, but to me it seems quite natural.

If by _predefined plots_, you mean the figures (and not the individual plots contained in the figures), then this is exactly the intended use of the _figure_ grouping.  What @gkurzbach was suggesting was rather grouping of the plots _inside_ a given figure.  @DagBruck, are you saying Dymola has generalized the idea of grouping down to the plot level?

HansOlsson:
> If by _predefined plots_, you mean the figures (and not the individual plots contained in the figures), then this is exactly the intended use of the _figure_ grouping. What @gkurzbach was suggesting was rather grouping of the plots _inside_ a given figure. @DagBruck, are you saying Dymola has generalized the idea of grouping down to the plot level?

I believe Dag meant the first, i.e., figure grouping, as I think the plots within a figure (or sub-plot in a plot as normally described in Dymola) are just unstructured.

DagBruck:
Sorry, I was unclear. I meant figure grouping of course, just that we applied grouping to create hierarchical menus.

henrikt-ma:
All PR targeting this MCP branch are now merged, and the MCP is now entering state _Under evaluation_ as agreed during yesterday's phone meeting.

I thought now was the time to remove the draft state of the PR, but it had already been removed, I don't remember when…

The three required reviews have been provided by the following persons, see above: @maltelenz, @DagBruck, @gkurzbach

henrikt-ma:
@HansOlsson is there any reason to delay the voting further, or could you please make the announcement?

henrikt-ma:
While currently in the state _Under evaluation_, some smaller issues have come to our attention, resulting in two PRs:
- #2691
- #2692

These are considered minor enough to not abort the ongoing voting for acceptance of the MCP.  Instead, we hope that the MCP can be accepted despite these issues, and that the issues can still be resolved before the figure annotations appear for the first time in a released version of the specification.

sjoelund:
Since it was unclear for some people voting, the generated PDF and HTML files are available in the [CI/CD system](https://test.openmodelica.org/jenkins/blue/organizations/jenkins/ModelicaAssociation%2FModelicaSpecification/detail/MCP%2F0033/22/artifacts/) in case you don't want to read the LaTeX code.

adeas31:
Is there a working prototype of this MCP available in any Modelica tool e.g., SystemModeler? SystemModeler 12.0.0.11 doesn't support this, right?

HansOlsson:
> Is there a working prototype of this MCP available in any Modelica tool e.g., SystemModeler? SystemModeler 12.0.0.11 doesn't support this, right?

Dymola 2021 included parts of it (but not the links-part as far as I recall - and there might have been some mistakes), and it will be improved in future Dymola versions (and likely the same will be included in versions of 3D Experience platforms as well).
They are found under Commands in the Simulation-tab, and you add them using Commands>Add command>plot window setup 'Based on proposed Modelica standard'.

Persons working on other tools will have to answer themselves.

otronarp:
> Is there a working prototype of this MCP available in any Modelica tool e.g., SystemModeler? SystemModeler 12.0.0.11 doesn't support this, right?

No SystemModeler 12.0.0.11 doesn't support it. 12.1 supports the proposal as it was around October 2019. However, it has changed too much since then so it cannot read the current format (there might be an intersection that works, but I have tried to find it). 12.2 will support the proposal in the current state.

HansOlsson:
The vote is now complete.

- 21 in favor
- 1 abstain
 - 0 against

Thus the MCP is accepted (every member should be able to follow the vote-link to confirm the result).
(Sorry for not sending it out yesterday afternoon when the vote closed.)

henrikt-ma:
Don't be sorry for the negligible delay.

Regarding the two open PRs against this branch, I would suggest that we now merge MCP/0033 and close this PR, but without removing the branch.  Then, we can just merge the branch again once we are done with the two remaining PRs.

PR #2504, opened 2020-02-20
Change SIunits to Units.SI as is the case as of MSL v4.0.0

dietmarw:
Question is that since MSL 4.0.0 is based on MLS v3.4 if this might also be backported to 3.4 spec, *if* there will be   maintenance update.


dietmarw:
OK.

PR #2526, opened 2020-03-15
Extend variability rules and fix invalid example

### LINKED ISSUE #2525: Invalid example of variability error
The example in 3.8.3 _Discrete-Time Expressions_ contains the following false statement about violating variability rules:
```
equation
  b = noEvent(x > 1) // error, since b is a discrete-time expr. and
                     // noEvent(x > 1) is not a discrete-time expr.
```

The problem is that there is no stated rule that allows use of variability to analyze normal equations.  All one can say is that the equation cannot be used to solve for anything, but with the current specification, one needs to draw that conclusion without help of expression variability rules.  Instead, one must go to the definition of variability of a variable to see that this equation cannot be used to solve for `b`, and then it becomes evident that the equation is useless and cannot be solved.

henrikt-ma:
Fixes #2525

### DIFF
--- a/chapters/classes.tex
+++ b/chapters/classes.tex
@@ -383,8 +383,11 @@ of initial value problem of the hybrid DAE):
 
 \begin{itemize}
 \item
-  A variable \lstinline!vc! declared with the parameter or constant prefixes remains
-  constant during transient analysis.
+  A variable \lstinline!vc! declared with \lstinline!constant! prefix remains constant during transient analysis,
+  with a value that is unaffected by the initialization problem.
+\item
+  A variable \lstinline!vc! declared with the \lstinline!parameter! prefix remains constant during transient analysis,
+  with a value determined by the initialization problem.
 \item
   A \emph{discrete-time} variable \lstinline!vd! has a vanishing time derivative between events.
   Note that this is not the same as saying that \lstinline!der(vd)=0! almost everywhere,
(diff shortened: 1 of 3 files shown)


gkurzbach:
Maybe I missed some discussion but 
`b2 = noEvent(x > 1);`
for me is clearly a variability error, because the right side may change outside of events but b2 not.

and a further question is whether the assignment from `i2 = x;` makes x to a discrete variable, because it cannot change between events.

henrikt-ma:
@gkurzbach, I think we all agree that there is a problem with the equation, the question is how to capture this intuition in the specification.  the reason we can't fail the normal equation `b2 = noEvent(x > 1)` due to variability error is that a normal equation has no prescribed direction of causality.  This is different in algorithms, inside when-clauses, and for declaration equations.  The point of this PR is to improve the variability rules so that this equation can be reported as an error with reference to variability rules in the specification, instead of needing to come up with a tool-specific explanation.

henrikt-ma:
Regarding `i2 = x`, and as seen in the ongoing discussion in #2521, I think there is disagreement regarding whether `x` should be considered discrete-time or not for purposes of variability analysis.  

This is my personal opinion on the matter: We shouldn't rely on this type of inferred variability when defining the variability of a variable.  Instead, I'd like to be able to declare `x` as `discrete Real`, and that `i2 = x;` would be a valid equation for `x` (that is, not requiring `x` to be assigned in a when-clause).

gkurzbach:
To the first point: I got it now. To the second: I agree with you.

PR #2537, opened 2020-03-23
Relax the uniqueness requirements of plot identifiers

otronarp:
I think that it would be better to only require that `Plot` `identifier` is unique within the `Figure`. For example if you want to duplicate a `Figure` with nice hand crafted identifiers for the plots it makes no sense to require the user to change them, it should be sufficient to only give the `Figure` a new `idenitifier`.

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -115,10 +115,11 @@ replacements, as described in \autoref{variable-replacements}, can be used in th
 \lstinline!title! of \lstinline!Figure! and \lstinline!Plot!.
 
 The \lstinline!identifier! in \lstinline!Figure! and \lstinline!Plot! is
-optional, and is intended for programmatic access.  An \lstinline!identifier!
-must be unique within the class containing the \lstinline!figures! annotation,
-without considering whether it belongs to \lstinline!Figure! or
-\lstinline!Plot!.
+optional, and is intended for programmatic access.  A \lstinline!Figure!
+\lstinline!identifier! must be unique within the class containing the
+\lstinline!figures! annotation, whereas a \lstinline!Plot!
+\lstinline!identifier! must be unique among the \lstinline!plots! in the
+\lstinline!Figure! annotation.
 
 \begin{nonnormative}
 For \lstinline!Figure!, this makes it possible to reference the plot from a


chapters/annotations.tex
```diff
@@ -115,10 +115,11 @@ \subsection{Annotations for Figures\doublelabel{annotations-for-figures}}
 \lstinline!title! of \lstinline!Figure! and \lstinline!Plot!.
 
 The \lstinline!identifier! in \lstinline!Figure! and \lstinline!Plot! is
-optional, and is intended for programmatic access.  An \lstinline!identifier!
-must be unique within the class containing the \lstinline!figures! annotation,
-without considering whether it belongs to \lstinline!Figure! or
-\lstinline!Plot!.
+optional, and is intended for programmatic access.  A \lstinline!Figure!
+\lstinline!identifier! must be unique within the class containing the
+\lstinline!figures! annotation.  Whereas a \lstinline!Plot!
```
henrikt-ma: Grammar:
> annotation, whereas a

henrikt-ma:
This is related to PR#2387, as a referring to a `Plot` `identifier` which is only unique within its `Figure` means that the Modelica URI must support more complicated forms than previously considered.  However, I don't think that the outlook of getting more complicated Modelica URIs would be reason to stick with the class-wide uniqueness requirement.  On the contrary, forcing ourselves into this more demanding situation now might help us avoiding painting ourselves into a corner with the new Modelica URIs.

chapters/annotations.tex
otronarp: Thanks, fixed.

otronarp:
> How are plots the referenced directly then? Probably with `figure_identifier.plot_identifier` or is it forbidden? This has to be clarified.
> 

The intention was that in a caption you can only reference plot within that figure, see https://github.com/modelica/ModelicaSpecification/blob/3a496ff852e4dc13892f68de7ab50d32a8a6f19f/chapters/annotations.tex#L233

Referencing plots from other figures would be done with an extension (yet to be designed, outside the scope of this MCP) to Modelica URIs.

> The formulation `must be unique within the class containing the \lstinline!figures! annotation` implies somehow that the figures are in the same namespace as the class components. This should be avoided, as annotations can not be references from Modelica. (This should be a comment to the whole MCP, because it is already in there.)

I guess we need to clarify that, the intention is that a figure identifie is unique among all figure identifiers within the same class (including inherited classes). Does that make more sense to you?

gkurzbach:
> The intention was that in a caption you can only reference plot within that figure, see
> 
> https://github.com/modelica/ModelicaSpecification/blob/3a496ff852e4dc13892f68de7ab50d32a8a6f19f/chapters/annotations.tex#L233
> 
I overlooked that.

> Referencing plots from other figures would be done with an extension (yet to be designed, outside the scope of this MCP) to Modelica URIs.
>
ok.

> I guess we need to clarify that, the intention is that a figure identifier is unique among all figure identifiers within the same class (including inherited classes). Does that make more sense to you?

Makes sense.

henrikt-ma:
@gkurzbach, one can get the impression that you were satisfied with the answers from @otronarp.  If you were, could you please update your review, if not, could you please clarify what changes you'd still like to see?

PR #2558, opened 2020-04-29
Simplify Modifications

olivleno:
To start the discussion


olivleno:
List of examples to look at:

Declare Record
```
record R
  Real x(min = 0.0);  // disallow default binding eq. 
  constant Real c = 1e5;
end R;

R r;

equation 
  r.x = 1.0;
```

Declare Array
```
Real A[2] = {1.0, 2.0};
```

Modification of a built-in attribute of a scalar
```
Real x (start=1);
```

Modification of a built-in attribute of a scalar within a record
```
record R
  Real x;
end R;

R r(x(start=1));
```

Next steps:
- Propose a way to get rid off variability prefixes in records.

olivleno:
Web-Meeting:

## Kinds of modifications:
1. Modify built-in types (discussion, see below)
2. Modify attributes (start, unit,...)
3. Modify values (non-attributes, including time varying) 

## Open issues: 
* Avoid merging of top level modifications
* final for attributes
* final for values
* modification of values (3.)

## Supported modifications with restrictions
* Modify of simple built-in types in type aliases (1.)
* Modify attributes (2.)
  * Modifications of top-level components only
    * Not allowed to modify on component declarations in records to avoid need of merging modifications
    * Though it's not overly complicated to support it would still add another layer of complexity by reducing the need for merging modifications that can be avoided.
    * To considered for later extensions.
```
// Example to illustrate need for restriction to top level modifications only
record Point
  Real x(unit="m") //not allowed, because not at the top level
  Real y(unit="m") //not allowed, because not at the top level
record end;
Point p(x(unit="cm")); 
```

### Modify built-in types

SI units are a very useful abstraction.
* Full support should be re-considered for v2,

Should allow generating FMUs from Flat Modelica.
* extensions of basic types (quantity, min, max, unit, ...)
* no records

```
// Full Modelica
record Point
  Modelica.SIUnits.Length x  //How to preserve this information of type Length?
  Modelica.SIUnits.Length y
record end;
Point p;

// Flat Modelica
record Point
  Real x annotation(original_type="Modelica.SIUnits.Length")
  Real y
record end;
Point p(x(unit="m"),y(unit="m"));

// Flat Modelica allowing basic types
constant Real 'Modelica.Math.Constants.pi' = 3.1415;
type 'Modelica.SIUnits.Length' = Real (unit="m", nominal='Modelica.Math.Constants.pi');  // only allowed for built-in scalar type
record Point
  'Modelica.SIUnits.Length' x(unit="cm") // not allowed
  'Modelica.SIUnits.Length' y
record end;
Point p(x(unit="cm"));  // allowed as modification of top level component (is no longer an 'Modelica.SIUnits.Length'!)
Point p2(x(nominal=2*pi ));  // allowed modification of top level component using literals or constant expressions 
```

Modification should give preference to simplicity over readability
* allow modification of built-in types, e.g. to declare a type
* allow modifications of top level components only

How to handle common expressions efficiently?
* That's a general concern, even more prominent for equations, not to be addressed here.

olivleno:
Web Meeting:
At which stage of the processing should unit checking be applied?
* There is no unit checking required to generate Flat Modelica.
* It shall be possible to apply unit checking to Flat Modelica.

unit in Modelica is an attribute of the instance that can optionally be used to do unit checking but has no impact on the behavioral semantics. The conclusion is that a unit is not tight to the type. In this sense is a record a type template.

Values assigned in Modelica are always considered as SI units.
```
Modelica.SIUnits.Length x1(unit="cm", displayUnit="mm") = 1; // has the value of 1 cm, computed as 1, displayed as 10
Modelica.SIUnits.Length x2(unit="m", displayUnit="mm") = 1;   // has the value of 1 m, computed as 1, displayed as 1000
```

How is this handled by functions?
- Functions check the passed value and check whether the unit is the same.
- unit checking is not standardized.

Allowing modifications in component declarations:
* improves readability
* requires the tools to be able to merge modifications, which is not overly complicated.

## Conclusion:
*  allow modification of built-in types, e.g. to declare a type
*  allow modifications on any component declarations (components in records, in functions, on top level)

olivleno:
Web-Meeting:

How to handle different modification styles?
```
Point p1(x(start=1.0,nominal=1.0));
Point p2(x.start=1.0,x.nominal=1.0);
Point p3(x(start=1.0),x.nominal=1.0);
```
Problem with dot notation is that additional checking for potential duplicates is required.
It's not super complicated to implement, but there is no need supporting dot notation without loss of functionality. 

Decision:
Restrict Flat Modelica to paranthesis notation only, disallowing dot notation.

Handling of constant and non-constant modifications?
```
// Full Modelica
record Point
  Modelica.SIUnits.Length x
  Modelica.SIUnits.Length y
record end;
Point p;

// Flat Modelica allowing basic types
constant Real 'Modelica.Math.Constants.pi' = 3.1415;
type 'Modelica.SIUnits.Length' = Real (unit="m", nominal='Modelica.Math.Constants.pi');  // only allowed for built-in scalar type
record Point
  'Modelica.SIUnits.Length' x(unit="cm");  // constant modifications shall be allowed
  'Modelica.SIUnits.Length' y(start=x);       // non-costant modifications shall be allowed only at top level component
record end;
parameter Point p1(x(unit="cm"));  // allowed as modification of top level component (is no longer an 'Modelica.SIUnits.Length'!)
Point p2(y(start=p1.x ));  // any modification allowed without restrictions 
parameter Point p2(y(start=p2.x ));  // crazy, but allowed and useful to avoid non-constant modifications in types
```

Conclusion:
At the top level no restrictions apply to modification variability.

How to handle of modification in arrays
```
Real x[2](start={4,5}, fixed={false,true});
```
This is needed for top level component declarations.
There would be a work around to apply modifications at the top level. Hence I should not be allowed as it increases the complexity of the internal representation.

How to express modifications of array element types? (shared by all elements)
Use of "each"
```
record Point
  Real x[2](each start=4, each fixed=false);
end Point; 
```
```
type P = Real[3];
record Point
  P x[2](each start={4,4,4}, each fixed={false,false,false});  // "each" cannot avoid the need for defining three value required to be       equal
  P y[2](start=4, fixed=false);  // disallow "each" but always considered all to be the same, as syntactic sugar for the below.
end Point; 

type  Real2 = Real(start=4, fixed=false);
type P2 = Real2[3];
record Point2
  P2 x[2];
end Point; 
```

Open question to be continued in next meeting.

dariomangoni:
Has any decision been made about the use of "each"?
Is it allowed to set `P y[3](fixed=true, start={0,0,0})` without the need to specify `each fixed` but just by assuming that `fixed=true` applies for all the components?

henrikt-ma:
> Has any decision been made about the use of "each"?
> Is it allowed to set `P y[3](fixed=true, start={0,0,0})` without the need to specify `each fixed` but just by assuming that `fixed=true` applies for all the components?

No; no decision so far.  Are you aware that this issue is only related to the MCP work on Flat Modelica (as indicated by the _MCP0031_ label, and the fact that the target of the PR is an MCP branch rather than _master_), and that any decision wouldn't apply to full Modelica?

dariomangoni:
Oh sorry, no I didn't get what MCP is.
My question was formulated to understand what is the correct behaviour for `each fixed`, since different compilers throw warnings when `fixed` is applied to array with missing `each` and I didn't find an answer on the specifications.

henrikt-ma:
> Oh sorry, no I didn't get what MCP is.
> My question was formulated to understand what is the correct behaviour for `each fixed`, since different compilers throw warnings when `fixed` is applied to array with missing `each` and I didn't find an answer on the specifications.

No problem.  Then this seems more relevant to you: https://github.com/modelica/ModelicaSpecification/issues/2630

olivleno:
How to name component declarations that directly belong to the flat modelica `model` (not like records, functions, type aliases)?
1. model component declaration
2. top level component declaration, used so far but not a good name

Decision:
Agreed on option 1: update accordingly [Henrik]

Decision:
Related changes in differences.md accepted.

Open Issues:
- modifications applied to component declarations in functions
- modifications of values
- handling of final

# modifications applied to component declarations in functions
Proposal:
Modifications in functions component declarations follow the same rules as in records.

Example:
```
function Foo
  input Real u(min=0);
  input Integer n(min=1);
  output Real y(min=u);                 //non-constant modification
protected
  Real x1(min=1, max=1);
  Real x2[n](min=linspace(0,1,n));  //non-constant modification
  Real k1(min=1, max=u);              //non-constant modification
  Real k2[n](min=linspace(0,u,n));  //non-constant modification
end Foo;
```
Should the restrictions apply?
- Good reason for non-constant modifier
- Good reason for non-equal array element types
  No need for different start attributes of array elements.
  Could be useful for min, max attributes to have individual values.
- Array dimensions can depend on the dimension of the inputs.
- Need for attributes in functions at all?
  unit: needed for unit checking
  min, max: could be useful
  start: not used, because 'assign before use' required for functions
  fixed, unbounded, nominal, stateSelect: not needed
- Array dimensions can depend on the values of the inputs.

Kai: For embedded applications dynamic memory allocation is not allowed.
Henrik: Dynamic sizing could easily be defined as unsupported feature of eFMI.

Conclusion:
- Some attributes are needed.
- No restriction to constant modifiers for inputs and outputs will make handling of function types very complicated.
- Restriction to constant modifiers for protected variables: 
  not needed for unit, 
  possible work around using assert for min, max
- restriction to equal element types does no harm

Proposal:
1. Same restrictions apply to functions as for types. votes: Henrik, Gerd, Hans, Martin
2. -

Decision:
For function component declarations the same restrictions shall apply as for record component declarations.

To Do:
- Apply this change to differences.md [Henrik]

Remaining Open Issues for next meeting: Fr. Oct. 9, 10.00-11.30
- modifications of values
- handling of final

olivleno:
Web Meeting

### Examples for Modifications of values

#### Should references in declaration equations of types be avoided?
```
record R
  constant Real c=1.0;
  parameter Real pc=2*c;   // will modify the default of pc when c is modified, adds significant complexity
end R;
```
```
record R
  Real x=1.0;
  Real xx=2*x;   // even worse as 2*x is non-constant
end R;

R r=R();   // using the record constructor overwriting the declaration equation is rather easy to handle

record R2 = R(x=2);
R r2=R2();   //call record constructor of R2, which is derived from record constructor of R
```
In Flat Modelica the scope of the short class declaration (see R2) is restricted such that one can only find constants.

Use cases of having declaration equations in records
* 

#### Handle package constants
Possible variants to capture package constants at the top level, in record or model:
```
  record R
    Real x=M.'Modelica.Constants.pi';
  end R;
  model M
    constant Real 'Modelica.Constants.pi' = 3.14;  //will be available in the results
    Real x = 2 * 'Modelica.Constants.pi';
  end M;

  constant Real 'Modelica.Constants.pi' = 3.14;  // not allowed in Modelica
  record R
    Real x='Modelica.Constants.pi';
  end R;
  model M
    Real x = 2 * 'Modelica.Constants.pi';
  end M;

  record 'Modelica.Constants'
      constant Real pi=3.14;
  end 'Modelica.Constants';
  record R
    Real x='Modelica.Constants'.pi;
  end R;
  model M
    Real x = 2 * 'Modelica.Constants'.pi;
  end M;
```
To be discussed further in next meeting.

#### Constant value modifications of a type allowed?
```
record R
  Real x=1.0;
  constant Real c=1.0;
  parameter Real p=1.0;
  parameter Real p2=2*c;   // will modify the default of p2 when c is modified, adds significant complexity
end R;

R r_const(
  x=2, 
  c=3,
  p=4);

R r_var(
  x=2 * r_var.c * time,        // could be turned into an equation
  c=3,
  p=4);

R r_eq(
  c=3,
  p=4);
equation
 r_eq.x =2 * r_eq.c * time;  // this requires the modifier of r_eq.x being removed
 r_eq.p2=2* r_eq.c;            // does not appreciate that p2 is not final, strictly speaking this can no longer be modified
```
For constants we require to have a declaration equation.
Should this apply to constants in records also?

Record constructors rely on the declaration equations for building the constructor function

henrikt-ma:
Option 3 doesn't look quite as bad if the Flat Modelica code is presented like this:
```
constant Real 'n_' = 5; // Value of constant for global access.
record R
  Real x(nominal = 'n_');  // Can't refer directly to M.'n', so using global constant instead.
end R;
model M
  constant Real 'n' = 'n_'; // Mainly included for appearance in simulation result.
M end;
```

olivleno:
Web Meeting:

Option 3 "Allow constants in global scope"
* pro: Improves readability allowing to keep constants and types outside the model that do not originate from inside the model.
  * e.g. indentation
  * easier to find components declarations and equation
* con: requires duplicates in the model to make them part of the results.
* con: change in grammar in terms of each mo file is expected to contain only one definition
  * This 

Option 2 "Package constant and record being defined inside model".
* pro: no change to grammar
* con: require "hide results" annotation
* con: model becomes a mix of constants from different sources (package)

Mix of Option 2 and 3:  "Allow constants in global scope and allow types in the model"

Conclusion
* Move this discussion into a separate pull request. [Henrik]
* Make a preliminary decision beginning of next meeting but then move on quickly.
* Open a new pull request for value modifications [Henrik]

henrikt-ma:
> Web Meeting:
> approved

I've forgot to merge those changes.  Doing it now, but only by merging the branch, as there are still unfinished topics in this PR that aren't covered by new subtopic PRs.

henrikt-ma:
New subtopic PRs:
- Simplify value modification: https://github.com/modelica/ModelicaSpecification/pull/2747
- Make constants available to type definitions: https://github.com/modelica/ModelicaSpecification/pull/2746

Unresolved subtopic without PR:
- Final modification

olivleno:
Web Meeting
Decision to open another PR for Final modification and close this branch. [henrik]

henrikt-ma:
New subtopic PR:
- Express final modification: #2748

With all unresolved subtopics having their own PRs now, we are now done with PR, and the branch will be deleted.

PR #2580, opened 2020-06-12
Make Figure title optional as decided at the 101th design meeting

otronarp:


chapters/annotations.tex
```diff
@@ -103,16 +103,17 @@ \subsection{Annotations for Figures}\doublelabel{annotations-for-figures}
...
-list of figures, and as it is generally hard for tools to automatically generate
-a meaningful title.  On the other hand, when the \lstinline!title! of a
-\lstinline!Plot! isn't provided, the tool produces a default, but the default is
-allowed to be empty.  Providing the empty string as \lstinline!title! of a
-\lstinline!Plot! means that no title should be shown.  The plot title is not to
-be confused with the plot \emph{label} which is never empty, see below. Variable
-replacements, as described in \autoref{variable-replacements}, can be used in the
-\lstinline!title! of \lstinline!Figure! and \lstinline!Plot!.
+Both \lstinline!Figure! and \lstinline!Plot! can have an optional title.
+Providing an empty string, or leaving the default of an empty string, for a
```
henrikt-ma, line 107: I think the pseudo code is usually enough to convey default behavior in a case like this.  Hence, one can simply write:
> When the `Figure` `title` is the empty string (the default), the tool must…

chapters/annotations.tex
```diff
@@ -103,16 +103,17 @@ \subsection{Annotations for Figures}\doublelabel{annotations-for-figures}
...
+\lstinline!Figure! \lstinline!title! means that the tool must produce a
+non-empty title based on the figure content.  On the other hand, when the
+\lstinline!title! of a \lstinline!Plot! isn't provided, the tool produces a
```
henrikt-ma, line 110: For symmetry, assuming the above was accpeted:
> … when the `Plot` `title` isn't provided, the tool…

Actually, a slightly bigger change might simplify specification a bit more:
> On the other hand, the `Plot` `title` has a tool-dependent default, but the default may be the empty string.  When the `Plot` `title` is the empty string, no title should be shown.

PR #2581, opened 2020-06-12
Clarify that leading and ending newlines are ignored in captions

otronarp:


chapters/annotations.tex
```diff
@@ -256,7 +256,8 @@ \subsubsection{Text Markup in Captions}\doublelabel{text-markup-in-captions}
 
 A sequence of one or more newlines (encoded either literally or using the \lstinline!\n!
 escape sequence) means a paragraph break.  (A line break within a paragraph is
-not supported.)
+not supported.)  A sequence of newlines before the first paragraph or after the last
```
maltelenz, line 259: If all newlines are ignored, maybe you could say "Any newlines before the..."? This could be interpreted in a way where you need multiple newlines to ignore them.

otronarp: That's indeed better, I have updated the PR.

chapters/annotations.tex
```diff
@@ -256,7 +256,8 @@ \subsubsection{Text Markup in Captions}\doublelabel{text-markup-in-captions}
 
 A sequence of one or more newlines (encoded either literally or using the \lstinline!\n!
 escape sequence) means a paragraph break.  (A line break within a paragraph is
-not supported.)
+not supported.)  Any newlines before the first paragraph or after the last paragraph are
+ignored.
```
henrikt-ma, line 260: Since we've already defined such a sequence to be some sort of paragraph break, I think we should stick to it:
> (… is not supported, and any paragraph break before the first paragraph or after the last paragraph has no impact.)

PR #2601, opened 2020-06-24
Improve and make use of 'definition' environment

### LINKED ISSUE #2570: Too many kinds of definitions
We should make all definitions use a small number of styles, but currently there is a mix of numbered ones, tables with `\firstuse`, `description` lists, and `\firstuse` placed in normal paragraphs.

The numbered ones are the ones that stand out the worst.  These are the ones:
- _Local Number of Unknowns_
- _Local Equation Size_
- _Locally Balanced_
- _Globally Balanced_
- _Plug-compatibility (= restricted subtyping)_
- _Default connectable_
- _Function-Compatibility or Function-Subtyping for Functions_

Should more definitions be numbered, or should we get rid of the numbering of these in favor of `description` lists or simply `\firstuse`?

henrikt-ma:
Fixes #2570.

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -1890,26 +1890,27 @@ encrypted and used with several different Modelica tools, using
 different encryption formats.
 \end{nonnormative}
 
-Definitions:
-\begin{longtable}[]{|p{2.5cm}|p{12cm}|}
-\hline
-\tablehead{Term} & \tablehead{Description}\\ \hline
-\endhead
-\firstuse{Protection} & Define what parts of a class are visible.\\
-\hline
-\firstuse{Obfuscation} & Changing a Modelica class or generated code so that it is difficult to inspect by a user
-(e.g.\ by automatically renaming variables to non-meaningful names).\\
-\hline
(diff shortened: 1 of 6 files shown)


henrikt-ma:
> * The style of "[The input argument is a] component expression" looks a bit odd, but I don't know how to  best format that - maybe it was better without []?. Clearly "component expression" is the main part, but the definition above already defined "component expression".

Yes, it is really hard to make this nice.  I think the real problem here is that the the terms being defined are just too long because of the need to also give that context (_the input argument is a_) as part of the definition.  I think one can take a step back here, and see that we aren't actually defining three new concepts here, just putting already defined concepts in a particular context.  Let's see what that would look like…

henrikt-ma:
> * The figures for piecewise-constant (etc) look a bit odd with the new format. And 16.3 seems redundant with 16.4. One possibility would be to replace those three figures by one larger figure containing everything; and "hold(r)" for the green dotted line.

I'm working on it, but it takes time to find a solution that works with both `\autoref` and LaTeXML…

PR #2603, opened 2020-06-25
Initialization of variables in Flat Modelica using start('v') and guess('v')

henrikt-ma:


henrikt-ma:
Aborting this PR, since the branch to be merged should have been in the central repository.

PR #2611, opened 2020-07-08
Use cleveref.sty for cross references

henrikt-ma:
The _cleveref.sty_ package provides better alternatives to `\autoref`, most importantly allowing cross references to be placed at the beginning of a sentence without forcing all link text to start with an uppercase letter.  It also unifies the formatting of all cross references using a single but flexible mechanism, and is supported by LaTeXML.

### DIFF
--- a/.CI/latexml/Dockerfile
+++ b/.CI/latexml/Dockerfile
@@ -23,7 +23,7 @@ RUN sed -i s,http://archive.ubuntu.com/ubuntu/,mirror://mirrors.ubuntu.com/mirro
  && $INSTALL/install-tl --profile texlive.profile \
  && rm texlive.profile \
  && tlmgr update --self --all --reinstall-forcibly-removed \
- && tlmgr install latexmk listings xcolor float multirow tocloft parskip etoolbox \
+ && tlmgr install latexmk listings cleveref xcolor float multirow tocloft parskip etoolbox \
  && cpanm JSON \
  && cpanm LaTeXML \
  && apt-get autoremove -qy make gcc curl wget git cpanminus libxml2-dev libxslt-dev \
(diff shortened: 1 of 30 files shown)


henrikt-ma:
> And the changes in DAE (and derivation of stream) does not seem to work with LaTeXML; could be that I have the wrong version - but I get creftype in the generated HTML.
> I would thus propose to not use it for equations yet; the benefit is so minor.

Good catch!  I've reported it here: https://github.com/brucemiller/LaTeXML/issues/1306

henrikt-ma:
> I would thus propose to not use it for equations yet; the benefit is so minor.

Fixed.

henrikt-ma:
> The Continuous Integration fails similarly as another PR using cleveref.

I haven't edited this file before, but just adding `cleveref` to the list of packages didn't work.  Should it be sufficient to make this change on the PR branch?  Anything else one needs to do in order to actually get the package installed?

henrikt-ma:
@sjoelund, do you have any suggestions for what else to try in order to get _cleveref.sty_ installed for the CI builds?  As of #2622 being merged into _master_, the _.CI/latexml/Dockerfile_ has been modified in the same way for _cleveref.sty_ as for other LaTeX packages we are using, but it hasn't had any effect so far.

sjoelund:
> @sjoelund, do you have any suggestions for what else to try in order to get _cleveref.sty_ installed for the CI builds? As of #2622 being merged into _master_, the _.CI/latexml/Dockerfile_ has been modified in the same way for _cleveref.sty_ as for other LaTeX packages we are using, but it hasn't had any effect so far.

You need to build, tag, and upload that docker image. And then modify the Jenkinsfile to use this docker image. As I'm back from vacation, I can do it.

sjoelund:
It builds fine now

henrikt-ma:
> It builds fine now

Great, thanks!

PR #2626, opened 2020-08-03
Build and use the new docker image with cleveref

sjoelund:


sjoelund:
I can document it in a new file in this PR

PR #2628, opened 2020-08-04
Clean up use of hyphens and dashes

henrikt-ma:
Now that we are in the LaTeX world, there are pretty clear guidelines for when to use hyphens, en-dash, and em-dash.  This PR tries to do it right.

### DIFF
--- a/.CI/Jenkinsfile
+++ b/.CI/Jenkinsfile
@@ -18,7 +18,7 @@ pipeline {
   stage('build') {
     agent {
       docker {
-        image 'modelicaspec/latexml:20200415'
+        image 'modelicaspec/latexml:20200803'
         label 'linux'
         alwaysPull true
       }
--- a/.CI/latexml/Dockerfile
+++ b/.CI/latexml/Dockerfile
@@ -23,7 +23,7 @@ RUN sed -i s,http://archive.ubuntu.com/ubuntu/,mirror://mirrors.ubuntu.com/mirro
  && $INSTALL/install-tl --profile texlive.profile \
  && rm texlive.profile \
  && tlmgr update --self --all --reinstall-forcibly-removed \
- && tlmgr install latexmk listings xcolor float multirow tocloft parskip etoolbox \
+ && tlmgr install latexmk listings cleveref xcolor float multirow tocloft parskip etoolbox \
  && cpanm JSON \
  && cpanm LaTeXML \
  && apt-get autoremove -qy make gcc curl wget git cpanminus libxml2-dev libxslt-dev \
--- a/.CI/latexml/Dockerfile.incremental
+++ b/.CI/latexml/Dockerfile.incremental
@@ -1,0 +1,5 @@
+FROM modelicaspec/latexml:20200415
+
+RUN tlmgr update --self \
+ && tlmgr install cleveref \
+ && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/ /root/.cpanm/work*
--- a/.CI/latexml/README.md
+++ b/.CI/latexml/README.md
@@ -1,0 +1,24 @@
+# Updating the docker image
+
+The docker image is stored at the [docker hub](https://hub.docker.com/r/modelicaspec/latexml).
+If you need write access to the organization, contact @sjoelund.
+
+To update the image, update the Dockerfile .
+Then build and tag the image and upload it.
+
+```sh
+docker build -t modelicaspec/latexml:`date +%Y%m%d` .
+docker push modelicaspec/latexml:`date +%Y%m%d`
+```
+
+If the change is small you can base it on the old image instead of creating a new one from scratch.
+Modify Dockerfile.incremental to be based on the image you want to update.
+Make sure that the regular Dockerfile is also updated (so future builds incorporates these additions).
+
+```sh
+docker build -t modelicaspec/latexml:`date +%Y%m%d` - < Dockerfile.incremental
+docker push modelicaspec/latexml:`date +%Y%m%d`
+```
+
+Once the docker image is updated, modify `../Jenkinsfile` to use this image.
+Put all of this in the same commit and see if the CI build accepts it.
--- a/chapters/abstract.tex
+++ b/chapters/abstract.tex
@@ -47,7 +47,7 @@ taken in the preparation of this document no responsibility for errors
 or omissions is assumed.
 
 The contributors to this and to previous versions of this document are
-listed in \autoref{modelica-revision-history}. All contributors worked voluntarily and without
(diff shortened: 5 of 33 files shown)


HansOlsson:
I think that if we are following a style-guide we should try to add a link to it.

However, my main concern is that many of the changes were replacing en-dash by em-dash, and I think we shouldn't do that.

Why:
When I searched I found  https://en.wikipedia.org/wiki/Dash - which is subtly different.
It indicates that one can use unspaced em dash or spaced en-dash as a break (and a pair as parenthesis) - and, of course, there are different variants.

However, this PR contains spaced em-dash, which seems less common - and some directly argue against. A more esoteric variant is thin-spaced em-dash that Chicago style manuals allegedly uses  `\,---\,` https://tex.stackexchange.com/questions/53413/what-is-the-latex-command-for-em-dash Some claim that newspapers use that, but looking at The Economist I found non-spaced em-dash, Scientific American non-spaced em-dash, and Wired seemed to use spaced en-dash.

For combined readability of both source and generated text I think that the current spaced en-dash is the best; as the Latex source for unspaced em-dash looks odd, and thinly spaced em-dash is too difficult to remember.

HansOlsson:
And I noticed that D. Knuth in The TeXbook uses a non-spaced em-dash, when explaining how it is used.

henrikt-ma:
> As indicated I don't like the spaced em-dashes.

Yes, that spacing might be a regional thing; at least in Sweden I think spaces around em-dash is the dominant style, but I've also seen the recommendation to not use spaces in other countries.  I was even considering introducing an `\emdash` macro, so that spacing can be controlled in a single place.

I just learnt about the "AP style" (https://www.apstylebook.com/), where spaces are recommended.

To me, this sounds like something we could poll on, but in the mean time I'll remove the spaces, as they would be trivial to put back considering that `\b---\b` will match every em-dash we have, and only those.

HansOlsson:
> > As indicated I don't like the spaced em-dashes.
> 
> Yes, that spacing might be a regional thing; at least in Sweden I think spaces around em-dash is the dominant style, but I've also seen the recommendation to not use spaces in other countries. 

Are you sure?

https://sv.wikipedia.org/wiki/L%C3%A5ngt_tankstreck

Claims that em-dash isn't used in contemporary Swedish (for that purpose - but used as a repetition-marker), instead spaced en-dash is used. 

However, the length of the dashes isn't that standardized between languages so it might be that a Swedish normal dash (typically with spaces around) is a bit longer than a typical English en-dash and thus easily confused with a spaced em-dash.

I was even considering introducing an `\emdash` macro, so that spacing can be controlled in a single place.

henrikt-ma:
> Are you sure?
> 
> https://sv.wikipedia.org/wiki/L%C3%A5ngt_tankstreck
> 
> Claims that em-dash isn't used in contemporary Swedish (for that purpose - but used as a repetition-marker), instead spaced en-dash is used.

Not any more.  Seems I've mixed up the Swedish style of putting spaces with the English style of dash length. I can't say I really recognize the short length of an en-dash as the _tankestreck_/_talstreck_ often used in Swedish, but now that I know about it I suppose I will start seeing it everywhere…

HansOlsson:
> To me, this sounds like something we could poll on, but in the mean time I'll remove the spaces, as they would be trivial to put back using some regexp matching.

To me it seems simpler to restore it to spaced en-dash - it should be as simple to change later.

henrikt-ma:
> To me it seems simpler to restore it to spaced en-dash - it should be as simple to change later.

Of course.  (This only shows that the idea of an `\emdash` command wouldn't have been the correct solution, as that puts focus on the character, not the semantics.)

henrikt-ma:
Another observation regarding spaces around dashes: Some fonts have spacing around the em-dash.

When I look at the following on GitHub/Chrome/macOS Catalina, things look as expected, and I'm somewhat disturbed by the lack of space around the em-dash:
> M-M
> M–M
> M—M

I'm not sure what font I'm looking at, but this is how it looks:
dashes-github-chrome

However, if I try the same in a document (TextEdit.app or Mail.app) using Helvetica, there is a nice little space around the em-dash:
dashes-helvetica-macos

This shows that besides style guidelines, there is also a font consideration giving even more reason to not add (additional) space around the em-dash; one should leave it up to the font designer to decide the amount of space around it.

HansOlsson:
> Another observation regarding spaces around dashes: Some fonts have spacing around the em-dash.

Ok, so it's a bigger mess. 

That's why I think we should stick with spaced **en**-dash for the moment; as it seems we primarily use that currently and thus it minimizes the current changes until we determine if there is a better solution.
I also find that it is more readable in the source.

Obviously there is also unspaced en-dash for other purposes, e.g., 2--3; but they are simple to tell apart.

henrikt-ma:
> That's why I think we should stick with spaced **en**-dash for the moment; as it seems we primarily use that currently and thus it minimizes the current changes until we determine if there is a better solution.

Sorry, I didn't understand that this was a change you wanted already now.  Fixed.

> I also find that it is more readable in the source.

Indeed.  (If we can avoid having them next to a newline in the source, it will be even easier to find them all in case we'd like to change them later.)

> Obviously there is also unspaced en-dash for other purposes, e.g., 2--3; but they are simple to tell apart.

Yes, those are a different story, and there seems to be broad agreement on how they should be used in this case.  (One of these uses that was new to me was the concatenation of author names in theorem names, etc: _Bolzano–Weierstrass theorem_)

henrikt-ma:
When scanning our code for missed cases of closed en-dashes to concatenate names, I found this one: _TU Hamburg-Harburg_.  To me, it looks very much like _Bolzano–Weierstrass_, but it was hard to find good sources in English to see how the name is written…

HansOlsson:
> When scanning our code for missed cases of closed en-dashes to concatenate names, I found this one: _TU Hamburg-Harburg_. To me, it looks very much like _Bolzano–Weierstrass_, but it was hard to find good sources in English to see how the name is written…

One option is "Hamburg University of Technology" :-)  and in German it seems to be renamed to "Technische Universität Hamburg" eliminating the problem - but keeping the abbreviation. Or I might totally misunderstand it.

henrikt-ma:
> > When scanning our code for missed cases of closed en-dashes to concatenate names, I found this one: _TU Hamburg-Harburg_. To me, it looks very much like _Bolzano–Weierstrass_, but it was hard to find good sources in English to see how the name is written…
> 
> One option is "Hamburg University of Technology" :-) and in German it seems to be renamed to "Technische Universität Hamburg" eliminating the problem - but keeping the abbreviation. Or I might totally misunderstand it.
> Ah, https://de.wikipedia.org/wiki/Technische_Universit%C3%A4t_Hamburg says that it was renamed in 2018.

OK, but that means that affiliations dating before 2018 should probably still be written with the old name.

PR #2632, opened 2020-08-10
Define meaning of empty Axis.unit

henrikt-ma:
This PR tries to address the problem that it is currently not allowed to leave the `unit` of an `Axis` empty, which in turn must be interpreted as also the axis specifications `x` and `y` of a `Plot` being mandatory.  I don't think it was intended to add the burden of specifying axes for every `Plot`, and specifying that it means to leave the `unit` empty opens up for an interesting feature currently not present in the MCP.

This PR is a follow-up to #2579.

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -169,6 +169,9 @@ end Axis;
 
 When an axis bound isn't provided, the tool computes one automatically.  When \lstinline!min! or \lstinline!max! is specified, it is an error to leave \lstinline!unit! empty.
 
+An empty \lstinline!unit! means that each expression plotted against it uses its own \lstinline!displayUnit!.  It is not required that the different expressions have compatible units; instead a unitless
+axis may be used, with unit information attached to curve legends instead.
+
 If a tool does not recognize the \lstinline!unit!, it is recommended to issue a warning and treat the \lstinline!unit! as if it was empty.
 
 When an axis label isn't provided, the tool produces a default label.  Providing the empty string as axis label means that no label should be shown.  Variable replacements, as described in
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -169,16 +169,12 @@ end Axis;
 
 When an axis bound isn't provided, the tool computes one automatically.
 
-The Modelica tool is responsible for showing the unit used for values at the
-axis tick marks, so the axis \lstinline!label! shall not contain the unit.
-
 If a tool does not recognize the \lstinline!unit!, it is recommended to issue a
 warning and treat the \lstinline!unit! as if it was not specified.
 
-When an axis label isn't provided, the tool produces a default label. Providing
-the empty string as axis label means that no label should be shown. Variable
-replacements, as described in \cref{variable-replacements}, can be used in
-the \lstinline!label! of \lstinline!Axis!
+When an axis label isn't provided, the tool produces a default label.  Providing the empty string as axis label means that no label should be shown.  Variable replacements, as described in
+\cref{variable-replacements}, can be used in the \lstinline!label! of \lstinline!Axis!  The Modelica tool is responsible for showing the unit used for values at the axis tick marks, so the axis
+\lstinline!label! shall not contain the unit.
 
 \subsubsection{Plot Curves}\label{plot-curves}
 The actual data to plot is specified in the \lstinline!curves! of a \lstinline!Plot!:
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -177,7 +177,7 @@ When \lstinline!unit! is empty, and axis bounds are to be determined automatical
 user, on the other hand, a tool may choose a unit for the variable such that the range of the variable values (expressed in the chosen unit) fit nicely with the range of the unitless axis.
 \end{nonnormative}
 
-If a tool does not recognize the \lstinline!unit!, it is recommended to issue a warning and treat the \lstinline!unit! as if it was empty.
+If a tool does not recognize the \lstinline!unit!, it is recommended to issue a warning and treat the \lstinline!unit! as if it was empty, as well as ignore any setting for \lstinline!min! and \lstinline!max!.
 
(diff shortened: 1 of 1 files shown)


chapters/annotations.tex
```diff
@@ -162,23 +162,26 @@ \subsubsection{Axis Properties}\label{axis-properties}
...
-The Modelica tool is responsible for showing the unit used for values at the
-axis tick marks, so the axis \lstinline!label! shall not contain the unit.
+An empty \lstinline!unit! means that the axis is unitless, and each expression plotted against it may use its own unit determined by the tool.  The tool is responsible for conveying the information
+about choice of unit for the different variables, for instance by attaching this information to curve legends.
 
-If a tool does not recognize the \lstinline!unit!, it is recommended to issue a
-warning and treat the \lstinline!unit! as if it was not specified.
+\begin{nonnormative}
+When \lstinline!unit! is empty, and axis bounds are to be determined automatically, a natural choice of unit could be the variable's \lstinline!displayUnit!.  When axis bounds are specified by the
+user, on the other hand, a tool may choose a unit for the variable such that the range of the variable values (expressed in the chosen unit) fit nicely with the range of the unitless axis.
```
henrikt-ma, line 177: Quoting @HansOlsson regarding these lines, so that we get a subthread for discussion:
> The following: "When axis bounds are specified by the user, on the other hand, a tool may choose a unit for the variable such that the range of the variable values (expressed in the chosen unit) fit nicely with the range of the unitless axis."
to me suggests that one can give min/max without unit and have the unit adapt to that in some way.
>
> I may be misinterpreting it, but it should at least be clear that min/max must be given in the unit and if the unit is missing or ignored so are min/max.

henrikt-ma: > to me suggests that one can give min/max without unit and have the unit adapt to that in some way.

No.  The _of the unitless axis_ at the end of the sentence is there to make it clear that the axis still doesn't have a unit; the units are always associated with the variables when `unit` is empty.  It is true that it becomes up to the tool to make a clever choice of unit when `min` and `max` are given, but `unit` is empty; a bad choice will result in a poor fit between the variable values and the axis range.

> I may be misinterpreting it, but it should at least be clear that min/max must be given in the unit and if the unit is missing or ignored so are min/max.

No.  A bad unit is an error, but there is still a suggested way to deal with it.  More importantly, `min` and `max` do have a well defined meaning when `unit` is empty, as described above.  That `min` and `max` are given in `unit` is expressed by the pseudocode record, expressed in one way in this PR, but more clearly in #2633:
```
record Axis
  Real min "Axis lower bound, in 'unit'";
  Real max "Axis upper bound, in 'unit'";
  String unit "Unit of axis tick labels";
  String label "Axis label";
end Axis;
```

HansOlsson: My suggestion is that if we don't recognize the unit, it's incorrect, or if it is empty we ignore the min and max as well.

The tools must be able to handle missing unit and min/max - so just fall back to that instead of having advanced logic to handle such problems. It seems simpler and gives more predictable result, and it encourages users to either specify the plots in detail - or not at all.

henrikt-ma: I don't see the reason to ignore `min` and `max` when `unit` is empty.  The possibility to have variables of different units on the same axis is quite useful, and you get this mode by using an empty `unit`.  In this mode, it is still perfectly meaningful to specify `min` and `max` boundaries for the unitless axis.

Something I think we could clarify, however, is the recommendation for what to do when `unit` is not recognized: Then, the `min` and `max` should be ignored (instead of being applied to the unitless axis) — I'll make a separate comment for this.

chapters/annotations.tex
```diff
@@ -162,23 +162,26 @@ \subsubsection{Axis Properties}\label{axis-properties}
...
+\end{nonnormative}
+
+If a tool does not recognize the \lstinline!unit!, it is recommended to issue a warning and treat the \lstinline!unit! as if it was empty.
```
henrikt-ma, line 180: Suggestion that came out of the discussion above:
```suggestion
If a tool does not recognize the \lstinline!unit!, it is recommended to issue a warning and treat the \lstinline!unit! as if it was empty, as well as ignore any setting for \lstinline!min! and \lstinline!max!.
```

chapters/annotations.tex
henrikt-ma: Here's a plot with unitless axis combined with explicit setting of `min` and `max` to get 0 vertically centered:
unitless-axis-with-min-max

HansOlsson: If we want to plot variables in different units with given min/max the most reasonable would be to say that they use the given unit (or displayUnit) of the variables; nothing fancy. If I for the curve above gave min=0, max=0.4 it's not clear if the tool should use "V" or "kV" - and to me it seems as if the specification recommends "kV" (and "kW") since more of the curve is shown in that case.
The other option would be something a lot more advanced.

chapters/annotations.tex
HansOlsson: Agreed.

chapters/annotations.tex
henrikt-ma: Your interpretation of the recommendation is in line with the intention, and I think we need to stick to mappings to the axis range that can be expressed by the choice of unit.

I don't like to for the plot to use the variable's `displayUnit`, since the `displayUnit` was probably set with just the variable at hand in mind, not how the range of the variable would match the ranges of some other variable when added to the same unitless plot axis.  If you don't like the tool-dependent choice of unit here, the compromise that comes to mind is to extend `Curve` with (optional) `xDisplayUnit` and `yDisplayUnit`, which would default to the `displayUnit` of the `x` and `y`, and then require that `xDisplayUnit` and `yDisplayUnit` are used on unitless axes.

otronarp: > If we want to plot variables in different units with given min/max the most reasonable would be to say that they use the given unit (or displayUnit) of the variables; nothing fancy. 

But the wording doesn't require you to do anything fancy, it only gives you enough leeway to do fancy stuff if you want to.

PR #2637, opened 2020-08-21
When speaking of when-equations, don't say "when-statements"

henrikt-ma:
(For when-statements, we already have the corresponding rule listed in 11.2.7.1 _Restrictions on When-Statements_.)


HansOlsson (approved):
Looks good to me.

PR #2666, opened 2020-09-16
Remove controversial part of merged with #2646

henrikt-ma:
In the phone meeting leading to the merging of #2646, this controversial part was not discussed.  A new issue will be created for putting it back.


HansOlsson (approved):
Ok.

PR #2687, opened 2020-10-11
Fix parskip.sty style paragraphs in LaTeXML build

henrikt-ma:
To have non-_parskip.sty_ style paragraphs looks even worse in HTML than in a PDF document, and I assume the reason it looks the way it does is just because of poor support for _parskip.sty_ in LaTeXML.


HansOlsson (approved):
Looks good.

PR #2694, opened 2020-10-28
Variability in record member declaration

henrikt-ma:
From the current placeholder text under the section _Variability in record member declaration_ in _differences.md_:

> The goal is to get rid of one or both of the variability prefixes `constant` and `parameter` from record member declarations.
>
> For `constant`, this is of particular interest due to the way we have restricted the attribute modifiers to only allow constant modifiers.  For value modification, this would make the most sense for — again — constant modifiers, but that seems to be mostly useful for record members declared `constant`.  If we would get rid of the `constant` modifier inside record definitions, this might allow us to conclude that we don't really need value modifiers in records at all.
>
> For `parameter`, we should keep in mind that many of the members declared `parameter` in full Modelica are actually structural parameters that will turn into `constant` in Flat Modelica.  Hence, we need to keep those structural parameters in mind when thinking about getting rid of `constant`.  Then, we must also consider the impact of removing the possibility to declara non-structural members with `parameter` variability.

### DIFF
--- a/RationaleMCP/0031/differences.md
+++ b/RationaleMCP/0031/differences.md
@@ -79,13 +79,25 @@ Seen this way, the rules about which functions may be called in the body of a fu
 
 This covers what one can currently express in full Modelica.  In the future, one might also introduce _pure discrete_ functions that don't have side effects, but that must be re-evaluated at events, even if the arguments are constant.
 
-## Variability in record member declaration
+## Variability specification inside types
 
-This section is work in progress.  The goal is to get rid of one or both of the variability prefixes `constant` and `parameter` from record member declarations.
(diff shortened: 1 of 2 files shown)


olivleno:
# Web Meeting Oct. 28, 2020:

Analyse role of prefix constant and parameter in Full Modelica
 * parameter in record in connector 
   * Assert for connect equation to verify that parameters are consistent. => Not relevant for Flat Modelica.
* parameter in record passed to function
* parameter in record return of function
* parameter in record be part of equation
* parameters and constants in record to have both in the results.

To be discussed further.

gkurzbach:
record ExtrapolatingPolynomial
		extends Modelon.Icons.Record;
		parameter Integer n "order of polynomial";
		Boolean assertBounds "if true, throw assert at boundary";
		Real[n + 1] p "polynomial coefficients";

	function integralXP "Indefinite integral of extrapolating polynomial xp(u)"
		input ExtrapolatingPolynomial xp1 "polynomial record";
		output ExtrapolatingPolynomial xp2(n=size(xp1.p, 1)) "polynomial record";

PR #2695, opened 2020-10-30
Fix typos in spec

GallLeo:


henrikt-ma (approved):
Fantastic!  Did you do this manually, or was this all you found using a spellchecker?

HansOlsson (approved):
Great!

GallLeo:
> 
> 
> Fantastic! Did you do this manually, or was this all you found using a spellchecker?

@henrikt-ma This has been done manually and I did not catch everything, for sure. 
Workflow:
1. Export PDF as txt
2. Load in Microsoft Word
3. Scroll visually through pages
4. If issue found: edit .tex in text editor

Automization: 
I have a Modelica dictionary for aspell, but still it does not work nicely for the specficiation text, because there are strange words and costructs which would need some more training of aspell.
And, the Grammar check of Microsoft Word works quite fine. There seem to be missing many "," missing in the specfication text, e.g. "Therefore,", but I did not dare to change due to my lack of native english. But, I could do a pull request and some native speaker could check.

henrikt-ma:
> And, the Grammar check of Microsoft Word works quite fine. There seem to be many "," missing in the specfication text, e.g. "Therefore,", but I did not dare to change due to my lack of native english. But, I could do a pull request and some native speaker could check.

Yes, I've also noticed the many missing ",".  It would be good to fix them, but I suggest we wait until the massive #2627 has been merged to avoid merge conflicts.

henrikt-ma:
> I have a Modelica dictionary for aspell, but still it does not work nicely for the specficiation text, because there are strange words and costructs which would need some more training of aspell.

Sounds like an interesting starting point for something we might be able to put on Jenkins in the end…

PR #2698, opened 2020-11-02
Make assertion level a structural parameter

### LINKED ISSUE #2642: Can AssertionLevel change dynamically?
In `ModelicaCompliance.Equations.Assert.AssertVarLevel` we have:

```
model AssertVarLevel
  Real x(start = 0);
equation
  x = time;
  assert(x < 0.5, "This assert should abort the simulation at time 0.6.", if x > 0.6 then AssertionLevel.error else AssertionLevel.warning);
  annotation(__ModelicaAssociation(TestCase(shouldPass = false, section = {"8.3.7"})), experiment(StopTime = 1), Documentation(info = "<html>Checks that the assertion level can itself be dependent on a
        condition.</html>"));
end AssertVarLevel;
```

How should one read this test?

A: The `shouldPass = false` means this is against the specification, and a tool should reject it.
B: The documentation string and and the assert `message` describe what is expected to happen, and a tool should handle this model, issuing first a warning at `time = 0.5`, then an error at `time = 0.6`.

I would vote for A, and only allow literals in the `level` of an `assert`. The example at the end of 8.3.7 seems to indicate using two separate asserts for this use case is the recommended way.

If we allow B, what happens to the implicit `noEvent` in the warning case? Does it have to be dynamically switched off and on as the level changes? #2571 seems relevant as well.

henrikt-ma:
Fixes #2642

### DIFF
--- a/chapters/classes.tex
+++ b/chapters/classes.tex
@@ -524,11 +524,10 @@ equation
 \end{lstlisting}
 \end{example}
 
-The \emph{expression} must be a \lstinline!Boolean! scalar expression, and must be a
-parameter-expression.
+The \emph{expression} must be a \lstinline!Boolean! scalar expression, and must be a parameter expression.
 
 \begin{nonnormative}
-A parameter-expression is required since it shall be evaluated at compile time.
+A parameter expression is required since it shall be evaluated at compile time.
 \end{nonnormative}
 
 A redeclaration of a component shall not include a condition attribute;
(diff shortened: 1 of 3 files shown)


maltelenz (approved):
Looks good to me.
You probably mean that this fixes #2642 though, not #2692.

henrikt-ma:
> Looks good to me.
> You probably mean that this fixes #2642 though, not #2692.

Thanks!  It's the second time this has happened to me recently.  I'm wondering if it's something with the aggressive auto-completion on GitHub that's changing the issue numbers for me…

HansOlsson (approved):
Ok, I was first a bit surprised by the change of parameter-expression, but since we almost everywhere else use 'parameter expression' it makes sense to be consistent.

PR #2704, opened 2020-11-04
Fix some white-space irregularities in the grammar

henrikt-ma:
I was just trying to answer a question by referring to the grammar, and found these irregularities that were really making reading the grammar more difficult that it needs to be.


HansOlsson (approved):
Looks good. (I will work on merging things tomorrow.)

PR #2708, opened 2020-11-08
Fix lost white-space after \textregistered

henrikt-ma:


HansOlsson (approved):
Looks good, I notice that some suggest `\textsuperscript{\textregistered}` instead

henrikt-ma:
> Looks good, I notice that some suggest `\textsuperscript{\textregistered}` instead

Good idea.

preamble.tex
```diff
@@ -468,7 +468,7 @@
 \end{center}
 ~\\[2\baselineskip]
 \fi
-Modelica\textsuperscript\textregistered -- A Unified Object-Oriented Language for Systems
+Modelica\textsuperscript\textregistered{} -- A Unified Object-Oriented Language for Systems
```
henrikt-ma, line 471: As @HansOlsson mentioned in his comment, there's a better way:
```suggestion
Modelica\textsuperscript{\textregistered} -- A Unified Object-Oriented Language for Systems
```

HansOlsson (approved):

PR #2714, opened 2020-11-11
Update figures after change of syntax colors

henrikt-ma:
Figures have been regenerated with the new syntax colors.  To make this possible, the part of _preamble.tex_ that should also be used by figures has been factored out.


HansOlsson (approved):
Looks ok.

PR #2717, opened 2020-11-17
Point to correct CSS file in Jenkinsfile

henrikt-ma:
By the way, @sjoelund, can we get rid of old files that are no longer in use?


henrikt-ma:
In addition to the fixed proposed here, I'm afraid we'll have to do more to fix the HTML formatting of everything related to listings…

sjoelund:
I think I removed too many files from the server since there are css files in the root of the zip-file still

sjoelund (approved):

HansOlsson (approved):
Seems ok.

PR #2722, opened 2020-11-18
Better selection of font (including line-height) in navbar

henrikt-ma:
The previous use of the CSS 'font' property was implying a 'line-height' of 1.0, making the table of contents rather hard to read.

By specifying 'font-size' and 'line-height' instead, we can not ensure that the font face is the same sans-serif as set in .body, and we can make a compromise between readability and need for scrolling by setting a 'line-height' somewhere between 1.0 and the inherited 1.5.


HansOlsson (approved):
I think it looks ok.

PR #2744, opened 2020-11-30
Add biblatex, biber and index to the docker image

sjoelund:
#2740 #2726


henrikt-ma (approved):
Looks good, but the proof of the pudding is in the eating…

HansOlsson (approved):
Seems good.

PR #2750, opened 2020-12-04
Fix example in 12.4.6

perost:
- Change atan to atan2 in the example, since the number of arguments is wrong otherwise.


henrikt-ma (approved):
Looks good to me.

PR #2781, opened 2020-12-26
Fix issues related to 'declaration equation' in index

henrikt-ma:
This both fixes an error in the index for 'declaration equation' pointing to the deprecated 'declaration assignment', and further improves the index on the topic of declaration equations.

(I encountered this while answering to #2780.)


HansOlsson (approved):
Looks good.

PR #2782, opened 2020-12-26
Don't treat 'exit' as keyword in listings

henrikt-ma:
Maybe `exit` was a keyword in the beginning, before revision history was introduced?


HansOlsson (approved):
Looks good.

PR #2786, opened 2020-12-28
Say 'variable value' instead of 'value-attribute'

henrikt-ma:
While I can see how the term _value-attribute_ can be understood by looking at the definitions of the primitive types, the term isn't used anywhere else, and I think it is better if it is avoided here as well.

Suggestions for other ways of phrasing this are welcome.


HansOlsson (changes requested):
I would say 'value of the variable' instead.

henrikt-ma:
> I would say 'value of the variable' instead.

Sure.  Fixed.

HansOlsson (approved):
Looks good now.

PR #2799, opened 2021-01-07
Improve navbar on top page

henrikt-ma:
This fixes two issues with the navbar (table of contents, shown as left sidebar) on the top page:
- The document title shouldn't be present in the navbar (only in page header, where it happens to be missing on the top page).
- Entire table of contents has excessive indentation under the document title item.

I suggest that this is also cherry-picked to _maint/3.5_, as it has an impact on the first impression of the document.


HansOlsson (approved):
Good!

PR #2808, opened 2021-01-08
Add 'MODELICAPATH' to the index

henrikt-ma:
Even though I'm not a big fan of `MODELICAPATH` being defined in the specification, it is wrong to not be able to find it in the document index.


HansOlsson (approved):
Looks good.

PR #2824, opened 2021-01-25
Fix typos in revisions notes of v3.5

GallLeo:
I just fixed some typos in the release notes of Modelica 3.5. This should be also merged to master?

Btw, it would be good to use present tense "clarify" and and past tense "clarified", consistently. I did not fix this, because I did not know which option to choose.


HansOlsson (approved):
Looks good. Since the revisions are non-normative it seems safe to correct this even now.

PR #2831, opened 2021-01-26
Align duplicated parts of lexing rules

henrikt-ma:
This isn't the first time we have issues with these being out of sync.  Hopefully, maintenance will become easier now with the duplicated part being a substring of the full lexing rules.

> - [ ] Text does not adapt to width of browser window. I only found this for for Chapter 2 in one column mode on Desktop. Strange.

@HansOlsson: This is another PR where it would be good to have a documented procedure for also targeting the current maintenance branch.


HansOlsson (approved):
Looks good.

GallLeo (approved):
I'm a stupid engineer, not knowing about lexing rules. But, this small change looks OK to me.

henrikt-ma:
> I'm a stupid engineer, not knowing about lexing rules. But, this small change looks OK to me.

It is because we are humans that we need to keep things as simple as possible and remove unnecessary variations. :)

PR #2836, opened 2021-02-01
Minor cleanup of inline code and math for Derivative annotation

henrikt-ma:
Triggered by the _inputVar1_ formatted as normal text.

### DIFF
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -1359,16 +1359,12 @@ DensityDerivative d_rho2=der (rho2);
 
 \begin{itemize}
 \item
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Seems ok.
However, we might revisit having these fragments as "grammar".
I understand the meta-symbols like `{}` are only available in grammar.

henrikt-ma:
> However, we might later revisit having these fragments as "grammar".
> 
> I understand that the meta-symbols like `{}` are only available in grammar, but it seems a bit odd that everything becomes grammar due to that.

True; the mix of grammar and pseudo-code records in this chapter is rather messy.

(I also didn't like at all the use of one item lists for introducing these sub-annotations.)

PR #2841, opened 2021-02-09
Avoid current HTML jump target being hidden behind page header

henrikt-ma:
> - [ ] When clicking on some heading in the table of contents (left bar), the document jumps to the right spot. But, the selected heading is hidden by the header bar.

### DIFF
--- a/css/MLS-navbar-left.css
+++ b/css/MLS-navbar-left.css
@@ -110,13 +110,24 @@ nav > div.ltx_TOC {
   padding: 0.5rem 1rem;
   background: #F6F6F6; /* Same as body. */
   color: black; /* Same as side bar. */
+  border-bottom-color: #DE1D31; /* "Bouncing ball red" */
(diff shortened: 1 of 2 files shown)


HansOlsson (changes requested):
It fixes the problem, but introduces a new one.
If you jump to the start of a chapter it is shown with less margin at the top - that is mostly ok, but the main page (MLS.html) has the top image so far up that the top part of the gray "M" is cut out.

henrikt-ma:
> It fixes the problem, but introduces a new one.
> If you jump to the start of a chapter it is shown with less margin at the top - that is mostly ok, but the main page (MLS.html) has the top image so far up that the top part of the gray "M" is cut out.

Well spotted!

HansOlsson (approved):
Looks good now.
(Except that the clicking in the index now has the same issue... Ignore that.)

PR #2849, opened 2021-02-12
Don't call the Dialog a "menu"

henrikt-ma:
This is a follow-up to #2819, containing additional changes from the discussion that never got included in that PR.


eshmoylova (approved):
Looks good to me.

HansOlsson (approved):
Looks good.

PR #2850, opened 2021-02-15
More systematic use of \lstinline for figure annotations Axis and Curve

henrikt-ma:
Just some minor formatting issues that popped up when working on #3749 and #3750.


maltelenz (approved):

HansOlsson (approved):
Looks good.

PR #2856, opened 2021-02-16
Minor formatting cleanup related to colon (:) dimensions

henrikt-ma:
See commit messages for details.

PR #2860, opened 2021-02-19
Add missing plural 's'

henrikt-ma:
This is a commit I had forgotten to push for #2856.


HansOlsson (approved):
Ok. This is one of the cases where it may make sense to have two commits - one to fix line breaks and one to fix the problem to make it easier to find the change.

henrikt-ma:
> Ok. This is one of the cases where it may make sense to have two commits - one to fix line breaks and one to fix the problem to make it easier to find the change.

Absolutely!  However, I'm more concerned about upcoming changes like when we have agreed on a style guide.  There will be lots of small changes spread out throughout the entire document, and having to make every little change in two steps will be rather annoying…  Then imagine having ongoing work for the implementation of Ternary in a parallel branch…

PR #2879, opened 2021-03-01
Break down first items describing ExternalObject

henrikt-ma:
This is a separate PR for the text cleanup suggested in #2862.

Having separate items for the constructor and destructor makes the item list easier to organize and digest.


HansOlsson (approved):
Looks good.

PR #2894, opened 2021-03-15
Fix description of "proposes to not inline"

henrikt-ma:
This fixes a regression introduced after 3.4.


HansOlsson (approved):
Good. Must be a copy-paste error.

PR #2908, opened 2021-03-30
Add newtxtt.sty to LaTeXML continuous integration builds

henrikt-ma:
Needed for CI build of #2785.

### DIFF
--- a/.CI/Jenkinsfile
+++ b/.CI/Jenkinsfile
@@ -18,7 +18,7 @@ pipeline {
   stage('build') {
     agent {
       docker {
-        image 'modelicaspec/latexml:20201130'
+        image 'modelicaspec/latexml:20210330'
         label 'linux'
         alwaysPull true
       }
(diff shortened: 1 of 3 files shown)


sjoelund:
Is the following intended for connect and der?

Screenshot_20210330_093657

henrikt-ma:
> Is the following intended for connect and der?

That is the topic of #2785.

sjoelund (approved):
Looks OK, but I think the keywords list could use some other language definitions where connect and der were printed as keywords that are not operators. When will latexml support minted so we could have a better syntax highlighting? :)

HansOlsson:
> Looks OK, but I think the keywords list could use some other language definitions where connect and der were printed as keywords that are not operators. When will latexml support minted so we could have a better syntax highlighting? :)

Or have a language dialect for that list and possibly for Partial Derivatives of Function.

henrikt-ma:
It needs to be seen in the light of #2785, which now builds thanks to this PR:
pr-2785-keywords

PR #2898, opened 2021-03-17
Constrain dimension of equalityConstraint

henrikt-ma:
Fixes #2886.

According to phone meeting poll: https://github.com/modelica/ModelicaSpecification/issues/2886#issuecomment-800323151

PR #2909, opened 2021-03-31
Restore non-strict inequality for dimension of equalityConstraint

henrikt-ma:
Addressing review comment by Elena in already merged PR: https://github.com/modelica/ModelicaSpecification/pull/2898#discussion_r604278037

Comment in original PR by @eshmoylova:
> Sorry, I am too late with my review.  I see the restriction of 'n > 0' was added. There is at least one Modelica library [that sets it to 0](https://github.com/lbl-srg/modelica-buildings/blob/bddfd49d63a9c056073551b1267591be680e4852/Buildings/Electrical/PhaseSystems/PartialPhaseSystem.mo#L30) @mwetter do you want to comment why it is necessary to have it as 0. The change was merged already. Should we open a separate issue for that?

Comment in original PR by @beutlich:
> It's also in MSL: [Modelica.Electrical.QuasiStatic.Types.Reference](https://github.com/modelica/ModelicaStandardLibrary/blob/f30a6ea93530d97db634ca8a1490a755376c20d8/Modelica/Electrical/QuasiStatic/Types/Reference.mo) @AHaumer

### DIFF
--- a/chapters/connectors.tex
+++ b/chapters/connectors.tex
@@ -813,7 +813,7 @@ record Record
   end equalityConstraint;
 end Record;
 \end{lstlisting}
-The array dimension $n$ of \lstinline!residue! shall be a constant \lstinline!Integer! expression that can be evaluated during translation, with $n > 0$.
+The array dimension $n$ of \lstinline!residue! shall be a constant \lstinline!Integer! expression that can be evaluated during translation, with $n \geq 0$.
 The \lstinline!equalityConstraint! expresses the equality between the two type instances \lstinline!T1! and \lstinline!T2! or the record instances \lstinline!R1! and \lstinline!R2!, respectively, as the $n$ non-redundant equation residuals returned in \lstinline!residue!.
 That is, the set of $n$ non-redundant equations stating that \lstinline!R1 = R2! is given by the equation (\lstinline!0! represents a vector of zeros of appropriate size):
 \begin{lstlisting}[language=modelica]


henrikt-ma:
Pinging people mentioned in quotes added when editing the comment above: @mwetter, @AHaumer

beutlich:
Can we claim, that if the residue dimension is zero, that the type actually is not an overdertermined type?

beutlich:
For Buildings.Electrical.PhaseSystems.PartialPhaseSystem.ReferenceAngle.equalityConstraint I wonder, why the two inputs theta1/theta2 have a vector dimension, where it is supposed to be scalar?

henrikt-ma:
> Can we claim, that if the residue dimension is zero, that the type actually is not an overdertermined type?

Isn't it the opposite, that the dimension of the residual is the difference between the degrees of freedom in the unconstrained type minus the rank of the constraint making the type overdetermined?  That is, an empty residual means that the values are completely determined by the type itself?

beutlich:
Thank for the explanation. I somehow still struggle with the wording that "completely determined" is a special case of "overdetermined".

henrikt-ma:
I suppose one should also read the example [9.4.3.1 _An Overdetermined Connector for Power Systems_|https://specification.modelica.org/master/connectors-and-connections.html#an-overdetermined-connector-for-power-systems] to see why a "completely determined" type is not a "ridiculously overdetermined" type.

HansOlsson (approved):
Ok,

mwetter:
In the `Buildings` library, we set the dimension to 0 at 

https://github.com/lbl-srg/modelica-buildings/blob/947abac770e5265eefdd5e8b37c88cb13bc10e55/Buildings/Electrical/PhaseSystems/PartialPhaseSystem.mo#L30

because we implemented it according to the example in the MSL that also sets it to 0, see 
https://github.com/modelica/ModelicaSpecification/blob/103210bd083a279de32025fb8052786a4136b871/chapters/connectors.tex#L1010

I don't recall another reason for doing so.

PR #2910, opened 2021-03-31
Cleanup of "Matching Function"

henrikt-ma:
This is a mix of adding missing formatting, fixing typos introduced after 3.4, and fixing a typo that was already present in 3.4, see individual commit messages.


HansOlsson (approved):
Ok

PR #2912, opened 2021-03-31
Remove blank after der

beutlich:


henrikt-ma (approved):
Looks good.  (It's unrelated that there are some binary operators – in particular the `=` signs – here that would look better with added space on each side.)

PR #2913, opened 2021-03-31
Add blank after extends

beutlich:


henrikt-ma (approved):
This is a good one!

HansOlsson (approved):
Good.
Note that this was present already in Modelica 3.4.

Issue #1048, opened 2018-11-04
GenerateEvents annotation not needed in Modelica 3.2 rev2


modelica-trac-importer:
**Reported by fcasella on 25 Mar 2013 10:40 UTC**
After the recent Design Meeting in Piacenza, I have done a thorough analysis of the use of the former InlineNoEvent = false annotation (renamed GenerateEvents = true in Modelica 3.2rev2 and Modelica 3.3) in Modelica.Media. 

The analysis is reported below, and has been reviewed by Hubertus, who agrees with my conclusions. Basically, the annotation was misplaced in a lot of places, either because the functions involved were smooth, or because the medium package was not intended to be used across phase boundaries, while it was deliberately not placed in the only package where it would have had some effect, namely WaterIF97_base. The conclusion is that the GenerateEvents can be removed entirely from Modelica.Media 3.2.1, without changing its semantics w.r.t. Modelica 3.2 (done in r6125). 

As the GenerateEvents annotation is not used anywhere else in the MSL, it is no longer needed in Modelica 3.2 rev2, nor it needs to be implemented by tool vendors in order to support it. If the only reason to have this annotation in Modelica 3.2 rev2 was to support MSL 3.2.1, then I would then strongly suggest to remove it from specification text, in particular as its current formulation has sparked some controversy in the language design group.

I think that in general supporting event generation by discontinuous functions would be a nice feature to have in Modelica (in particular for future medium models), but it should clearly defined in the function semantics, not introduced by some kind of back-door annotation. Please refer to ticket #1026 for further comments on this topic.

----

In MSL 3.2, there were 8 functions in the Water IF97 model that had the InlineNoEvent = false annotation (which has now been more correctly renamed GenerateEvents = true), all of them in functions of the IF97_Utilities package:

phase_ps, phase_ph, phase_dT, s_pT, cp_pT, beta_pT, kappa_pT

As far as I understand, only the first three make sense, because the models using the \*_pT function cannot handle crossing the phase boundary at all (pT is a bad choice of independent variables if there are phase changes). 

So, I would say that only the annotations in phase_ps, phase_ph, and phase_dT were meaningful. But then, these functions are only called by WaterIF97_fixedregion, where they are completely useless: if the a priori assumption is that the region is fixed, there cannot be phase changes, and therefore no event generation. They are \*not\* used by functions in WaterIF97_base, where they might be useful if smoothModel = false.

[...]

PR #2914, opened 2021-04-10
Remove outdated reference of MSL for GenerateEvents annotation

beutlich:
See also #1048


HansOlsson (approved):
Looks good. As far as I can see there's no current use of it in MSL so nothing to replace it with.

PR #2932, opened 2021-05-04
Add a meta viewport tag to counter small text

sjoelund:
Android Chrome has very small text, which this fixes.
This is one of the problems reported in #2825.


henrikt-ma (approved):
Sure, let's give it a try!

PR #2933, opened 2021-05-04
Do not use block links in the page header

sjoelund:
An issue was reported in #2825 where clicking the menu in top left
instead went to the block link the header.

### DIFF
--- a/css/MLS.css
+++ b/css/MLS.css
@@ -89,3 +89,6 @@ a:hover { text-decoration: underline; }
 .ltx_indexrefs > .ltx_text:first-child:before {
   content: " –";
 }
+
+.ltx_page_header *[rel~="up"],
+.ltx_page_footer *[rel~="up"] { display: table; margin: 0 auto; text-align: center; }


henrikt-ma:
Do you have a more specific link to the issue that this fixes?

sjoelund:
> Do you have a more specific link to the issue that this fixes?

No. One would need to be created in that case. Or the message can be updated as this is an issue in itself.

henrikt-ma:
> No. One would need to be created in that case. Or the message can be updated as this is an issue in itself.

I'm just curious to understand what it is that is fixed by this change.

css/MLS.css
```diff
@@ -89,3 +89,6 @@ a:hover { text-decoration: underline; }
 .ltx_indexrefs > .ltx_text:first-child:before {
   content: " –";
 }
+
+.ltx_page_header *[rel~="up"],
+.ltx_page_footer *[rel~="up"]   { display:table; margin: 0 auto; text-align:center; }
```
henrikt-ma, line 94: Matching formatting with surrounding code:
```suggestion
.ltx_page_footer *[rel~="up"] { display: table; margin: 0 auto; text-align: center; }
```

henrikt-ma:
> > No. One would need to be created in that case. Or the message can be updated as this is an issue in itself.
> 
> I'm just curious to understand what it is that is fixed by this change.

Just found it.  This is the kind of link I was looking for: https://github.com/modelica/ModelicaSpecification/issues/2825#issuecomment-831720349

css/MLS.css
henrikt-ma: Looks good now.  Resolving.

henrikt-ma (approved):
Now that I think I understand what this is about, it looks like a plausible fix.

sjoelund:
Basically, it stops this area from being clickable:
Screenshot_20210505_080920

henrikt-ma:
Sounds great!

sjoelund:
I just made a new commit to see how the link to the comment renders, and it seems fine. Just need to wait for the CI.

PR #2946, opened 2021-05-21
Clean up formulations and formatting related to 'is not provided' in 'Synchronous Language Elements'

henrikt-ma:
Two changes:
- Make sure symbolic argument is math-formatted.
- Remove redundant "argument" before math-formatted symbolic reference to argument.


HansOlsson (approved):
Looks good.

PR #2957, opened 2021-06-17
Modernize presentation of some more annotations

henrikt-ma:
What started with the intention of just fixing a minor formatting issue related to `missingInngerMessage` ended up as a wider sweep of modernizing the presentation of annotations.  With this, all annotations presented outside the _Annotations_ chapter use the modern style.  In particular, all annotations in the _Functions_ chapter are now presented in the same way.

Ironically, `missingInngerMessage` still belongs to one of the annotation sections where the presentation hasn't been modernized.  It is one of the two more obvious sections next in turn for modernization (better handled in a separate PR to avoid scope creep in the current PR):
- 18.7 _Annotations for the Graphical User Interface_ (presentation would benefit enormously from new style)
- 18.8 _Annotations for Version Handling_ (can be done with very little work)

Then there are also the annotations presented using more or less deeply nested structures of pseudo code records, and these are not as clear how to modernize.  Examples include:
- 18.2.2 _Annotations for Figures_
- 18.9 _Annotations for Access Control to Protect Intellectual Property_


henrikt-ma:
With the merge conflicts now resolved, maybe this would be a good time to get this merged?

henrikt-ma:
@HansOlsson don't you think merging this could be considered a nice step in the direction for annotation cleanup decided at the last web meeting?

HansOlsson:
> @HansOlsson don't you think merging this could be considered a nice step in the direction for annotation cleanup decided at the last web meeting?

Possibly, but I saw that only 1 of 2 tasks were completed, and thus skipped it.

chapters/functions.tex
```diff
@@ -2331,65 +2333,110 @@ \subsection{Aliasing}\label{aliasing}
...
+\lstinline!SourceDirectory! & Directories containing source code& \Cref{modelica:SourceDirectory}\\
+\hline
+\end{tabular}
+\end{center}
+
+They can all specify either a scalar value or an array of values as indicated by the \lstinline[language=grammar]!string-or-string-array! syntax introduced with the \lstinline!Library! annotation.
+
+The \lstinline!ModelicaLibraryName! used below for \lstinline!IncludeDirectory!, \lstinline!LibraryDirectory!, and \lstinline!SourceDirectory! indicates the top-level class where the annotation is found in the Modelica source code.
+
+\begin{annotationdefinition}[Library]
```
HansOlsson, line 2356: I would prefer if we just write that it should be a string or an array of strings, and don't use grammar fragments unless needed.

henrikt-ma: Are you sure this isn't a good place to use it?  It is defined once, used in the definitions of 5 different annotations so far, and also mentioned in the running text.  It would take considerably more text and hand waving to do it without giving a name to the idea.

While we might be able to change many of the annotations to the pseudo-code style of presentation, the ones we have here belong to the category where specification using grammar actually makes sense as we can't express the type of the annotation using a Modelica type.

henrikt-ma: Couldn't we just keep the grammar fragment for now, and revisit this once we take the bigger grip on presentation of annotations as decided at the last phone meeting (can't find in what issue that decision was noted)?

henrikt-ma:
@HansOlsson, would it help moving this forward if I request reviews from other people?  I feel we are making progress on the presentation on so many annotations now, and it would be nice if we could move on to the principal discussion about what to do with things like the deeply nested pseudo-code record structures and the unstructured annotations that can't be presented as pseudo-code records.

henrikt-ma:
Merge conflicts have been resolved, so this is once again ready.  Adding another reviewer, hoping for more action.

HansOlsson:
> @HansOlsson, would it help moving this forward if I request reviews from other people? I feel we are making progress on the presentation on so many annotations now, and it would be nice if we could move on to the principal discussion about what to do with things like the deeply nested pseudo-code record structures and the unstructured annotations that can't be presented as pseudo-code records.

Possibly, but:
- I don't think we should think of it as pseudo-code records, but as pseudo-code _declarations_. Thus singleInstance can still be defined as a pseudo-code declaration (`Boolean singleInstance=false`), similarly `Boolean choicesAllMatching`.
- Even Include could be kind of a pseudo-declaration, `StringScalarOrVector Include`.
- And I think we should clearly prefer pseudo-code declarations; as the grammar fragments have caused a number of problems.

eshmoylova (commented):
I looked through the changes. But before I start excessively commenting on individual items, I'd like to clarify what the goal of this change. Do we want to update all instances of annotation definitions? What are we updated them to? I see 3 styles:
1.  Pseudo-record definition
2. Grammar style
```
documentation-class-annotation :
   annotation "(" DocumentationClass "=" true ")"
```
3. Piece of code, e.g.
```
annotation(preferredView = view)
```
I see some of the style 3 converted into style 2 by this PR but not all. Is that ok or do we want to find all instances?

Also, it seems to me that the decision was to convert more to the record style. Of course, that would apply only to structured annotations. For consistency, it would make sense to present simple annotations as component declarations:
```
String defaultComponentName = "name";
```
But it is harder with cases where only couple of values are allowed. For example, preferredView can only be "info", "diagram", "icon" and "text". You would need to say something like:
```
String preferredView = view;
```
where view is one of "info", diagram", "icon", and "text".

HansOlsson:
> I see some of the style 3 converted into style 2 by this PR but not all. Is that ok or do we want to find all instances?
> 
> Also, it seems to me that the decision was to convert more to the record style.

To me the record style is preferable.

> Of course, that would apply only to structured annotations. For consistency, it would make sense to present simple annotations as component declarations:
> 
> ```
> String defaultComponentName = "name";
> ```
> 
> But it is harder with cases where only couple of values are allowed. For example, preferredView can only be "info", "diagram", "icon" and "text". You would need to say something like:
> 
> ```
> String preferredView = view;
> ```
> 
> where view is one of "info", diagram", "icon", and "text".

Or we could use another annotation to describe that:
```
parameter String preferredView 
  annotation(choices(
       choice="info" "Documentation view", 
       choice="diagram" "Graphical diagram", 
       choice="icon" "Icon of model",
       choice="text" "Modelica source code"));
```
I understand it looks close to recursive, but it might still be useful.

HansOlsson:
> I thought about using choices. What I am not sure about is: does it indicate that no other values are allowed? The description of choices only says that each choice indicates a suitable redeclaration or modification. But are those suitable values the only ones that are allowed? Maybe it does't matter. Probably if the preferredView is something other that the ones above it would be ignored. Or is it supposed to give an error?

I would say it doesn't matter much. I could imagine that we later want to extend this (similarly as "icon" was added); so we shouldn't make it a hard error.

henrikt-ma:
My intention with this PR is only to modernize the presentation of _some more annotations_, not to fix the global problem of getting all annotations into the new form, whatever that is.

henrikt-ma:
> > I thought about using choices. What I am not sure about is: does it indicate that no other values are allowed? The description of choices only says that each choice indicates a suitable redeclaration or modification. But are those suitable values the only ones that are allowed? Maybe it does't matter. Probably if the preferredView is something other that the ones above it would be ignored. Or is it supposed to give an error?
> 
> I would say it doesn't matter much. I could imagine that we later want to extend this (similarly as "icon" was added); so we shouldn't make it a hard error.

Didn't we just spend quite a lot of time to come up with the current design where the variable _view_ is used in the "structured presentation" of the annotation, with a description of the allowed values for _view_ given in the text below?  I don't think it's that bad, and at least less strange compared to explaining annotations in terms of `choices`.  On the other hand, I like tables, so I think expressing the the possible values for _view_ using a table instead would be a nice variant.

HansOlsson:
> > I see some of the style 3 converted into style 2 by this PR but not all. Is that ok or do we want to find all instances?
> 
> My intention with this PR is only to modernize the presentation of _some more annotations_, not to fix the global problem of getting all annotations into the new form, whatever that is.

To me that seems odd.

I would first define what style we prefer as the modern one (as stated above I prefer the "record-style" regardless of whether it is records, scalars, arrays or pseudo-records), and then try to work to get all (or most) annotations to use that.

It might also be that this PR is trying to do two things (which is always complicated):

- Present how the annotation works in a new way
- Present the syntax/grammar/what-ever for the annotation in a new way

And we focus on different parts of this, you might be looking at the first part and the rest of us are stuck on the second part.

henrikt-ma:
> > > I see some of the style 3 converted into style 2 by this PR but not all. Is that ok or do we want to find all instances?
> > 
> > 
> > My intention with this PR is only to modernize the presentation of _some more annotations_, not to fix the global problem of getting all annotations into the new form, whatever that is.
> 
> To me that seems odd.

Maybe someone got the impression that this PR is partly introducing new styles in order to partly solve the annotation cleanup we decided to do long time ago.  That's not the case.  This PR only does what can be done easily using the modern style for annotations that is already in place.

> I would first define what style we prefer as the modern one (as stated above I prefer the "record-style" regardless of whether it is records, scalars, arrays or pseudo-records), and then try to work to get all (or most) annotations to use that.

The point was to do away with the easy parts, so that only the parts needing more discussion would need to wait.  By this split, it also gets easier for us to filter out the annotations for which the current "modern style" is insufficient.

> It might also be that this PR is trying to do two things (which is always complicated):
> 
> * Present how the annotation works in a new way
> * Present the syntax/grammar/what-ever for the annotation in a new way
> 
> And we focus on different parts of this, you might be looking at the first part and the rest of us are stuck on the second part.

No, I don't think that's the case, but I have the impression that it is the fear of this being the case that's holding it back.

HansOlsson:
> The point was to do away with the easy parts, so that only the parts needing more discussion would need to wait. By this split, it also gets easier for us to filter out the annotations for which the current "modern style" is insufficient.

Before anything else I believe we should clarify what we are discussing with 'modern' presentation. That ambiguity is a general problem with adjectives like 'modern', 'new' etc.

- Is it a new way of presenting annotations in terms of new latex-commands.
- Or is it a different way of presenting the syntax/grammar/what-ever for the annotation?

henrikt-ma:
> > The point was to do away with the easy parts, so that only the parts needing more discussion would need to wait. By this split, it also gets easier for us to filter out the annotations for which the current "modern style" is insufficient.
> 
> Before anything else I believe we should clarify what we are discussing with 'modern' presentation. That ambiguity is a general problem with adjectives like 'modern', 'new' etc.

Sure.

> * Is it a new way of presenting annotations in terms of new latex-commands.

No.

> * Or is it a different way of presenting the syntax/grammar/what-ever for the annotation?

Yes, and this style was introduced long time ago (buy now).  It was probably simply too much work or risk of introducing merge conflicts with other ongoing changes for it to get immediately applied widely.  This PR does a lot of that work, and allows us to see more clearly what remains to design.

I don't see any major problems with extending the design to also work in the remaining cases, but I think it would help us if we could separate the application of a style that has already been introduced from introducing new design that can also handle the remaining annotations.

HansOlsson:
> Yes, and this style was introduced long time ago (buy now). It was probably simply too much work or risk of introducing merge conflicts with other ongoing changes for it to get immediately applied widely. This PR does a lot of that work, and allows us to see more clearly what remains to design.

I don't recall this design choice, and I think it makes sense to revisit it after seeing the consequences.

The basic problem is that by presenting it as a grammar we get something that looks neither as the Modelica code you would write nor as the declaration of the relevant items, and things get lost in the syntax.

The styles we currently have are:

- Declared in Modelica as used in https://specification.modelica.org/master/annotations.html#annotations-for-figures and https://specification.modelica.org/master/annotations.html#graphical-primitives https://specification.modelica.org/master/annotations.html#access-control-to-protect-intellectual-property which is my preferred variant. It was also the basis for the custom annotations proposal.
- Correct grammar as in https://specification.modelica.org/master/annotations.html#annotations-for-simulations which to me hides the relevant parts in syntax and should be replaced by declarations in Modelica. A key observation is that we previously only considered records, but we also need annotations of simple types as well.
- Something between example code and grammar as in https://specification.modelica.org/master/annotations.html#annotations-for-version-handling which is complicated for a number of reasons. Just changing to a grammar will not fix that, there are other tweaks needed as well.

The reason we used a mix of example code and grammar is that it's easier to read `choices` than `"choices"`, when referring to choices-annotation (and similarly in other cases).

The reasons I prefer the first variant is actually two-fold:

- It seems the easiest to understand so far; and an easier design reduces the risk of errors.
- It _constrains_ our thinking for future annotations. Instead of considering designing a grammar/language for something new we will be looking at designing its data (or data-structure). I know that we cannot get all annotations in that form, but the more the better.

The second point has an interesting implication for one of the discussion points - IncludeDirectory which currently can be a string or a string-vector. If we start thinking of data-structures it would make sense that IncludeDirectory is a string-vector, with some backward compatibility fix for a scalar string. Such observations are valuable and comes from restricting the choices to a data-structure.

eshmoylova:
I am with @HansOlsson. I do not believe the grammar style is a good way to present the annotations. It is not clear to me that there was a group decision to use the grammar style as the preferred style. From the discussions that I remember the preference was given to the record style. There was a hesitancy because we can only use records for structured types. But as we already discussed in this PR it can be accomplished by using types. Perhaps we should call it type-record style instead of just record type. So I do not think we should spend time on converting anything into the grammar style as done in this PR, and my preference is to close this PR.

@henrikt-ma if you feel strongly that there is reason to use the grammar style for annotations then I suggest we leave this open and bring it up for discussion at the next meeting.

henrikt-ma:
Just replying quickly before going on vacation for one week.  The goal of the PR is not to favor the use of grammar over pseudo-code, the goal is to transform the presentation into the style where we have tables summarizing the annotation described in a section, plus a structured way of framing the presentation of each annotation.

The use of grammar is only a consequence of not being able to directly apply the pseudo-code style that we have successfully applied to some of our other annotations.  In other words, I'm trying to do the best of what we have today.  If we had better alternatives available I'd happily apply them.

For me, it would be OK to work out those better alternatives as part of this PR, although my preference would be to have a separate discussion about this later, and I think it will be easier to have that discussion if we can see clearly which annotations that currently need grammar in order to describe their structure with sufficient precision.

I'll be back on March 14, take care until then!

chapters/inheritance.tex
```diff
@@ -983,60 +983,107 @@ \subsection{Restrictions on Redeclarations}\label{restrictions-on-redeclarations
...
+Using \lstinline!choices! to request that a checkbox is used to select modification for a \lstinline!Boolean! variable:
 \begin{lstlisting}[language=modelica]
 parameter Boolean useHeatPort = false annotation(choices(checkBox = true));
 \end{lstlisting}
 \end{example}
+
+\begin{annotationdefinition}[choicesAllMatching]
+\begin{synopsis}[grammar]\begin{lstlisting}
+"choicesAllMatching" "=" ( true | false )
+\end{lstlisting}\end{synopsis}
```
henrikt-ma, line 1070: OK, let's start with something relatively simple, like this one.

I suppose an alternative starting point would be this:
```
Boolean choicesAllMatching
```

To express that we just want constant values, should we use `constant`?
```
constant Boolean choicesAllMatching
```

Should we on top of this write in the text that the value of the annotation should be a literal, or should we introduce some new pseudo-code way of saying this systematically?
```
literal Boolean choicesAllMatching
```

henrikt-ma: (Still, I'd much rather have separate discussions about how to deal with various annotations that we cannot express precisely using today's pseudo-code style.  In other words, go with grammar for now, only to keep it until we have invented something else allowing us to present annotations systematically and without loss of precision.)

eshmoylova: I agree that it's worth opening a separate issue to bring up for discussion. Sorry, I totally did not catch that the idea was to update the presentation of tables for annotations. I will have to review it all again under that assumption and ignore the way annotations are written.

henrikt-ma:
> Regarding how to write `choicesAllMatching` I prefer the simple `Boolean choicesAllMatching`. Partially because it is simple, and partially because there are various enhancements ideas of supporting non-literal values for such parameters; and we shouldn't just rule them out immediately. One is using `parameter Boolean p=true;parameter Real x=2 annotation(Evaluate=p);` (added due to demand in Dymola, I believe there might already be an issue proposing that, but I couldn't find it) another is #3133
> 
> Update: clarified "Issue".

Changing from literal values to `Boolean` parameter or constant would be a welcome change, but I don't think we should mix changes with cleanup of the presentation.  Either a change like this should be discussed as a separate issue, or we should have a bigger MCP level discussion about general relaxation of annotation expressions.

Having a cleaned up presentation of where we are today will be the best starting point for future discussions about how to change semantics.

HansOlsson:
> > Regarding how to write `choicesAllMatching` I prefer the simple `Boolean choicesAllMatching`. Partially because it is simple, and partially because there are various enhancements ideas of supporting non-literal values for such parameters; and we shouldn't just rule them out immediately. One is using `parameter Boolean p=true;parameter Real x=2 annotation(Evaluate=p);` (added due to demand in Dymola, I believe there might already be an issue proposing that, but I couldn't find it) another is #3133
> > Update: clarified "Issue".
> 
> Changing from literal values to `Boolean` parameter or constant would be a welcome change, but I don't think we should mix changes with cleanup of the presentation. 

True, and the problem with this PR is exactly that it is perceived as changing more than one thing - and the intended changes are lost among the other ones.

HansOlsson:
> 

@sjoelund have you changed anthing?

henrikt-ma:
> @henrikt-ma, @HansOlsson, @sjoelund, I would like to view the build for this PR, so that I don't have to build it myself.

While waiting for CI builds to become available again, I'll send the document directly to @eshmoylova.

chapters/annotations.tex
```diff
@@ -1808,51 +1896,105 @@ \subsubsection{Conversion Rules}\label{conversion-rules}
...
-one of the following ways in a directory given in the \lstinline!MODELICAPATH!:
+The top-level class $\mathit{packageName}$ with version $\mathit{packageVersion}$ can be stored in one of the following ways in a directory given in the \lstinline!MODELICAPATH! (\cref{the-modelica-library-path-modelicapath}):
 \begin{itemize}
 \item
-  The file \lstinline!IDENT ".mo"!\\
+  The file $\mathit{packageName}$\filename{.mo}\\
   Example: \filename{Modelica.mo}
 \item
-  The file \lstinline!IDENT " " VERSION-NUMBER ".mo"!\\
+  The file $\mathit{packageName}$\textvisiblespace$\mathit{packageVersion}$\filename{.mo}\\
```
eshmoylova, line 1908: \textvisiblespace looks a lot like an underscore in the pdf, depending on the zoom. It is not really clear that it should be a space and not something else
image

eshmoylova (commented):
The addition of tables looks good. As already discussed, I don't really like that you had to convert some annotation definitions to the grammar style but I see how it was necessary in some cases to introduce some formal definition to fit with the rest of the style. So it's not easy to just get back to what it was without losing the tables. I would be ok with keeping the grammar style annotations as presented in this PR for the time being just so that we can move on with adding the tables and then, as discussed, opening a new issue and new PR for unifying the format of annotations. Perhaps it would be better to rename this PR into "Modernize table presentation of some more annotation"?

One more thought about the format of type/record vs grammar before I forget it: how would you represent the choice annotation in a record format? It is supposed to be any possible modification on the class for which it is specified, so the structure of the annotation is the same as the corresponding class.

chapters/inheritance.tex
```diff
@@ -983,60 +983,107 @@ \subsection{Restrictions on Redeclarations}\label{restrictions-on-redeclarations
   \end{nonnormative}
 \end{itemize}
 
-\subsection{Annotation Choices for Suggested Redeclarations and Modifications}\label{annotation-choices-for-suggested-redeclarations-and-modifications}
+\subsection{Annotations for Suggested Redeclarations and Modifications}\label{annotations-for-suggested-redeclarations-and-modifications}
 
-A declaration can have an annotation \fmtannotationindex{choices} containing modifiers on \lstinline!choice!, where each of them indicates a suitable redeclaration or modifications of the element.
+The annotations listed below are used to suggest certain redeclarations and modifications that should be made conveniently available in tools.
```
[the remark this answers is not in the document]
henrikt-ma: I think of user interfaces, but I'm not sure we need/should be that specific.  In addition to direct appearance in the Modelica tool's user interface, I could also imagine various kinds of model export where this information could be included for an improved user experience in some other context where the model is imported.  Could this be enough imagination to not be more specific?

chapters/annotations.tex
```diff
@@ -1263,11 +1281,11 @@ \section{Graphical User Interface}\label{annotations-for-the-graphical-user-inte
   $\ldots$
   flow Modelica.Units.SI.Force f[3]
   annotation(unassignedMessage =
-      "All Forces cannot be uniquely calculated. The reason could be that the
+    "All Forces cannot be uniquely calculated. The reason could be that the
      mechanism contains a planar loop or that joints constrain the same motion.
```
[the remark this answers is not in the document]
henrikt-ma: I'm not quite sure what you mean, but what I see is a pre-formatted message with lots of white-space after the line breaks that is going to give a really ugly result when included in the error message presented to the user.

I am generally in favor of the idea to try to set an example of good style in the specification that users can copy, although it may require some tricks to at the same time comply with the page width constrains of the specification.  In this case, I'll make use of string concatenation to fit the width constraint, although I might have considered using a very long line in an actual Modelica file.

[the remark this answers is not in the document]
henrikt-ma: OK now?

chapters/annotations.tex
henrikt-ma: I'm not sure what better alternatives there are.  Can't the example be considered sufficient clarification; when you see the example, don't you realize what that little underscore-looking thing really means?

henrikt-ma:
> Perhaps it would be better to rename this PR into "Modernize table presentation of some more annotation"?

I did a small variant of that.

HansOlsson:
Looking more I don't think that using textvisiblespace is significantly better than just plain space.
If we want to make it clearer we could use Modelica for something like `\lstineline!"$packageName$"+" "+"$packageVersion$"+".mo"!`

chapters/annotations.tex
henrikt-ma: Also, an important feature of the visible space is that it will also be there regardless of where line is wrapped – in general of special importance for readers of HTML on small screens (in theory we could always proof read the PDF to make sure lines are not wrapped at bad places…), although this case is too near the beginning of the line to really be at risk.

How about adding a comment for extra clarity?
```suggestion
  The file $\mathit{packageName}$\textvisiblespace$\mathit{packageVersion}$\filename{.mo} (with one space character separating $\mathit{packageName}$ from $\mathit{packageVersion}$)\\
```

eshmoylova: The only reason I know it's supposed to be a space and not an underscore (and a weird one at that) is because I look at the LaTeX source. So, my first reaction while looking at the pdf only was that there was a typo in the example and the underscore was missing. Actually, that's not true, my first reaction was: I had no idea underscore is supposed to separate the name and the version. I have to say that I find the grammar-style representation that was before this change, much clearer and concise. Since, we are not getting rid of the grammar-style anyways, at least for now, maybe we should just leave it as it was.

eshmoylova:
The title of the PR looks good. But I wonder if the description should be updated a bit as well. Maybe just adding updating "modernizing the presentation of annotations" to "modernizing the presentation of annotations using tables" in the first paragraph?

henrikt-ma:
> The title of the PR looks good. But I wonder if the description should be updated a bit as well. Maybe just adding updating "modernizing the presentation of annotations" to "modernizing the presentation of annotations using tables" in the first paragraph?

Sure, no problem.  Done.

chapters/annotations.tex
henrikt-ma: I'd like to avoid that for several reasons.  The obvious one is that our overall goal is to avoid use of grammar whenever possible.  Another one is that `VERSION-NUMBER` was actually never defined, so we'd have to introduce even more grammar to give this a meaning.  The third is that we should technically use the lexing language (like when `IDENT` is defined) and not grammar in order to be able to express that there's no room for white-space or comments between the tokens.  If we'd like to avoid overuse of grammar, it should be even more important to avoid overuse of lexing.

henrikt-ma: By the way, I don't think we need to be afraid of ending up adding comments like the one I sugest all over the document; the need to express separation by exactly one space character is unique to this section as far as I know, and therefore some kind of comment might actually be a good idea no matter what the formal specification looks like.

chapters/annotations.tex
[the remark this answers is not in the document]
eshmoylova: Sorry, I was not clear. The sentence "The reason could be..." is on two lines here. I was wondering if it was necessary to create that line break in the sentence manually because it is part of lsllisting. Sorry, I don't remember how the formatting of lines is performed inside lstlisting.

chapters/annotations.tex
eshmoylova: I have been waiting to see if anybody else would comment, but nobody has volunteered so far. @HansOlsson, do you have an opinion?

eshmoylova: > Another one is that `VERSION-NUMBER` was actually never defined, so we'd have to introduce even more grammar to give this a meaning. 

I thought the definition of `VERSION-NUMBER` is given in [18.8.1 Version Numbering](https://specification.modelica.org/master/annotations.html#version-numbering). But I see that it never actually uses the term `VERSION-NUMBER`. Should it? It is used a lot in 18.8.2

chapters/annotations.tex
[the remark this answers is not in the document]
henrikt-ma: We want all listings to have hard line breaks so that every single line fits within the width constraint of the PDF build.  This also improves readability of the HTML build, as one never needs to scroll horizontally to see the full contents of a listing.  Usually, this means one just has to combine suitable places for line breaks with proper indentation of the following lines, but it is different inside string literals as a line break isn't just whitespace there.  Instead, the line length constraint needs to be fulfilled by breaking up a long string in smaller parts, and concatenate the parts using the `+` operator.  This is what I did here, but as explained above, the reader needs to understand that this is probably not the way to do it in a real Modelica library.

chapters/annotations.tex
henrikt-ma: > I thought the definition of `VERSION-NUMBER` is given in [18.8.1 Version Numbering](https://specification.modelica.org/master/annotations.html#version-numbering). But I see that it never actually uses the term `VERSION-NUMBER`. Should it? It is used a lot in 18.8.2

As of this PR, it is not.  However, there is now a definition of the production rule _package-version_. Maybe it could be a solution to point out in the text that `\mathit{packageVersion}` needs to match _package-version_, along with a cross reference to the definition?

henrikt-ma: I went ahead and implemented that idea, and it turned out that as _package-version_ was defined to be a package version string literal, I had to actually introduce _PACKAGE-VERSION_ for the contents of such as string, which is what the `\mathit{packageVersion}` has to match.

chapters/inheritance.tex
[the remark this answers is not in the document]
eshmoylova: OK, good enough. Resolving.

chapters/inheritance.tex
eshmoylova: Resolving, since it was decided to move it into a separate discussion.

chapters/annotations.tex
eshmoylova: That looks good. I'll keep it unresolved until the next week (it's a holiday here tomorrow and maybe holiday on Monday in other countries) to see if anybody else wants to chime in about the use of `\textvisiblespace`.

eshmoylova: Ok, there were no other comments. So I guess, it's OK to keep `\textvisiblespace`. The suggestion [above with the clarifying comment](https://github.com/modelica/ModelicaSpecification/pull/2957/files#r841122964), would be good. Sorry, I thought it was already committed. But I see now that it's only in the comments so far.

henrikt-ma: Done.

henrikt-ma: Please mark as resolved if OK.

chapters/functions.tex
henrikt-ma: Any chance we can return to this once we turn to the question of how to present the things that can't be expressed easily in the record style established so far?

henrikt-ma:
Would it be meaningful to resolve the merge conflicts at this point, or should I keep waiting?

eshmoylova:
> Would it be meaningful to resolve the merge conflicts at this point, or should I keep waiting?

I think you can resolve them.

henrikt-ma:
Merge conflicts have been resolved, and at least there's one approval from @eshmoylova now. Maybe we could merge and then turn our attention to the more challenging questions about how to present annotations?

HansOlsson:
The general idea with tables seem good.
But:

- There are too many changes unrelated to the main idea. A PR shall be more focused. In this case there have been so many unrelated changes that it is difficult to see what the main idea is.
- Changing to using grammar for annotations reduces readability. It would be understandable for annotations that weren't described well previously, but here it is also applied to e.g. versionDate which was a good example of using a non-record declaration.

HansOlsson (changes requested):
The general idea with tables seem good.
But:

* There are too many changes unrelated to the main idea. A PR shall be more focused. In this case there have been so many unrelated changes that it is difficult to see what the main idea is. That's why the PR has been stuck - it is not possible to see what the actual idea is.
* Changing to using grammar for annotations reduces readability. It would be understandable for annotations that weren't described well previously, but here it is also applied to e.g. versionDate which was a good example of using a non-record declaration. Using a grammar to describe that is just a step backwards.

chapters/inheritance.tex
```diff
@@ -983,60 +983,107 @@ \subsection{Restrictions on Redeclarations}\label{restrictions-on-redeclarations
   \end{nonnormative}
 \end{itemize}
 
-\subsection{Annotation Choices for Suggested Redeclarations and Modifications}\label{annotation-choices-for-suggested-redeclarations-and-modifications}
+\subsection{Annotations for Suggested Redeclarations and Modifications}\label{annotations-for-suggested-redeclarations-and-modifications}
```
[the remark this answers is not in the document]
henrikt-ma: I checked it, and only the last `\label` is turned into the `id` of a `section` tag in the HTML.  Hence, we should only have one `\label`, and it would be good if it was more stable than the exact formulation of the section title.

chapters/inheritance.tex
```diff
@@ -983,60 +983,107 @@ \subsection{Restrictions on Redeclarations}\label{restrictions-on-redeclarations
...
-  constrainedby Modelica.Media.Interfaces.PartialMedium
-  annotation(choicesAllMatching = true);
+    choice(redeclare ConstantFriction a(c = 1) "$\ldots$"),
+    choice(redeclare TableFriction a(table = "$\ldots$") "$\ldots$"),
+    choice(redeclare FunctionFriction a(func = exp) "$\ldots$")
+  ));
 \end{lstlisting}
+\end{example}
 
-It can also be applied to nonreplaceable declarations, e.g.\ to describe enumerations.
```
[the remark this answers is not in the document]
henrikt-ma: It is there, but formulated as a stand-alone introduction to a separate example:
> Using \lstinline!choices! to suggest suitable value modifications for a non-replaceable declaration:

chapters/revisions.tex
```diff
@@ -178,7 +178,7 @@ \subsection{Main changes in Modelica 3.5}\label{main-changes-in-modelica-3-5}
 Ticket \href{https://github.com/modelica/ModelicaSpecification/issues/2288}{\#2288}.
 \item Added test-case annotation for incorrect models, \cref{modelica:TestCase}.
 Ticket \href{https://github.com/modelica/ModelicaSpecification/issues/2340}{\#2340}.
-\item Clarified choicesAllMatching, \cref{annotation-choices-for-suggested-redeclarations-and-modifications}.
+\item Clarified \lstinline!choicesAllMatching!, \cref{annotations-for-suggested-redeclarations-and-modifications}.
```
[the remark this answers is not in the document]
henrikt-ma: Never making edits to _revisions.tex_ isn't possible with a document that gets reorganized over time.  I also don't see the value of un-touched LaTeX source code.

As an alternative to updating cross references from the revision history, one could consider replacing obsolete cross references with a standard formulation such as _(original context of change is no longer available)_, instead of updating the cross references to point to the current best available approximation of the original reference.

chapters/annotations.tex
```diff
@@ -437,15 +437,31 @@ \section{Annotations for Simulations}\label{annotations-for-simulations}
 
 \section{Annotation for Single Use of Class}\label{annotation-for-single-use-of-class}
 
-For state machines it is useful to have single instances of local classes.
```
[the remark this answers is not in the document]
henrikt-ma: It is still there:
> This is useful for local classes in the state machines described in \cref{state-machines}.

chapters/annotations.tex
```diff
@@ -437,15 +437,31 @@ \section{Annotations for Simulations}\label{annotations-for-simulations}
...
+\hline
+\lstinline!singleInstance! & Only allow one component instance of class & \Cref{modelica:singleInstance}\\
+\hline
+\end{tabular}
+\end{center}
 
-The annotation \fmtannotationindex{singleInstance} in a class indicates that there should only be one component instance of the class, and it should be in the same scope as the class is defined.
+\begin{annotationdefinition}[singleInstance]
+\begin{synopsis}[grammar]\begin{lstlisting}
+"singleInstance" "=" true
```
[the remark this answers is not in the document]
henrikt-ma: Neither do I expect the average tool to implement this based on the grammar rules, but the formalism has the advantage of greater precision.  For example, it is straight-forward to express that the annotation expects a literal value, and not a general parameter expression that might need messy evaluation of parameters etc.

Still, I am ready to switch to something such as pseudo-variable as soon as we have worked out a systematic way of making use of them.  The use of grammar should therefore only be considered a temporary fix using the best formalism available for this kind of annotations today.

chapters/annotations.tex
```diff
@@ -1478,85 +1496,154 @@ \subsection{Connector Sizing}\label{connector-sizing}
...
 
-Version numbers are of the forms:
+Version numbers can be in one of several forms:
+\begin{lstlisting}[language=grammar]
+package-version :
+   """ ( main-version | pre-release-version | unordered-version ) """
+
+main-version : UNSIGNED-INTEGER { "." UNSIGNED-INTEGER }
+
+pre-release-version : UNSIGNED-INTEGER { "." UNSIGNED-INTEGER } " " { S-CHAR }
```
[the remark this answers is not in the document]

chapters/annotations.tex
```diff
@@ -1200,9 +1216,11 @@ \section{Graphical User Interface}\label{annotations-for-the-graphical-user-inte
...
 model World
   $\ldots$
-  annotation(defaultComponentName = "world",
-  defaultComponentPrefixes = "inner replaceable",
-  missingInnerMessage = "The World object is missing");
+  annotation(
+    defaultComponentName = "world",
+    defaultComponentPrefixes = "inner replaceable",
+    missingInnerMessage = "The World object is missing"
+  );
```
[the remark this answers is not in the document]
henrikt-ma: I could factor it out, but the problem is that starting another PR from the branch of this PR easily gets confusing in the Github workflow.  I see the problem with scope creep, but I sometimes find it better to do it this way than dealing with a potentially large number of tiny PRs that would be full of merge conflicts after merging the main PR (or, even worse, cause a large number of merge conflicts for the main PR).

chapters/annotations.tex
```diff
@@ -1271,11 +1289,11 @@ \section{Graphical User Interface}\label{annotations-for-the-graphical-user-inte
...
-      "All Forces cannot be uniquely calculated. The reason could be that the
-     mechanism contains a planar loop or that joints constrain the same motion.
-     For planar loops, use in one revolute joint per loop the option
-     PlanarCutJoint=true in the Advanced menu.
-     ");
+    "All Forces cannot be uniquely calculated. The reason could be that the " +
+    "mechanism contains a planar loop or that joints constrain the same motion.\n" +
+    "For planar loops, use in one revolute joint per loop the option " +
+    "PlanarCutJoint = true in the Advanced menu."
+  );
```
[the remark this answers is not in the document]
henrikt-ma: Same here.  The good thing is that everything is under the umbrella of cleaning up the presentation of annotations.  This PR is free of semantical changes that would require the attention of the bigger language group.

chapters/annotations.tex
```diff
@@ -1407,7 +1425,7 @@ \subsection{Connector Sizing}\label{connector-sizing}
 
 block Step
   // nIn cannot be set through the dialog (but maybe shown)
-  parameter Integer nIn=0 annotation(Dialog(connectorSizing=true));
+  parameter Integer nIn = 0 annotation(Dialog(connectorSizing = true));
```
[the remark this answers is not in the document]
henrikt-ma: Sorry, same reasons as above.

chapters/annotations.tex
```diff
@@ -1817,41 +1905,93 @@ \subsubsection{Conversion Rules}\label{conversion-rules}
...
-one of the following ways in a directory given in the \lstinline!MODELICAPATH!:
+The top-level class $\mathit{packageName}$ with version $\mathit{packageVersion}$ (matching \lstinline[language=grammar]!PACKAGE-VERSION! defined in \cref{version-numbering}) can be stored in one of the following ways in a directory given in the \lstinline!MODELICAPATH! (\cref{the-modelica-library-path-modelicapath}):
 \begin{itemize}
 \item
-  The file \lstinline!IDENT ".mo"!\\
+  The file $\mathit{packageName}$\filename{.mo}\\
   Example: \filename{Modelica.mo}
 \item
-  The file \lstinline!IDENT " " VERSION-NUMBER ".mo"!\\
+  The file $\mathit{packageName}$\textvisiblespace$\mathit{packageVersion}$\filename{.mo} (one space character separating $\mathit{packageName}$ from $\mathit{packageVersion}$)\\
```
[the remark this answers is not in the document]
henrikt-ma: Just because it is customary to use `\textvisisblespace` where close attention to spaces is needed in the LaTeX world of high quality technical documentation.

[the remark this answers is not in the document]
henrikt-ma: It was also the resolution of the length conversation ending in https://github.com/modelica/ModelicaSpecification/pull/2957#discussion_r867151745.

chapters/annotations.tex
```diff
@@ -1817,41 +1905,93 @@ \subsubsection{Conversion Rules}\label{conversion-rules}
...
   Example: \filename{Modelica 2.1.mo}
 \item
-  The directory \lstinline!IDENT! with the file \filename{package.mo} directly inside it\\
+  The directory $\mathit{packageName}$ with the file \filename{package.mo} directly inside it\\
   Example: \filename{Modelica/package.mo}
 \item
-  The directory \lstinline!IDENT " " VERSION-NUMBER! with the file \filename{package.mo} directly inside it\\
+  The directory $\mathit{packageName}$\textvisiblespace$\mathit{packageVersion}$ with the file \filename{package.mo} directly inside it\\
```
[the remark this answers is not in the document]
henrikt-ma: Same here, see: https://github.com/modelica/ModelicaSpecification/pull/2957#discussion_r867151745

chapters/annotations.tex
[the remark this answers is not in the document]
HansOlsson: Well, it was resolved that way due to the lack of other comments. And now we have another comment.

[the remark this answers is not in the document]

chapters/annotations.tex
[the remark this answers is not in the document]

HansOlsson:
As far as I can see:
- The "table presentation" part of this was done in #3628 (at least partially).
- For consistency this tried to change annotations into grammar-constructs. Instead of doing that more annotations are now defined as Modelica-declarations due to #2999 and #3448 (with additional semantics).
- I have not investigated the changes (adding white-space etc).

I thus think it is best to close this one, and if needed open a new one for the stylistic changes from scratch, and if possible add more "table presentation".

PR #2972, opened 2021-08-09
Align keywords for 'grammar' with keywords for 'modelica'

henrikt-ma:
This fixes a problem I noted when looking at the grammar describing partial derivative functions.  Since `der` wasn't a keyword for `language=grammar`, it was typeset as the name of a grammar production rule.

In addition to treating `der`, this PR also makes `enumeration` a grammar keyword, fixing a similar issue in the `short-class-specifier` rule.

### DIFF
--- a/mlsshared.sty
+++ b/mlsshared.sty
@@ -102,12 +102,21 @@
     breaklines=true,
     breakatwhitespace=true,
     morekeywords=[1]{%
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Ok, makes sense to make it consistent.

PR #2976, opened 2021-08-11
Start bullet item with upper case letter

henrikt-ma:


henrikt-ma:
By the way, should it be clarified that this rule is about connecting to the entire expandable connector, that is, that the rule is not meant to forbid the typical uses like:
```
  connect(bus.speed, actuator.speed);
```

HansOlsson (approved):
Looks good to me

PR #2982, opened 2021-08-26
Small cleanup of scoping chapter

henrikt-ma:
Found a period in a section heading, thought I'd better convert the chapter to sentence-based before making actual changes, found more minor stuff while going through the document source file, and finally got to fix the section heading.

Please look at individual commit diffs rather than the total diff, as the first commit is conversion of line breaking and messes up the total diff.


HansOlsson (approved):
Looks ok.

PR #2993, opened 2021-09-07
Avoid just a '0' in \lstinline when generating HTML

### LINKED ISSUE #2991: Missing integer zero literal in section 2.4.2
The integer literal `0` is not properly rendered in the generated HTML:

https://github.com/modelica/ModelicaSpecification/blob/80f268aa5f113830fd85c74568381b8542dd273e/chapters/lexicalstructure.tex#L156

https://specification.modelica.org/master/lexical-structure.html#integer-literals

grafik

henrikt-ma:
Fixes #2991.  Actually, it's more of a temporary workaround and we'll have to return this once we've switched to a future version of LaTeXML where the upstream issue has been fixed.


HansOlsson (approved):
Looks good for now.
Will revert with updated latexml

PR #2995, opened 2021-09-08
Tick off "Restricted rules for use of 'start' attribute for parameter initialization"

henrikt-ma:
According to [yesterday's web meeting](https://github.com/modelica/ModelicaSpecification/pull/2748#issuecomment-914116630), I'm setting up this PR for documenting why we're ticking off this item on the road map.  The PR also contains a forgotten transfer of grammar between documents.

The use of the `start` attribute for parameter initialization is now restricted by means of the following:
- The `p.start` in full Modelica is replaced by the implicitly declared guess value parameter `guess('p')` in Flat Modelica.
- The parameter `'p'` and `guess('p')` are not allowed to end up in the same equation system.
- A parameter equation for `guess('p')` will constrain the equation systems to ensure that a guess value may be provided during initialization.
- The full Modelica rule about fixed parameters with `start` but no declaration equation is not carried over to Flat Modelica; the tool reducing full Modelica to Flat Modelica will be responsible for notifying the user and turning this into valid Flat Modelica parameter initialization.

As an example of the last item, the full Modelica
```
parameter Real p(start = 4.2);
```
should result in a warning during reduction to Flat Modelica, where it could look like this:
```
parameter Real 'p' = 4.2;
```

### DIFF
--- a/RationaleMCP/0031/ReadMe.md
+++ b/RationaleMCP/0031/ReadMe.md
@@ -60,7 +60,7 @@ These are subtopics that are considered necessary to resolve for a first version
 - [ ] Investigate need for `final`.
 - [ ] Origin of modifications (for start value prioritization).
 - [ ] Get rid of `false` as default for `fixed`.
-- [ ] Restricted rules for use of `start` attribute for parameter initialization.
+- [x] Restricted rules for use of `start` attribute for parameter initialization.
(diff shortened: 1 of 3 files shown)


olivleno:
New keyword: "guess", needs to be added [Henrik]

Avoid the notion of "same equation system" to avoid misunderstanding.
To clarify: equation system means a set of equations solved independently.
Proposal: Include a formulation that one needs to be solved before the other.
Poll on changing the formulation
in favour of: Hans, Gerd
abstain: Oliver, Kai, Martin, Henrik

decision:
Change the formulation with review Request to Hans and Gerd [Henrik].

After approval by Gerd and Hans, merge and close [Henrik]

henrikt-ma:
Now updated to avoid _in the same equation system_.

PR #2998, opened 2021-09-10
Turned equations into assignments in when-algorithm

casella:
I guess this was an oversight due to copy-and-paste that survived several versions of the spec. Maybe it's time to fix it now 😄


henrikt-ma (approved):
Looks good to me.

HansOlsson (approved):
Ok

PR #3000, opened 2021-09-15
Fix formatting of 'when', 'elsewhen', and 'when initial()'

henrikt-ma:


HansOlsson (approved):
Ok, formatting, but the parenthesis does not make sense. (Not a new issue.)

Currently it says:
"having \lstinline!reinit! of the same variable in \lstinline!when! and \lstinline!elsewhen! of the same variable is allowed"
That seems repetitive and the second "same variable" should be something else: that reinit of the same variable should be possible in both when and elsewhen of one when-equation.

henrikt-ma:
> Ok, formatting, but the parenthesis does not make sense. (Not a new issue.)
> 
> Currently it says:
> "having \lstinline!reinit! of the same variable in \lstinline!when! and \lstinline!elsewhen! of the same variable is allowed"
> That seems repetitive and the second "same variable" should be something else: that reinit of the same variable should be possible in both when and elsewhen of one when-equation.

Right, but how about taking this part in #3001?

PR #3002, opened 2021-09-17
Fix year in date reference

beutlich:
refs #2993


henrikt-ma (approved):
Well spotted, thanks!

PR #3011, opened 2021-10-05
Cleanup file hierarchy listings

henrikt-ma:
Initiating in _Draft_ state, as I also need to check for more listings that should use the new language.

### DIFF
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -2442,30 +2442,31 @@ package MyExternalFunctions
 end MyExternalFunctions;
 \end{lstlisting}
 Directory structure:
+{\upshape
 \begin{tabbing}
 \hspace{15em}\=\hspace{1em}\=\kill
 \filename{ExternalFunctions}\\
-\quad\filename{package.mo}                        \>-- \>\emph{Modelica code from above}\\
+\quad\filename{package.mo}                        \>-- \>Modelica code from above\\
 \quad\filename{Resources}\\
-\quad\quad\filename{Include}                      \>-- \>\emph{Include files}\\
-\quad\quad\quad\filename{ExternalFunc1.h}         \>-- \>\emph{C header file}\\
-\quad\quad\quad\filename{ExternalFunc2.h}         \>-- \>\emph{C header file}\\
-\quad\quad\quad\filename{ExternalFunc3.c}         \>-- \>\emph{C source file (not ideal)}\\
-\quad\quad\filename{Library}                      \>-- \>\emph{Object libraries for different platforms}\\
+\quad\quad\filename{Include}                      \>-- \>Include files\\
+\quad\quad\quad\filename{ExternalFunc1.h}         \>-- \>C header file\\
+\quad\quad\quad\filename{ExternalFunc2.h}         \>-- \>C header file\\
+\quad\quad\quad\filename{ExternalFunc3.c}         \>-- \>C source file (not ideal)\\
+\quad\quad\filename{Library}                      \>-- \>Object libraries for different platforms\\
 \quad\quad\quad\filename{win32}\\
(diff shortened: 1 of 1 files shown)


beutlich (commented):
Note that, even it is mono-space in TeX source it is not in he rendered HTML, which is especially apparent for the indented comment.

grafik

henrikt-ma:
> Note that, even it is mono-space in TeX source it is not in the rendered HTML, which is especially apparent for the indented comment.

Yes, I noted that too, but couldn't think of a quick fix for this.  I'm noting that our filenames don't look quite right in the PDF either, with the variable-width glyphs being placed on a grid.  Let's see if we can fix it as part of this PR.  I can make an attempt with the `tabbing` environment instead, and see how that goes with LaTeXML.  Any other ideas?

beutlich:
> Any other ideas?

TeXExchange says dirtree is my friend: https://tex.stackexchange.com/questions/5073/making-a-simple-directory-tree

henrikt-ma:
> TeXExchange says dirtree is my friend: https://tex.stackexchange.com/questions/5073/making-a-simple-directory-tree

Looks very pretty, but I see no signs of it being supported by LaTeXML.

dietmarw:
@henrikt-ma Have you already tried it in combination with LaTeXML?

henrikt-ma:
Apart from a LaTeXML glitch (to be reported), the result with a `tabbing` environment doesn't look too bad to me:
tabbing-html

The pdfLaTeX output shows my intention:
tabbing-pdf

beutlich:
From the very same TeXExchange reference: forest package can also do it in a nice way.

henrikt-ma:
> From the very same TeXExchange reference: forest package can also do it in a nice way.

But did you see any sign of it being supported by LaTeXML?

sjoelund:
> > From the very same TeXExchange reference: forest package can also do it in a nice way.
> 
> But did you see any sign of it being supported by LaTeXML?

Some packages work without any intervention, right? But the most problematic parts are things like verbatim text, which I assume this is... Might be good to try

henrikt-ma:
> @henrikt-ma Have you already tried it in combination with LaTeXML?

No, and as @sjoelund suggests, it seems to be worth a try even though I didn't find any signs of support when searching the web.

henrikt-ma:
OK, after verifying that the PDF looks as expected with _dirtree.sty_, this is how it fails with LaTeXML:
```
Warning:missing_file:dirtree Can't find binding for package dirtree
	at mini.tex; line 3 col 6 - line 3 col 6
	Anticipate undefined macros or environments
	search paths are /Users/henrikt/Documents/minimal-LaTeXML
	In Core::Definition::Constructor[\usepac... /opt/local/lib/perl5/vendor_perl/5.28/LaTeXML/Package/LaTeX.pool.ltxml; line 798

Error:undefined:\dirtree The token T_CS[\dirtree] is not defined.
	at mini.tex; line 4 col 8 - line 4 col 8
	Defining it now as <ltx:ERROR/>
	Next token is T_BEGIN[{]
	In Core::Gullet[@0x7faf9901d500] /Users/henrikt/Documents/minimal-LaTeXML/mini.tex; from line 4 col 9 to line 4 col 9
	 <= Core::Stomach[@0x7faf9901cdb0] <= ...
 4.38 sec) 4.85 sec)
```

Considering that _forest.sty_ provides a _PGF/TikZ-based mechanism for drawing_, is it even interesting for us?

henrikt-ma:
> Apart from two lines it looks good. Would it be possible to change them to avoid it?

Yes.  Done.

henrikt-ma:
Removing _Draft_ state as I couldn't find any other file hierarchy listings to clean up.

PR #3027, opened 2021-11-12
Fix problems with 'parenthesis' in the singular

henrikt-ma:


chapters/operatorsandexpressions.tex
```diff
@@ -104,7 +104,7 @@ \section{Operator Precedence and Associativity}\label{operator-precedence-and-as
 Non-associative exponentiation and array range operator:
 \begin{lstlisting}[language=modelica]
 x ^ y ^ z     // Not legal, use parentheses to make it clear.
-a:b:c:d:e:f:g // Not legal, and scalar arguments gives no legal interpretation.
+a : b : c : d // Not legal, and scalar arguments gives no legal interpretation.
```
eshmoylova, line 107: There is also a discrepancy between plural and singular here: "gives" should be used for singular, but "arguments" are plural. Do you know what that part of the sentence tries to say? Would it be legal if the arguments were vectors?

henrikt-ma: Grammatically, I think the sentence should be read as an abbreviation of this to fit within the line:
> Not legal, and even having scalar arguments gives no legal interpretation.

So it is the _even having_ that isn't in the plural.

But what the comment really wants to convey, I don't know.

eshmoylova: I see. Hmmm, so there may be something missing that would make it grammatically correct and give meaning. Right now, my mind just hitches on the discrepancy and then cannot make any sense out of it. So I really don't want to leave it as is. I'll start a new issue, since it's not really what you changed. That would hopefully attract some other comments.

eshmoylova (approved):

HansOlsson (approved):
Looks good.

PR #3043, opened 2021-11-21
Tighten connection between \firstuse and document index

### LINKED ISSUE #3041: "!" in "constant!variable"
There is "!" in definitions of "constant!variable" in the code
https://github.com/modelica/ModelicaSpecification/blob/7ad198650e8ac01f6e50c700ba3e30e79eb335b4/chapters/classes.tex#L344
and it shows up in HTML and PDF too. This is from HTML:
image

Is that some artifact of `\firstuse` definition? Or is it really intended to be called constant!variable? Same for parameter!variable.

henrikt-ma:
Fixes #3041.

More importantly, this instruments `\firstuse` with capabilities for adding introduced terminology to the index.  The point is to strengthen the connection between introducing terminology in the text and making the terminology searchable in the document index.  By default, the argument to `\firstuse` is added as-is to the index, but since the desired appearance in the index is often not exactly the same as the appearance in the text, `\firstuse` also takes an optional argument for specifying how the terminology should be presented in the index.  Finally, on rare occasions one doesn't want an entry in the index at all (for example if a word is already present in the form of a keyword), and then one can pass an em-dash as the optional argument (for example, `\firstuse[---]{partial}`).

To support reviewers, I made a diff of the `MLS.idx` to show the effect of the commit _Make \firstuse add to document index, with possibility to override_:
```
10c10
< \indexentry {literal|hyperpage}{11}
---
> \indexentry {literal (constant)|hyperpage}{11}
110c110
< \indexentry {named elements|hyperpage}{44}
---
> \indexentry {named element|hyperpage}{44}
213a214
> \indexentry {derived from|hyperpage}{81}
221c222
< \indexentry {replaceable@\robustinline {replaceable}|hyperpage}{84}
---
> \indexentry {replaceable@\robustinline {replaceable}|hyperpage}{85}
263c264
< \indexentry {connection equations|hyperpage}{117}
---
> \indexentry {connection equation|hyperpage}{117}
```

Beside the page number change (possibly because of insufficient pdfLaTeX recompilation), the changes seen are intentional and the result of manually going through all occurrences of `\firstuse`.


chapters/scoping.tex
```diff
@@ -55,7 +55,7 @@ \section{Static Name Lookup}\label{static-name-lookup}
 \subsection{Simple Name Lookup}\label{simple-name-lookup}
 
 % Not adding 'encapsulated' as normal word to index; the keyword variant seems sufficient
-A class declared with the keyword \lstinline!encapsulated!\indexinline{encapsulated} (see \lstinline[language=grammar]!class-definition! in the grammar) is called an \firstuse{encapsulated} class.
+A class declared with the keyword \lstinline!encapsulated!\indexinline{encapsulated} (see \lstinline[language=grammar]!class-definition! in the grammar) is called an \firstuse[---]{encapsulated} class.
```
eshmoylova, line 58: That shows up as "is called an encapsulatedNOTHING class." in [the Jenkins build](https://test.openmodelica.org/jenkins/job/ModelicaAssociation/job/ModelicaSpecification/view/change-requests/job/PR-3043/).

eshmoylova (changes requested):
I looked at a couple of changes and how they look in the build by Jenkins. It appends CUSTOM and NOTHING to the text. I am not sure if the rule is not defined correctly or something went wrong with the build. Could you check what you get in your build and if something needs to be updated? Also styleguide.md should be updated to document the new options.

chapters/annotations.tex
```diff
@@ -198,7 +198,7 @@ \subsubsection{Plot Curves}\label{plot-curves}
 
 \subsubsection{Escape sequences}\label{text-markup-escape-sequences}
 
-In an attribute inside a figure where the variable replacements of \cref{variable-replacements} or the text markup of \cref{text-markup-in-captions} can be used, the following use of \firstuse{text markup escape sequences}\index{text markup escape sequence}\index{escape sequence!text markup} applies.
+In an attribute inside a figure where the variable replacements of \cref{variable-replacements} or the text markup of \cref{text-markup-in-captions} can be used, the following use of \firstuse[text markup escape sequence]{text markup escape sequences}\index{escape sequence!text markup} applies.
```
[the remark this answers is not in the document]
henrikt-ma: Oh no, I was so focused on creating the _MLS.idx_ file diff so that I forgot about those "debug prints".  I'll fix them shortly.

chapters/scoping.tex
henrikt-ma: Same as above.

chapters/annotations.tex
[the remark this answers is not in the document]
henrikt-ma: Fixed.

chapters/scoping.tex
henrikt-ma: Fixed.

eshmoylova (commented):
I did not check every item (maybe 15 or so, at random that have different styles). Looks good apart from the comment about encapsulated (but may be it needs to be a separate issue?) and also an update to the styleguide.

henrikt-ma:
> Also styleguide.md should be updated to document the new options.

How about [this](https://github.com/henrikt-ma/ModelicaSpecification/blob/cleanup/firstuse/styleguide.md#definitions)?

chapters/scoping.tex
henrikt-ma: The reason in this case is that we already have `\indexinline{encapsulated}` in the same paragraph, and it seemed a bit excessive to both have entries for the keyword and the natural language word pointing to the same paragraph.  Don't you think?

styleguide.md
```diff
@@ -155,9 +155,12 @@ base-clock | `base-clock` | Similarly: sub-cock
...
-As a general rule, `\firstuse` should appear in combination with `\index` for adding the term to the document index.
-(The use of `\firstuse` instead of just `\emph` helps us both produce consistent formatting and makes it easier to spot cases where the additional use of `\index` has been forgotten.
-The reason that `\firstuse` doesn't also do the job calling `\index` is that the form of the term presented to `\firstuse` isn't always in the base form expected in the document index, that there can be a need for special styling tricks in the `\index` argument, etc.)
+As a general rule, terminology introduced with `\firstuse` should appear in the document index, and by default the mandatory argument to `\firstuse` is automatically passed to `\index`.
+To change the appearance of the index entry, the default may be overridden using an optional argument to `\firstuse`, for example, `\firstuse[array!variable]{array variable}`.
+This is also useful when capitalization or plural/singular differs; excapt for things like names, lower case should be used in the index, and terms should typically appear in the singular, for example, `\firstuse[vector]{Vectors}`.
```
eshmoylova, line 160: Typo:
```suggestion
This is also useful when capitalization or plural/singular differs; except for things like names, lower case should be used in the index, and terms should typically appear in the singular, for example, `\firstuse[vector]{Vectors}`.
```

chapters/classes.tex
```diff
@@ -202,7 +202,7 @@ \subsection{Component Declaration Static Semantics}\label{component-declaration-
 
 \subsubsection{Declaration Equations}\label{declaration-equations}
 
-An environment that defines the value of a component of built-in type is said to define a \firstuse{declaration equation}\index{declaration equation}\index{declaration equation!zzzzz@\igobble |seealso{binding equation}} associated with the declared component.
+An environment that defines the value of a component of built-in type is said to define a \firstuse{declaration equation}\index{declaration equation!zzzzz@\igobble |seealso{binding equation}} associated with the declared component.
```
eshmoylova, line 205: Should this mechanism for adding see also be documented in the styleguide as well? Or is it something that everybody knows?

chapters/scoping.tex
eshmoylova: Oh, sorry. I missed that there were three entries in the index (HTML):
image
Actually, in PDF there are only two entries:
image
I guess two of them on the same page, so they got merged?

Right, I first looked at the pdf and tried to follow the links there. But the pdf shows the whole page when you click on the link, and I wanted to see the exact place. So I switched to the HTML. I clicked on the first two links because in my mind there were only two, and missed the third one completely.

Is there a way to sort the entries in the index? The first item on that list is in Section 5.3.2, the second in  5,3,3, and the third in 5.3.1 is this paragraph above. It would make more sense to have 5.3.1 to be the first one.

eshmoylova (changes requested):

chapters/classes.tex
henrikt-ma: To me, this is more of a technicality of the _index.sty_ (or _makeidx.sty_) package than content for a style guide.  But maybe our style guide needs to be both a style guide and a technical guide for how to write LaTeX code?

I mean, describing how to write the source code in order to maintain good style is one thing that has a clear place in the style guide.  It's when it becomes technical about things not directly related to style that I hesitate.

henrikt-ma: By the way, this is not a trick I assume everybody knows, and it took me some time to find it.  However, I'm afraid that the LaTeX world (and even just the packages more or less supported by LaTeXML) is an endless source of the need for such tricks.  If we had a document with our own selection of tips and tricks, would anyone read it?  My guess is that most authors would just look for similar situations in existing code of the document, and not even remember that there's a collection of tips and tricks.  A style guide, on the other hand, has a normative role and gives us a place for having discussions about what we want the document to look like.

henrikt-ma:
(I still don't know how you do it, but I can't figure out how to reply directly to the comments.)

> Oh, sorry. I missed that there were three entries in the index (HTML):
> image
> Actually, in PDF there are only two entries:
> image
> I guess two of them on the same page, so they got merged?

Yes, I believe it should work that way.

> Right, I first looked at the pdf and tried to follow the links there. But the pdf shows the whole page when you click on the link, and I wanted to see the exact place. So I switched to the HTML. I clicked on the first two links because in my mind there were only two, and missed the third one completely.
>
> Is there a way to sort the entries in the index? The first item on that list is in Section 5.3.2, the second in 5,3,3, and the third in 5.3.1 is this paragraph above. It would make more sense to have 5.3.1 to be the first one.

Not that I know, unfortunately.  I suppose it could be the topic of a new LaTeXML issue.

eshmoylova:
> (I still don't know how you do it, but I can't figure out how to reply directly to the comments.)

I usually go to the Files tab and look through the comments there (Ctrl + f can be helpful when like in this change there are so many files).  I cannot make sense sometimes of how the Conversation tab combines comments especially if they are many comments that are part of one "Review"). If I see a comment in an email that I cannot find in the Files tab I search through the Conversation.

> > Is there a way to sort the entries in the index? The first item on that list is in Section 5.3.2, the second in 5,3,3, and the third in 5.3.1 is this paragraph above. It would make more sense to have 5.3.1 to be the first one.
> 
> Not that I know, unfortunately. I suppose it could be the topic of a new LaTeXML issue.

Maybe it's irrelevant because the assumption is that you click on all items.

chapters/scoping.tex
eshmoylova: The discussion got moved to the Converstation tab. This is not really a problem with the PR, so resolving.

eshmoylova (approved):

chapters/classes.tex
eshmoylova: > By the way, this is not a trick I assume everybody knows, and it took me some time to find it. 

That's why I asked. It had the look of a trick/hack that required some labor to come up with.

> However, I'm afraid that the LaTeX world (and even just the packages more or less supported by LaTeXML) is an endless source of the need for such tricks. If we had a document with our own selection of tips and tricks, would anyone read it? My guess is that most authors would just look for similar situations in existing code of the document, and not even remember that there's a collection of tips and tricks. A style guide, on the other hand, has a normative role and gives us a place for having discussions about what we want the document to look like.

That is a good point. There are not so many people who currently author the content, unfortunately. ( Ahem, guilty myself.) Maybe we need to first have a broader discussion of how to attract more people to making the changes and what is holding them up? @HansOlsson, should we put it on the agenda for the next Phone meeting?

HansOlsson: Would make sense to discuss yes, but I would think that other issues that LaTeX-hacks are more blocking.
When adding them I generally add references to them (e.g. to tex.stackexchange.com).

I also think that we should avoid such hacks when possible  - and an alternative in this case would be that the text of "Declaration Equation" section would say that it is a sub-set of "Binding Equations" with reference; instead of using the hack to put "see also" in the index.

HansOlsson, line 205:
```suggestion
An environment that defines the value of a component of built-in type is said to define a \firstuse{declaration equation} associated with the declared component. These are a subset of the binding equations, see \cref{equation-categories}.
```
(I hope I got the section name right.

chapters/connectors.tex
```diff
@@ -871,7 +871,7 @@ \subsection{Overconstrained Equation Operators for Connection Graphs}\label{over
 Connections.root(A.R)
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
-The overdetermined type or record instance \lstinline!R! in connector instance \lstinline!A! is a (definite) \firstuse{root node}\index{definite root node}\index{root node!definite} in a virtual connection graph.
+The overdetermined type or record instance \lstinline!R! in connector instance \lstinline!A! is a (definite) \firstuse[definite root node]{root node}\index{root node!definite} in a virtual connection graph.
```
HansOlsson, line 874: May "(definite)" should be inside firstuse, as the index entry is definite root node.

HansOlsson (commented):
Looks good except for the specific comments.

chapters/classes.tex
HansOlsson: For background see previous thread: https://github.com/modelica/ModelicaSpecification/pull/3043#discussion_r757119274_

henrikt-ma: Having lots of LaTeX hacks in the source is of course not good per se, but on the other hand I think that having _see also_ entries in the index is a good way of making the index more useful.  In this regard, having examples in the source of how it can be done has value in itself, and for this reason I'd like to keep this one as it is.

henrikt-ma: Can we resolve this and continue in https://github.com/modelica/ModelicaSpecification/pull/3043#discussion_r757120747 to try to keep the discussion in one place?

chapters/connectors.tex
henrikt-ma: Sure, why not?  Done.  OK?

chapters/classes.tex
eshmoylova: Resolving to continue the discussion in the comment as suggested above.

eshmoylova: One thing that looks odd with the way it is done currently is the comma at the beginning of the line "see also" line:
image
This is in HTML. PDF looks good. I thought it was meant to be after "2nd item" on the line above but somehow it escaped to the next line, but in PDF there is no comma on the line above and no in front of "see also".

HansOlsson: To me having the "see" in the actual text makes sense here; and I think we should first consider that.

I can understand that we might want to describe this hack somewhere, but keeping it for that reason this doesn't seem like a good solution. Can't we put a link to this hack in some common area (or possibly as a comment in preamble) instead?

If someone were to add "seealso" I wouldn't expect them to scan the entire document to find uses.

henrikt-ma: OK.  I'll do it.

henrikt-ma: Done.  OK now?

HansOlsson (approved):
Looks good.

PR #3061, opened 2021-12-08
Fix minor issues in chapter 'Connectors and Connections'

henrikt-ma:
Initiated by a coworker that brought the bad line breaks in a listing to my attention.

Note that most changes of this PR are about finishing the transition to sentence-based line breaking in the affected chapter; please see individual commits to isolate actual changes of content.


HansOlsson (approved):
Looks good.
The bad line-breaks may have been due to conversion to LaTeX.

PR #3085, opened 2021-12-23
Fix indentation of Ellipse pseudo-code record definition

henrikt-ma:


HansOlsson (approved):
Looks better than before

PR #3086, opened 2021-12-23
Address fancyhdr warning about too small 'headheight'

henrikt-ma:
I just made a fresh installation of texlive, and without this I'm getting a warning from fancyhdr on every page, making all other messages drown completely in the fancyhdr noise.


HansOlsson (approved):
Seems to make sense.

PR #3088, opened 2022-01-16
Clean up around parameter initialization example

henrikt-ma:
Cleanup of example, as decided in https://github.com/modelica/ModelicaSpecification/pull/3035#issuecomment-972723480, plus a couple of more nearby cleanup changes.


HansOlsson (approved):
Looks ok.
(My space-preferences are clearly different.)

PR #3100, opened 2022-01-25
Further improve short section names

henrikt-ma:
After #3073 was merged, I realized that I still had pending review comments.  As I was turning these into concrete suggestions I also made a sweep to see if there were other section headings one could shorten to further improve the readability of the table of contents.

This is all done in the spirit of #3073, addressing a problem described in #2825.


chapters/arrays.tex
```diff
@@ -1010,7 +1010,7 @@ \section{Indexing}\label{array-indexing}\label{indexing}
 \end{center}
 \end{table}
 
-\subsection{Indexing with Boolean or Enumeration Values}\label{indexing-with-boolean-or-enumeration-values}
+\subsection{Boolean or Enumeration Indices}\label{boolean-or-enumeration-indices}
```
HansOlsson, line 1013: Ok, although it loses the similarity to "Indexing with end"

HansOlsson: ('end Index' looks ugly. The best I could think of was 'Using end Index', but it is not so similar.)

chapters/connectors.tex
```diff
@@ -671,7 +671,7 @@ \subsection{Balancing Restriction and Size of Connectors}\label{balancing-restri
 \end{lstlisting}
 \end{example}
 
-\section{Overconstrained Connection-Based Equation Systems}\label{equation-operators-for-overconstrained-connection-based-equation-systems1}\label{overconstrained-connection-based-equation-systems}
+\section{Overconstrained Connectors}\label{equation-operators-for-overconstrained-connection-based-equation-systems1}\label{overconstrained-connectors}
```
HansOlsson, line 674:
```suggestion
\section{Overconstrained Connections}\label{equation-operators-for-overconstrained-connection-based-equation-systems1}\label{overconstrained-connections}
```
To me it doesn't make sense to say that a connector is overconstrained in itself. (At most it is "constrained").
Especially as the problem only occurs if there are loops.

Clearly the previous variant was a bit too much, and I think that "Overconstrained Connections" is both short and somewhat accurate.

HansOlsson: (Ah well, and then we define 'overconstrained connectors' later on. I still believe we shouldn't use it here, and that we maybe later rename them.)

chapters/synchronous.tex
```diff
@@ -1026,7 +1026,7 @@ \subsection{Sub-Clock Inferencing}\label{sub-clock-inferencing}
 64 bit internal representation of numerator and denominator with sign can be used and gives minimum resolution $1.08\times 10^{-19}$ seconds and maximum range $9.22\times 10^{18}$~seconds = $2.92\times 10^{11}$~years.
 \end{nonnormative}
 
-\section{Clocked-Discretized - Continuous-Time Equations in Clocked Partitions}\label{continuous-time-equations-in-clocked-partitions}\label{clocked-discretized-continuous-time-equations-in-clocked-partitions}
+\section{Clocked Partition Equation Solving}\label{continuous-time-equations-in-clocked-partitions}\label{clocked-discretized-continuous-time-equations-in-clocked-partitions}
```
HansOlsson, line 1029: I don't think that clearly represents the contents - the section is focused on using solverMethod to discretize equations.
Originally it was "Continuous-Time Equations in Clocked Partitions", and then changed to add "Clocked-Disretized - ", since "continuous-time" wasn't fully correct.
I would thus prefer "Clocked-Discretized" or "Clocked-Discretized Partition".
```suggestion
\section{Clocked-Discretized Partition}\label{continuous-time-equations-in-clocked-partitions}\label{clocked-discretized-partition}
```

chapters/arrays.tex
henrikt-ma: To me it didn't look too problematic that the similarity was lost.  The fact that the new section names are concise makes them look good anyway.

chapters/connectors.tex
henrikt-ma: I agree that the connectors by themselves are only constrained, so I'll just go with your suggestion for now.

chapters/synchronous.tex
henrikt-ma: Before committing the suggestion, I'd like to get confirmation that it really is a better name considering that it is also under this heading that _clocked discrete-time partition_ is defined.

(And I guess there shouldn't be any hyphen in _clocked discretized partition_, right?)

HansOlsson: There's another discussion of having better names for this #3091

HansOlsson: Even if clocked discrete-time partition is defined here, it is just as the complement to clocked discretized; which is the main focus of the section - but as noted in #3091 it is not defined in a good way.

HansOlsson (approved):
Ok

eshmoylova (changes requested):
Looks good except for one section in the Connectors and Connections

chapters/connectors.tex
```diff
@@ -863,7 +863,7 @@ \subsection{Operators for Connection Graphs}\label{overconstrained-equation-oper
 \end{nonnormative}
 
 
-\subsection{Converting the Connection Graph into Trees and Generating Connection Equations}\label{converting-the-connection-graph-into-trees-and-generating-connection-equations}
+\subsection{Generation of Connection Graph Equations}\label{generation-of-connection-graph-equations}
```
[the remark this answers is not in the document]
HansOlsson: I think the last one would work.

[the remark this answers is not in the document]
henrikt-ma: I was trying to have some similarity with the section 9.2 _Generation of Connection Equations_, and to me the suggested _Connection Graph Equations_ is meant to have the same meaning as _Equations from Connection Graphs_.  For comparison, we didn't say _Generation of Equations from Connect Equations_.

How about just changing the first one to _Connection Graphs and Their Operators_?  (This still a shorter title than the one below, so the length of it feels acceptable.)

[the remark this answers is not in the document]
eshmoylova: That is the dilemma: Have a better fitting name or a shorter one? I like *Connection Graphs and Their Operators* but *Operators for Connection Graphs* is shorter and fits fully into the menu. I would still prefer *Equations from Connection Graphs* instead of *Connection Graphs Equations* because it suggests to me more of the process. *Connection Graph Equations* suggests that it's a concept in its own right, which it is not.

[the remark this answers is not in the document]
henrikt-ma: Since you say you like _Connection Graphs and Their Operators_ and this is still doesn't extend further to the right than two other headings in the same chapter, I went with this change.

Regarding the _Generation of Connection Graph Equations_, I tried to fix this by making _Connection Graph Equations_ just as much as concept as _Connection Equation_ and _Stream Connection Equations_.  What do you think of that?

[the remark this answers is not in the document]
eshmoylova: OK. Resolving.

eshmoylova (approved):

PR #3104, opened 2022-01-28
Version annotation indent

henrikt-ma:
Just a small cleanup of listings related to library version handling.  (I was surprised to find out that this wasn't already done in #2957, which I will also make sure is up to date by resolving its merge conflicts.)


HansOlsson (approved):
ok

PR #3105, opened 2022-01-31
Fix indentation of listings related to versionDate and friends

henrikt-ma:
Following same style of indentation as in the section "Version Handling" above.


HansOlsson (approved):
Ok, even if the two uses ones don't seem that necessary.

PR #3113, opened 2022-02-08
Changed y into x in spatialDistribution() definitions

casella:
Addresses the issue raised by @eshmoylova in #3111

### DIFF
--- a/chapters/operatorsandexpressions.tex
+++ b/chapters/operatorsandexpressions.tex
@@ -810,18 +810,18 @@ In this case, the internal state of the operator is hidden from the ODE solver.
 
 \lstinline!spatialDistribution! allows the infinite-dimensional problem below to be solved efficiently with good accuracy
 \begin{align*}
-\frac{\partial z(y,t)}{\partial t}+v(t)\frac{\partial z(y,t)}{\partial y} &= 0.0\\
+\frac{\partial z(x,t)}{\partial t}+v(t)\frac{\partial z(x,t)}{\partial x} &= 0.0\\
 z(0.0, t) &= \mathrm{in}_0(t) \text{ if $v\geq 0$}\\
 z(1.0, t) &= \mathrm{in}_1(t) \text{ if $v<0$}
 \end{align*}
-where $z(y, t)$ is the transported quantity, $y$ is the normalized spatial coordinate ($0.0 \le y \le 1.0$), $t$ is the time, $v(t)=\mathrm{der}(x)$ is the normalized transport velocity and the boundary conditions are set at either $y=0.0$ or $y=1.0$, depending on the sign of the velocity.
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Looks good I think.
One minor issue is that we have two variants of `x` - both the total coordinate for the material and the part within the "plug"; but having different unrelated names for them seems to cause more confusion.

chapters/operatorsandexpressions.tex
```diff
@@ -810,26 +810,26 @@ \subsubsection{spatialDistribution}\label{spatialdistribution}
...
 The calling syntax is:
 \begin{lstlisting}[language=modelica]
 (out0, out1) = spatialDistribution(in0, in1, x, positiveVelocity,
                                    initialPoints = {0.0, 1.0},
                                    initialValues = {0.0, 0.0});
 \end{lstlisting}
-where \lstinline!in0!, \lstinline!in1!, \lstinline!out0!, \lstinline!out1!, \lstinline!x!, \lstinline!v! are all subtypes of \lstinline!Real!, \lstinline!positiveVelocity! is a \lstinline!Boolean!, \lstinline!initialPoints! and \lstinline!initialValues! are arrays of subtypes of \lstinline!Real! of equal size, containing the y coordinates and the $z$ values of a finite set of points describing the initial distribution of $z(y, \mathit{t0})$.
+where \lstinline!in0!, \lstinline!in1!, \lstinline!out0!, \lstinline!out1!, \lstinline!x!, \lstinline!v! are all subtypes of \lstinline!Real!, \lstinline!positiveVelocity! is a \lstinline!Boolean!, \lstinline!initialPoints! and \lstinline!initialValues! are arrays of subtypes of \lstinline!Real! of equal size, containing the x coordinates and the $z$ values of a finite set of points describing the initial distribution of $z(x, \mathit{t0})$.
```
eshmoylova, line 824: One issue that was brought up by @henrikt-ma  in #3111, which I think is worth including in this PR is the mentions on `v` here. Yes, `v` is used elsewhere in the text. But when the sentence starts with "where in0, in1,..." I assume that it defined the quantities used in the formula that precedes this statement. So I also find it confusing that `v` is mentioned here but doesn't actually appear in the formula above. For the general context `v` is already explained on line 817 and how it is actually used with `spatialDistribution` appears only in the examples at the end.

casella:
> Looks good I think. One minor issue is that we have two variants of `x` - both the total coordinate for the material and the part within the "plug"; but having different unrelated names for them seems to cause more confusion.

@HansOlsson, I'm sorry but I don't understand what you are referring to. What are the "total coordinate for the material" and the "plug"? I understand there is only one x here, which is explicitly defined as a normalized spatial coordinate.

casella:
@eshmoylova, I'm afraid I messed up with your review, there are so many features popping up in github all the time and I'm getting a bit lost. I think I addressed your last issue, if so can you please close your review so we can merge this in? Otherwise, I'm afraid we'll still have mixed x and y in Modelica 3.8 😄

Thanks!

PR #3121, opened 2022-02-22
A class is not a specialized class

beutlich:
According to https://github.com/modelica/ModelicaStandardLibrary/pull/2500 and section 4.6 a (specialized) class cannot be a simulation model.

### DIFF
--- a/chapters/introduction.tex
+++ b/chapters/introduction.tex
@@ -10,7 +10,7 @@ It provides object-oriented constructs that facilitate reuse of models, and can 
 The semantics of the Modelica language is specified by means of a set of rules for translating any class described in the Modelica language to a flat Modelica structure.
 The semantic specification should be read together with the Modelica grammar.
 
-A class (of specialized class \lstinline!model!, \lstinline!class! or \lstinline!block!) intended to be simulated on its own is called a \firstuse{simulation model}.
+A class (of specialized class \lstinline!model! or \lstinline!block!) intended to be simulated on its own is called a \firstuse{simulation model}.
 
 The flat Modelica structure is also defined for other cases than simulation models; including functions (can be used to provide algorithmic contents), packages (used as a structuring mechanism), and partial models (used as base-models).
 This allows correctness to be verified for those classes, before using them to build the simulation model.


henrikt-ma:
> According to [modelica/ModelicaStandardLibrary#2500](https://github.com/modelica/ModelicaStandardLibrary/pull/2500) and section 4.6 a (specialized) class cannot be a simulation model.

I fail to see what you are referring to in 4.6, could you provide a quote?

beutlich:
Section 4.6 says that a specialized class is record, type, model, block, package, function and connector, but not class itself.

henrikt-ma:
OK, but then there is currently no real support in the specification for not considering a `class` a simulation model – we only have a statement in ModelicaReference to lean on at this point.  If ModelicaReference is ignored since it should just be repeating the rules of Modelica 3.4, an alternative resolution could be to allow `class` but without referring to it as a specialized class.

Apart from this consideration, I am not against the change per se.  I just want it to be properly discussed by the language group before it gets applied.

beutlich:
> Accept the PR, I think this is the best.

This is also my preference.

henrikt-ma (commented):
So far, all comments speak in favor of accepting the suggested change.  This should make it easy for the language group to make up its mind if this is added to the agenda of the next phone meeting.

eshmoylova:
I agree that `class` is not really listed as specialized class as per section 4.6, but I am not so certain that `class` cannot be used as a simulation model.  From other places in the specification, it seems to me that `class` us supposed to be a generic Modelica class. Maybe at some point `class` was supposed to be like the `Object` superclass in Java? From what I can see, the way it is now in the specification, the difference between `model` and `class` is that a `model` can be a base class only for another `model` but a `class` can be a base class for any other type of class (7.1.3) and equations and variables in a `model` must be balanced (4.7). It says in 4.6 that `model` is "The normal modeling class in Modelica." I interpret it as that's what people usually use but it does not have to be a model. There are examples where a `class` has declarations and equations (4.5, 7.1, 7.3.2). So even though ModelicaReference says that `class` should be used for ExternalObject and and can be used for classes merely containing documentation and/or graphics, there is nothing in the specification that says it has to be limited to those things only. If we want to enforce the rule from ModelicaReference it would need to be updated in many more places than in the introduction. Also that would make `class` a specialized class in a sense. It is interesting that, as it stands now, a `class`, `model`, or `block` can be used as a simulation model but the restriction on equations and variables to be balanced is specified for `model` and `block` only (4.7).

henrikt-ma:
> Also that would make class a specialized class in a sense.

While we used to call them _class restrictions_, we now speak of _class specialization_ instead.  Hence, instead of thinking of `class` as being restricted to not be a valid simulation model, we now (according to this proposal) think of `model` and `block` having the special feature of allowing simulation.

> There are examples where a class has declarations and equations (4.5, 7.1, 7.3.2).

To me it seems worth the effort to look for places in the specification where a `class` is currently used as simulation model, and change these to `model` or `block`.

I actually couldn't find anything in need of change in 4.5.  In 7.1 and 7.3.2, I think changing to `model` would make sense since `class` should mainly be used for external objects.

HansOlsson (approved):
LGTM.
Will ensure that it is voted on just to be sure.

henrikt-ma:
Some more input to the language group when it comes to voting on this: It seems rather clear already today that `class` isn't useful for simulation models:
> All specialized classes can be derived from \lstinline!class!\indexinline{class}, provided that the resulting class fulfills the restriction of the specialized class.
> A \lstinline!class! may only contain class definitions, annotations, and \lstinline!extends!-clauses (having any other contents is deprecated).

I mean, what would you simulate if there aren't any component declarations?

henrikt-ma:
I think we should also address the problem bought up by @eshmoylova, namely that `class` is currently used in incorrect/misleading ways.  For example, one find things like this in 7.1:
```
class A
  parameter Real a, b;
end A;
```
Not very good, considering that the component declaration isn't a class definition.

Similarly, 7.3.2 contains examples like this:
```
class A
  parameter Real x;
end A;
```

To further reduce the confusion about `class`, it would help to not only state in 12.9.7 that external objects is the only use of `class`, but to also mention this observation non-normatively in 7.3.1 (with a link to 12.9.7).

PR #3160, opened 2022-04-27
Clarify odd cases of IconMap and DiagramMap 'extent'

### LINKED ISSUE #3150: What is the definition of "null region"
Section [18.6.3 Extends-Clause](https://specification.modelica.org/master/annotations.html#extends-clause) says:

> If the extent of the extends-clause defines a null region (the default), the base class contents is mapped to the same coordinates in the derived class, and the coordinate system (including preserveAspectRatio) can be inherited as described in [section 18.6.1.1](https://specification.modelica.org/master/annotations.html#coordinate-systems).

But "null region" is not defined anywhere. I can see 4 possibilities for its meaning:
1. A region with zero area (i.e. x1 = x2 _or_ y1 = y2)
2. A point-shaped region (i.e. x1 = x2 _and_ y1 = y2)
3. Exactly the default value (i.e `{{0, 0}, {0, 0}}`)
4. The annotation does not contain an extent, so the default value is used.

What is the intended interpretation?

I would interpret it as nr 1, but such extents are allowed for `Text`, making it somewhat weird to call that a "null region".

Also, [18.6.1.1 Coordinate Systems](https://specification.modelica.org/master/annotations.html#coordinate-systems) uses "null-region" instead of "null region".

henrikt-ma:
Fixes #3150.

According to [phone meeting decision](https://github.com/modelica/ModelicaSpecification/issues/3150#issuecomment-1109892842).

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -742,11 +742,16 @@ If the \lstinline!primitivesVisible! attribute is false, components and connecti
 
 \begin{itemize}
 \item
-  If the extent of the \lstinline!extends!-clause defines a null region (the default), the base class contents is mapped to the same coordinates in the derived class, and the coordinate system (including \lstinline!preserveAspectRatio!) can be inherited as described in \cref{coordinate-systems}.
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Seems ok.

PR #3161, opened 2022-04-29
New annotation for evaluated parameters

henrikt-ma:


henrikt-ma:
Full Modelica model to illustrate the problem of what to express through annotations:
```
model M
  parameter Integer p = 3;
  final parameter Integer q = p;
  Real x[q] = ones(q);
end M;
```

olivleno:
# Web Meeting: Hans, Henrik, Gerd, Martin, Oliver

Henrik: It's not just about filtering the results. It is also about preserving the information that the variability has changed during transformation to flat Modelica.

Hans: There are various ways how we could address it. Until we know more precisely what it will be used for we should not make any change.

Gerd: We need this information already today.

Hans: How much is it used?

Gerd: If you have a binary and need to know whether you have to re-compile or not then this is crucial.

Hans: If it's a constant, then you cannot change it. 

Henrik: I expect that most tools will come up with a vendor specific solution to handle this case and if that's the case the we should have something standardized.

Gerd: Another information is if the original model allowed this parameter to be changed by the user.

Hans: We not should take the risk of having to life with a non-optimal solution.

Henrik: The example above illustrates that in this case the information of a changed variability is not sufficient. Therefore I agree with Hans that we should first find vendor specific annotations and come up with a solution later.

Options:
1. Decide this concern after first prototype implementations exist.
2. Propose an annotation to the best of our knowledge now and revise it later if needed.

Poll
Option1:  Henrik, Hans, Gerd
Option2: none
abstain: Oliver, Martin

Decision:
Decide later

Conclusion:
Close without merging.
Move road map item to next revision.

PR #3163, opened 2022-05-03
Investigate need for for-equations

henrikt-ma:
When initiating this PR, the change of the PR means removal of `for`-equations.  This should only be taken as a starting point for the discussion, not as an actual proposal to go ahead with this change.

### DIFF
--- a/RationaleMCP/0031/grammar.md
+++ b/RationaleMCP/0031/grammar.md
@@ -302,7 +302,7 @@ end _F;
 
 > _for-indices_ → _for-index_ ( **,** _for-index_ )*
 
-> _for-index_ → _IDENT_ ( **in** _expression_ )?
+> _for-index_ → _IDENT_ **in** _expression_
 
 > _while-statement_ →\
 > &emsp; **while** _expression_ **loop**\
--- a/RationaleMCP/0031/grammar.md
+++ b/RationaleMCP/0031/grammar.md
@@ -248,7 +248,7 @@ end _F;
 > _equation_ →\
 > &emsp; ( _simple-expression_ ( **=** _expression_ )?\
 > &emsp; | _if-equation_\
-> &emsp; ~~| _for-equation_~~\
+> &emsp; | _for-equation_\
 > &emsp; ~~| _connect-clause_~~\
 > &emsp; | _when-equation_\
 > &emsp; )\
@@ -290,10 +290,10 @@ end _F;
 > &emsp; )?\
 > &emsp; **end** **if**
 
-> ~~_for-equation_ →~~\
-> ~~&emsp; **for** _for-indices_ **loop**~~\
-> ~~&emsp;&emsp; ( _equation_ **;** )* ~~\
-> ~~&emsp; **end** **for**~~
+> _for-equation_ →\
+> &emsp; **for** _for-indices_ **loop**\
+> &emsp;&emsp; ( _equation_ **;** )* \
+> &emsp; **end** **for**
 
 > _for-statement_ →\
 > &emsp; **for** _for-indices_ **loop**\
--- a/RationaleMCP/0031/grammar.md
+++ b/RationaleMCP/0031/grammar.md
@@ -248,7 +248,7 @@ end _F;
 > _equation_ →\
 > &emsp; ( _simple-expression_ ( **=** _expression_ )?\
 > &emsp; | _if-equation_\
-> &emsp; | _for-equation_\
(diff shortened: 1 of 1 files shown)


henrikt-ma:
My initial reaction when setting up this PR is that it seems a bit drastic to remove the `for`-equations as long as we still have the `for`-statements.  I mean, could the `for`-equations really be that harmful to keep?

perost:
In OpenModelica we sometimes keep for-equations in our flat format, to avoid unrolling loops unnecessarily and sometimes even generating for-equations when doing automatic vectorization. Not having for-equations would probably make Flat Modelica unsuitable for at least some of our applications, so keeping them would be nice unless there are good reasons not to.

casella:
> In OpenModelica we sometimes keep for-equations in our flat format, to avoid unrolling loops unnecessarily and sometimes even generating for-equations when doing automatic vectorization. Not having for-equations would probably make Flat Modelica unsuitable for at least some of our applications, so keeping them would be nice unless there are good reasons not to.

I also had a similar comment. Eventually, this could also be interesting for Flat modelica, since avoiding to unroll arrays and for loops is essential for high-performance and compact code generation when (large) arrays are involved.

BTW, the description of the flattening process in [Section 5.6.2](https://specification.modelica.org/maint/3.5/scoping-name-lookup-and-flattening.html#generation-of-the-flat-equation-system) states: 
>  The flat equation system consists of a list of variables _with dimensions_, flattened equations and algorithms, and a list of called functions which are flattened separately.

I understand "variables with dimensions" means scalars, but also (multi) dimensional arrays. In that case, it would be odd to allow having arrays in the variable definitions, but not keeping the corresponding for-equations in the equations set.

olivleno:

A `for` statement can be transformed into a `while` statement.

Gerd: It's always better for optimization to keep the `for` statement. Hence `for` statements should be allowed.


Reasons to keep `for` equation 
* Need to support for statements.
* Commonly used together with array variables.
* Unrolling can be bad for optimization.
* Unrolling can slow down symbolic processing.

Reason to remove `for`equations
* Smaller equation language.

Get rid of implicit iterator ranges of short hand:
```
for i loop
  x[i]=y[i]+1;
end for; 
```

```
for i in 1:size(x,1) loop
  x[i]=y[i]+1;
end for; 
```

Requires non-trivial processing and may lead to different error messages.
I order to keep flat Modelica simple this would be beneficial to being removed.

Poll on `for` statements and equations:
Option 1: keep both
Option 2: keep just statements

Option 1: Henrik, Hans, Gerd, Oliver 
abstain: Martin

Decision: Option 1

Poll on implicit iterator ranges
Option 1: keep
Option 2: remove

Option 1: none
Option 2: Henrik, Hans, Gerd, Oliver
abstain: Martin

Decision: Option 2

Update PR [Henrik]

olivleno:

Full Modelica allows:
```
// for equation
for i in 1:5, j in 1:10 loop
  x[i,j] = a[i]+b[j];
end for;

// for statement
for i in 1:5, j in 1:10 loop
  x := x * (a[i]+b[j]);
end for;

//reduction expression
product(a[i]+b[j] for i in 1:5, j in 1:10)

//construct multi-dimensional arrays
{a[i]+b[j] for i in 1:5, j in 1:10}
```

This can be considered as syntactic sugar for:
```
// for equation
for 'i' in 1:5 loop 
  for 'j' in 1:10 loop
    'x'['i','j'] = 'a'['i']+'b'['j'];
  end for;
end for;

// for statement
for 'i' in 1:5 loop 
  for 'j' in 1:10 loop
    'x' := 'x' * ('a'['i']+b['j']);
  end for;
end for;

//reduction expression
product(product(a[i]+b[j] for i in 1:5) for  j in 1:10)

//construct multi-dimensional arrays
{{a[i]+b[j] for i in 1:5} j in 1:10}
```

Hans: This is a typical example of syntactic sugar that can be removed.

Henrik: If needed the original form can easily be reconstructed. No information is lost.
original
```
sum(j for j in 1:i, i in 1:4)
```
expanded
```
sum(sum(j for j in 1:i) for i in 1:4)
```
Whenever there is a `sum(sum(...))` one can rewrite it in the compact form applying the rewrite-rule of the Modelica spec. in reverse.

Option 1: Don't support the compact form and always expand it in flat Modelica.
Option 2: Support also the compact form in flat Modelica.

Poll:
Option 1: Hans, Henrik, Gerd, Martin
abstain: Oliver

Decision
Accept Option 1.
Merge

PR #3175, opened 2022-05-11
New builtin function for testing exact equality of Real parameter expressions

henrikt-ma:


henrikt-ma:
An interesting alternative design would be to allow `==` and `<>` on `Real` outside functions as long as the operands are parameter expressions.

olivleno:

Option 1: New function realParameterEqual (Hans, Henrik, Gerd, Martin)
Option 2: Allow operator on real parameter expression (Henrik)
Option 3: Both (Henrik)
Abstain: (Oliver)

Discussion:
Hans: Pro for function is that it allows to have an extra parameter to hold the tolerance.

Henrik: Con for function is that it looks like a function but it has a special semantics related to parameter expression, which is something different than a Modelica function.

Hans: Function stands out more clearly showing that this a special case.

Gerd: Why restrict it to parameter expression if we consider the tolerance?

Hans: We can have an argument with the builtin tolerance as default value.

Henrik: The function does not have to be restricted, but we don't want to encourage the user to make use of it elsewhere.

Hans: Name referring to parameter is good to avoid broader use, but could be conflicting in the future.

Gerd: In the past I sometimes needed such a function.

Henrik: The more general case should be handled by another function in full Modelica.

Henrik: realParameterEqual is for sure not an event generating function, which is obvious for parameter expressions and simplifies the design. Considering time varying variables would be much more involved with that respect and should be a different function.,

Decision:
Go forward with Option 1.
Apply the tolerance that the parameters are stored with.

Example:
```
parameter Real p = pi/2;
realParameterEqual(cos(p),0)
```
Will most likely evaluate to false

Conclusions:
* Keep the design as is. Only if needed a tolerance may be added later.
* Stick with the proposed function name.
* Approve and merge PR [Henrik]

PR #3177, opened 2022-05-13
Use bibliography for gettext reference

henrikt-ma:
The discussion about whether a bibliography is good or not seems out of the PR scope, so we'd better stick to the current style of using it.


gkurzbach (approved):
Yes it's better to have it formally correct.

PR #3183, opened 2022-05-15
Undo accidentally removed indentation

henrikt-ma:
Undoes what looks like a mistake in #3181.

As I see it, this line belongs to the line above, and should therefore have more indentation than the interior of the model definition.  @tobolar, please indicate if it was intentional to have the description string indented the same as the interior of the model definition.


HansOlsson (approved):
Ok, but to me it would be simpler to just have:
`partial model BaseProperties "Interface of medium model"`
The extra text isn't that important, and not worth a line-break.

PR #3185, opened 2022-05-15
Avoid bad leading white-space inside strings in StateSelect definition

henrikt-ma:
Further improving a listing changed in #3181.

The reason to fix this is not that it really matters for the `StateSelect` definition itself, but that we should avoid showing listings that could give the wrong impression of how to handle long strings in actual code.


HansOlsson (approved):
Ok.

PR #3221, opened 2022-08-16
Fix hard-coded cross references related to function application

henrikt-ma:


henrikt-ma:
(I have verified that the cross references from the listings work also in the LaTeXML build.)

HansOlsson (approved):
Ok.

PR #3295, opened 2022-11-28
Source locations pointing back to the original Modelica code

henrikt-ma:
Initiating PR in _Draft_ state.


RationaleMCP/0031/source-locations.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Design goals
+
+Some design goals to keep in mind when comparing the design alternatives:
+- The detailed format of a source location needs to be flexible to serve the needs of different tools.
+- A tool should be able to ignore source locations originating from other tools.
+- Source locations may be needed on many levels of the Flat Modelica code, including component declarations, equations, modifications, and expressions.
+- Reduced Flat Modelica soruce code human readability is a concern when soruce locations are added.
+- Simple compression of the source locations could be handy.
+
```
olivleno, line 14: Web Meeting (Hans, Henrik, Gerd, Martin, Oliver)

1. What kind of information has to be captured?

- Reference to the source file (filename, line, column) or instance tree (fully qualified name)
- Could be anything, not to be decided now for prototyping.
- There must be the option to express vendor specific stuff, e.g. transformations.

2. Where should the information be stored?

- External file, referenced by an annotation on model level.
```
package 'mymodel'
model 'mymodel'
...
end 'mymodel';
annotation(SourceLocations(...));
end 'mymodel';
```
- inline, special kind of annotation 

4. How to attach the information in the code?
5. How to express the associated information?

olivleno:

Poll on `SourceLocations` annotation for tool-specific table:

Proposal: see: https://github.com/modelica/ModelicaSpecification/blob/MCP/0031%2Bsource-locations/RationaleMCP/0031/source-locations.md#tool-specific-table-of-source-location-details 

in favor: Henrik, Martin, Gerd, Hans, Oliver

Decision: Accept this subsection.

RationaleMCP/0031/source-locations.md
```diff
@@ -0,0 +1,123 @@
...
+- Number of supported formats:
+  - Standardize on just one of the alternatives.
+  - Allow more than one of the alternatives.
+- Number of informations included in a single source location decoration
+  - Just a single piece of information.
+  - Multiple pieces of information (possibly in different formats).
+
+### Conclusion
+
+None yet, but all discussion so far has pointed in the direction of just a single integer.
```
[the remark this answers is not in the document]
henrikt-ma: According to today's decision above:
```suggestion
For now, only allow a single integer.  It would be easy to extend this in future versions of Flat Modelica.
```

RationaleMCP/0031/source-locations.md
```diff
@@ -0,0 +1,123 @@
...
+  Real 'y' = @23 'x';           // Attach 23 to component declaration
+equation
+  der('x') = @24 1;             // Attach 24 to equation
+  der('y') = 1 @25;             // Attach 25 to "1"
+end 'MyModel';
+annotation(SourceLocations(…));
+end 'MyModel';
+```
+
+The precedence of `@` is lower than the expression operators, so that no parentheses are needed when attaching a source locaiton to a modification or binding.
```
henrikt-ma, line 97: Typo:
```suggestion
The precedence of `@` is lower than the expression operators, so that no parentheses are needed when attaching a source location to a modification or binding.
```

RationaleMCP/0031/source-locations.md
```diff
@@ -0,0 +1,123 @@
...
+
+Use infix `@` as decoration attachment operator:
+```
+package 'MyModel'
+model 'MyModel'
+  parameter Real 'p' = 42 @21;  // Attach 21 to binding expression "42"
+  Real 'x'(min = 0 @22);        // Attach 22 to modification "0"
+  Real 'y' = @23 'x';           // Attach 23 to component declaration
+equation
+  der('x') = @24 1;             // Attach 24 to equation
```
olivleno, line 90: `@24 der('x') = 1;`

RationaleMCP/0031/source-locations.md
```diff
@@ -0,0 +1,123 @@
...
+
+#### Low precedence infix `@` operator
+
+Use infix `@` as decoration attachment operator:
+```
+package 'MyModel'
+model 'MyModel'
+  parameter Real 'p' = 42 @21;  // Attach 21 to binding expression "42"
+  Real 'x'(min = 0 @22);        // Attach 22 to modification "0"
+  Real 'y' = @23 'x';           // Attach 23 to component declaration
```
olivleno, line 88: `@23 Real 'y' = 'x'`

RationaleMCP/0031/source-locations.md
```diff
@@ -0,0 +1,123 @@
...
+### Design alternatives
+
+#### Low precedence infix `@` operator
+
+Use infix `@` as decoration attachment operator:
+```
+package 'MyModel'
+model 'MyModel'
+  parameter Real 'p' = 42 @21;  // Attach 21 to binding expression "42"
+  Real 'x'(min = 0 @22);        // Attach 22 to modification "0"
```
olivleno, line 87: `Real 'x'(@22 min = 0);`

olivleno: Web Meeting: Hans, Henrik, Gerd, Martin, Oliver

There is preference for the alternative syntax at the beginning of a construct.

Currently there is nothing said about models, records and functions.

- The natural way to do it would be in front of `model`, `record`, `function`.

Except of expressions a prefix can be used, unless using parenthesis, but this would require additions in many places.
Relying on operator precedence we could avoid this.

```
a * b * (c + d) @30     // refers to the entire expression
a * b * (c + d @30)     // refers to only the subexpression `c+d`
a * b * (c + (d @30))  // refers to only the subexpression `d`
```

We could also allow a prefix with parenthesis:
```
(@30 a * b * (c + d))     // refers to the entire expression
a * b * (@30 c + d)     // refers to only the subexpression `c+d`
a * b * (c + (@30 d))  // refers to only the subexpression `d`
```

Both should work.
The postfix avoiding parenthesis.
The prefix being more consistent with previous cases.
Both are possible to be combined.

To conclude it would make sense to run some tests with a parser.
Should be easy to enhance the parser to consider @ signs to add an integer. [Henrik]

To be concluded in next meeting based on the insights gained from the tests.

henrikt-ma: I have now carried out the exercise of allowing these decorations in a modified full Modelica parser.

Here's the test package I'm using:
```
@101 package Decorations
  @102 model M
    Real x = 1 + 2 @103;
    Real y = (@104 3 + 4);
    @105 Real z;
    Real w;
  equation
    @106 6 + 7 @107 = atan2(8 + 9 * 10 @108, 500);
  algorithm
    @109 w := x + y;
  end M;
end Decorations;
```
For this package I get the following output as proof of concept:
```
Expression decoration 103: 1 + 2
Prefix expression decoration 104.
Component decoration 105.
Expression decoration 108: 8 + 9 * 10
Expression decoration (in eq LHS) 107
Equation decoration 106.
Statement decoration 109.
Class definition decoration 102.
Stored definition class definition decoration 101.
```

The decoration 101 on the outermost package is included just to highlight that it is different from class definitions further down in the syntax tree, but I think this is a place where it isn't needed to support a source location decoration in Flat Modelica.

Here is an example of something that is a syntax error in my test implementation:
```
    Real x = 1 @103 + 2;
```
I believe it wouldn't be difficult to extend the grammar to also allow this, but I don't think that the added value could motivate the increased complexity of a grammar that systematically would allow decorations after any expression in the expression tree where ambiguity can be avoided.  Future versions of Flat Modelica could always add such possibilities if one would regret a decision to not allow it now.

Let me know if you think it's too early to update this PR with changes to the Flat Modelica grammar.  I'll give it a couple of days, and unless someone tells me to wait, I'll then update the PR so that we can look at the grammar changes next meeting.

henrikt-ma: Silence, so I'm proceeding with updating the grammar in the PR.

henrikt-ma: In the end, there turned out to be one thing that presented too much resistance for me to add it to my experimental parser, namely prefix decorations for function arguments:
```
atan2(@1 8 + 9 * 10 @108, @2 z);
```

Right now, I can't explain what the problem comes down to.  It might be related to the way our grammar is structured, os possibly something related to how the Modelica grammar is given in the specification, I don't know.  Either way, even if I could make it work in the experimental parser, I'm afraid that I could easily mess it up when transferring the changes in my experimental grammar to our Flat Modelica grammar.  Next meeting we should discuss how important it is to support the prefix form for function arguments.

olivleno: Web Meeting (Hans, Henrik, Martin, Oliver, Gerd)

Henrik's experience from his test implementation above:

- postfix form was straight forward
- prefix form for non-expressions also straight forward: class declaration, component declaration, modification, equation
- prefix form for function arguments are not easy to do, not implemented as of now.
- prefix form would need some changes in the grammar.

Options for supporting prefix form for expressions:

1. Restructure the grammar to support prefix form in all places.
2. Restrict the prefix form, e.g. require parenthesis always, e.g. consider the support of prefix for function arguments in later versions.
3. No prefix form for expressions in general, at least in version 1.

Pros and cons
To 1.) Might fail or lead to acceptance issues if too involved. 
To 2.) Not supporting all cases may lead to confusion.

Discussion:
Gerd: Supporting both could be problematic for expressions. Hence, if prefix does not work in all cases for expressions, then we should stick to the postfix form.

Hans: Currently it's not clear what is the issue with the prefix form. Therefore we should not rely on it.

Martin: Avoiding ambiguities as they occur through the prefix form will make reading and parsing easier. Hence, we should either avoid the prefix form or have parenthesis.

Martin: postfix is ambiguous also.

Poll on the option 1-3 above:
Option 1: -
Option 2: -
Option 3: Henrik, Gerd, Oliver, Hans, Martin

Next steps:
Review the changes made to the grammar.

RationaleMCP/0031/grammar.md
```diff
@@ -66,10 +66,11 @@ The _S-CHAR_ accepts Unicode other than " and \\:
...
 > &emsp; _VERSION-HEADER_\
 > &emsp; **package** _IDENT_\
-> &emsp;&emsp; ( _class-definition_ **;**\
-> &emsp;&emsp; | _global-constant_ **;**\
+> &emsp;&emsp; ( _decoration_? _class-definition_ **;**\
+> &emsp;&emsp; | _decoration_? _global-constant_ **;**\
 > &emsp;&emsp; )*\
-> &emsp;&emsp; **model** _long-class-specifier_ **;**\
+> &emsp;&emsp; _decoration_? **model** _long-class-specifier_ **;**\
+> &emsp;&emsp; ( _annotation-comment_ **;** )?
```
olivleno, line 73: Web Meeting (Hans, Gerd, Martin, Oliver)
unsure about this added line

olivleno: Not related to the decoration but to SourceLocations annotation for tool-specific table, see decision Dec 14. 2022.

RationaleMCP/0031/grammar.md
```diff
@@ -265,7 +266,8 @@ end _F;
 ## B26 Equations
 
 > _equation_ →\
-> &emsp; ( _simple-expression_ ( **=** _expression_ )?\
+> &emsp; _decoration_?\
+> &emsp; ( _simple-expression_ _decoration_? ( **=** _expression_ )?\
```
olivleno, line 270: Web Meeting (Hans, Henrik, Gerd, Martin, Oliver)
No issue with having no parentheses here, as they are only related to the dropped prefix option.

RationaleMCP/0031/grammar.md
```diff
@@ -355,12 +358,16 @@ end _F;
...
+
+> _expression_ → _expression-no-decoration_ _decoration_?
+
+> _expression-no-decoration_ → _simple-expression_ | _if-expression_
 
 > _if-expression_ →\
 > &emsp; **if** _expression_ **then** _expression_\
 > &emsp; ( **elseif** _expression_ **then** _expression_ )* \
-> &emsp; **else** _expression_
+> &emsp; **else** _expression-no-decoration_
```
olivleno, line 370: Web Meeting (Hans, Henrik, Gerd, Martin, Oliver)
A bit unfortunate to not having decorations on the else branch.
Using parenthesis could make it clear that the decoration belongs to the else branch.
```
Real x = if b then 1 @41 else 2 @42; // common use of decoration @42 belonging to entire if expr
```
```
Real x = if b then (1 @41) else (2 @42); // decoration @42 belonging to else branch, adding parenthesis also for @41
```
It's possible to omit the parenthesis in the other branches. 

Solution options:
A) as proposed, require as few parenthesis as possible.
B) require parenthesis for each branch and branching condition always. 1st example would be illegal.

Poll
A: 
B: Henrik, Gerd, Oliver
abstain: Hans, Martin

Decision:
Go for B.

PR #3303, opened 2022-12-08
Missing type specifier in example

eshmoylova:
Missing type specifier in the example for Removing modifiers.

Trying out if I can create a pull request from the source without a local copy.

### DIFF
--- a/chapters/inheritance.tex
+++ b/chapters/inheritance.tex
@@ -1271,3 +1271,4 @@ end C;
 In this case the deselection removes all of the connect-statements.
 \end{example}
 \end{nonnormative}
+
--- a/chapters/inheritance.tex
+++ b/chapters/inheritance.tex
@@ -1271,4 +1271,3 @@ end C;
 In this case the deselection removes all of the connect-statements.
 \end{example}
 \end{nonnormative}
-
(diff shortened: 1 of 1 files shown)


chapters/inheritance.tex
```diff
@@ -1270,4 +1270,4 @@ \section{Selective Model Extension}\label{selective-model-extension}
 \end{lstlisting}
 In this case the deselection removes all of the connect-statements.
 \end{example}
-\end{nonnormative}
\ No newline at end of file
+\end{nonnormative}
```
eshmoylova, line 1273: I used "Edit this file" on github to modify it, and it also modified the white space at the end of the file. Not sure if this is crucial or how to revert that.

henrikt-ma: It should be there.  Let's try this:
```suggestion
\end{nonnormative}

```

henrikt-ma (changes requested):
Please commit the suggestion to restore the newline at end of file.

henrikt-ma (approved):
After correcting my bad attempt at cleaning up the changes, this is now back to where it started, and it is good.

HansOlsson (approved):
Ok.

PR #3304, opened 2022-12-08
connect-clause was renamed to be connect-equation

eshmoylova:
Matching other definitions in the chapter


henrikt-ma (approved):
Good catch!

HansOlsson (approved):
Ah, yes, good that we good to get that corrected - it was introduced by the selective model extension.

PR #3317, opened 2022-12-13
Implicit iterator examples

henrikt-ma:
When reviewing #3306 I was struck by the bad quality of the example in that section.  This is a cleanup that both makes the section look better, and also tries to motivate the existence of the second part of the old example.


HansOlsson (approved):
Ok, seems cleaner.

PR #3319, opened 2022-12-15
Add globs to cleaning targets

henrikt-ma:
With this, I have a clean `git clean -nx` after running the cleaning targets.


HansOlsson (approved):
Ok, does not seem like any dangerous files.

PR #3323, opened 2022-12-18
Only load cmap.sty when building PDF

### LINKED ISSUE #3302: Broken references
It looks like generation of references is broken. For example, in [State Machines](https://github.com/modelica/ModelicaSpecification/blob/d609f556d7a28bdd016c15fc551da112dcc0f160/chapters/statemachines.tex#L4) there is a reference:
```
\parencite{Pouzet2006LucidSynchrone30}
```
It appears as plain text in [the generated html](https://specification.modelica.org/master/state-machines.html):
image

Similarly in [Selective Model Extension](https://github.com/modelica/ModelicaSpecification/blob/d609f556d7a28bdd016c15fc551da112dcc0f160/chapters/inheritance.tex#L1148):
```
\textcite{Buerger2019SelectiveModel}
```
which  appears [as plain text](https://specification.modelica.org/master/inheritance-modification-and-redeclaration.html#selective-model-extension).

If I click on References in the menu, it is empty:
image

This is on Chrome.

In pdf, references are generate. For both cases there are hyperlinks. But the appearance is different "(Harel 1987)" vs "Bürger (2019)". In both cases only the year is a hyperlink.

henrikt-ma:
Fixes #3302.

### DIFF
--- a/MLS.tex
+++ b/MLS.tex
@@ -1,6 +1,12 @@
 \documentclass[10pt,a4paper]{report}
-\usepackage{cmap}
 \NeedsTeXFormat{LaTeX2e}
+\usepackage{ifpdf}
+\ifpdf
+% While the cmap.sty README says that the package should be included "at the beginning of your preamble",
+% we need to at least load it after ifpdf due to incompatibility with LaTeXML, see:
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Seems ok.

PR #3325, opened 2022-12-18
Clarify relation between predefined type names and keywords

### LINKED ISSUE #2813: Clarify in what ways names like Real are different from keywords
Consider the following model:
```
model EnumeratorsAreNotElements
  type Types = enumeration(Real, Integer, Boolean);
  Types x = Types.Real;
end EnumeratorsAreNotElements;
```

My reading of the specification suggests that this is a valid model:
> The names `Real`, `Integer`, `Boolean` and `String` are reserved such that it is illegal to declare an element with these names.

Hence, I think the example could be added as an example that clarifies in what ways these names are different from keywords.  Maybe there are other examples worth mentioning as well?

(Ideally, I would have liked these names to be keywords, so that we didn't have to add this kind of clarifications at all.)

henrikt-ma:
According to phone meeting decision: https://github.com/modelica/ModelicaSpecification/issues/2813#issuecomment-1351604814

Fixes #2813.

### DIFF
--- a/chapters/classes.tex
+++ b/chapters/classes.tex
@@ -1390,7 +1390,7 @@ Note that only the values of \lstinline!x! and \lstinline!y! are declared to be 
 
 It is not possible to combine extends from the predefined types, enumeration types, or this \lstinline!Clock! type with other components.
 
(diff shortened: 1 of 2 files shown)


HansOlsson (approved):
Ok, even though the StateSelect example seemed a bit unnecessary.

PR #3346, opened 2023-03-08
Minor: delete a verb making nonsense sentence

tobolar:


henrikt-ma (approved):
Looks good.

HansOlsson (approved):
Good change. It seems it was intended to be two sentences - one saying that they were listed below, and one that stating that they do not generate events.

PR #3348, opened 2023-03-14
Remove stray braces in FORTRAN 77 type mapping table

henrikt-ma:


beutlich:
Should it also be back-ported to maint/3.6?

HansOlsson (approved):
ok!

PR #3352, opened 2023-03-15
Clarify that unit/displayUnit-attributes can be empty

### LINKED ISSUE #3345: Explicitly setting unit = ""
According to the [unit string syntax](https://specification.modelica.org/master/unit-expressions.html#the-syntax-of-unit-expressions), the following modifications are illegal:
```
Real x(unit = "", displayUnit = "");
```

However, this is precisely how the defaults for the `unit` and `displayUnit` of `Real` are given:
```
  parameter StringType unit        = "" "Unit used in equations";
  parameter StringType displayUnit = "" "Default display unit";
```

This requires some kind of resolution.

With the advent of `break`-modifications, one could argue that they would be the canonical way of "removing" a `unit` or `displayUnit` without violating the unit string syntax.  However, one should keep in mind that there are things that cannot be expressed using a `break`-modification:
```
constant Boolean useUnit = false;
Real x(unit = if useUnit then "m" else "");
```

henrikt-ma:
Fixes #3345.

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -163,8 +163,9 @@ end Axis;
 
 When an axis bound is not provided, the tool computes one automatically.
 
-An empty \lstinline!unit! means that the axis is unitless, and each expression plotted against it may use its own unit determined by the tool.  The tool is responsible for conveying the information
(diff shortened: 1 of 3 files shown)


HansOlsson (approved):
Looks good.

PR #3365, opened 2023-05-03
Replace stray 'c' in definition of clocked sample

henrikt-ma:
This looks like a leftover from an earlier presentation where the clock was denoted `c`.

Discovered when preparing #3367.


HansOlsson (approved):
Seems ok

PR #3368, opened 2023-05-03
Sentence-based line breaks in synchronous.tex

henrikt-ma:
Completing the transition to sentence-based line breaks in this chapter.

Triggered by the need to deal with line breaks in #3366, and serving as a preparation for working on #3367 without the disturbance of line breaking.


HansOlsson (approved):
Looks good. I especially checked the captions.

PR #3377, opened 2023-05-15
Don't use \emph for content of lesser importance

henrikt-ma:
This is a small follow-up to #3372, addressing [comment](https://github.com/modelica/ModelicaSpecification/pull/3372#discussion_r1194351260).

It it was considered important, it shouldn't have been put in a parenthesis in the first place.

Also putting \emph where it should be used to quote words in the listing.


HansOlsson (approved):
Looks good

PR #3384, opened 2023-06-11
Fix missing styling in within-clause example

henrikt-ma:


HansOlsson (approved):
Looks consistent with previous line where packageprefixname is used.

PR #3400, opened 2023-08-08
Fix typo

beutlich:


HansOlsson (approved):
Looks good.
It was introduced during work on #2719

PR #3407, opened 2023-08-23
Add comment about requested LaTeXML feature being marked as fixed

henrikt-ma:


HansOlsson (approved):
Looks good as note - will see if it looks good in the final version as well before activating it (later).

PR #3430, opened 2023-11-06
Clarify complicated sentence about 'input' prefix exception in functions

henrikt-ma:
The previous formulation didn't make sense grammatically to me, as I expected another condition for the exception to apply following the _and_.  This PR fixes this along with writing out a few more things that make the rule easier to parse.


perost:
I think `special class` should be changed to `specialized class` in the first sentence too while you're at it.

henrikt-ma:
> I think `special class` should be changed to `specialized class` in the first sentence too while you're at it.

Done.

HansOlsson (approved):
Ok, makes sense.

PR #3432, opened 2023-11-06
Add 'short class definition' to document index

henrikt-ma:
Addressing a complaint that it was not possible to find _short class definition_ in the document index.


HansOlsson (approved):
Ok.

PR #3434, opened 2023-11-07
External call constant arguments and literals

### LINKED ISSUE #3427: Is 1 + 1 a scalar constant?
Is this a valid call to the external function `myExtFun`?
```
model OnePlusOne
  function f
    output Integer y;
  external "C"
    y = myExtFun(1 + 1);
      annotation(Include = "
int
myExtFun(int x)
{
  return x;
}
");
  end f;
  
  Integer k = f();
end OnePlusOne;
```

The specification says that _scalar constants_ are allowed as arguments, but I'd like clarification on whether this is supposed to only allow literal values rather than more general constant scalar expressions.

henrikt-ma:
Fixes #3427.


HansOlsson (approved):
Ok.

Issue #1668, opened 2018-11-04
BuildProjects to automatically compile Modelica binaries


modelica-trac-importer:
**Reported by sjoelund&#x2e;se on 26 Feb 2015 12:41 UTC**
Since some time we have had a directory Modelica/Resources/BuildProjects to make it easier to compile the binary libraries used by the package. This makes maintenance easier since you do not need to compile and keep svn/git updated with new versions of the library all the time (since the user/tool provider can easily compile the source code).

I propose that this is specified and made automatic by Modelica tools (that support external C). I even made a prototype implementation in OpenModelica.

The following uses the library name ModelicaExternalC2 instead of ModelicaExternalC because OpenModelica itself has ModelicaExternalC pre-compiled:
```mo
model M
function f
   input Real time;
   output Real r;
external "C" annotation(Library="ModelicaExternalC2");
end f;

   Real x = f(time);
end M;
```

The prototype works by looking in the usual library directories for a file libmylib.a or libmylib.so (as well as OPENMODELICAHOME/lib/omc OPENMODELICAHOME/lib and ~/.openmodelica/binaries/PACKAGENAME). If no library is found, it tries to look for modelica://M/Resources/BuildProjects/\*/autogen.sh. It then tries to run that project using the command below.

The generated library for ModelicaTables is linked against hdf5 if the system has hdf5 installed. If it is not installed, it asks the OS if the hdf5 is available in the package manager, and if it is it asks for admin rights to install it.

![Image](hdf5-install.png)

```
[1] 13:32:08 Translation Notification
[/home/marsj/tmp/a.mo: 2:1-6:6]: Removed directory /tmp/omc_compile_ModelicaExternalC2_7UJy2p

[2] 13:32:08 Translation Notification
[/home/marsj/tmp/a.mo: 2:1-6:6]: Compiled /home/marsj/.openmodelica/binaries/M/libModelicaExternalC2.so by running build project /home/marsj/tmp/Resources/BuildProjects/autotools

[3] 13:32:08 Translation Notification
[/home/marsj/tmp/a.mo: 2:1-6:6]: Succeeded with compilation and installation of the library using: command: sh ./autogen.sh && ./configure --libdir='/home/marsj/.openmodelica/binaries/M' && make && make install
[...]
[...]

PR #3455, opened 2023-12-07
Describe common ways of using ModelicaUtilities.h

henrikt-ma:
This was factored out of #3452, as promised in https://github.com/modelica/ModelicaSpecification/pull/3452#discussion_r1419690390.

In addition to examples and other non-normative text, this PR makes clear that `#include "ModelicaUtilities.h"` shall work inside an `Include` annotation.

### DIFF
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -2660,7 +2660,8 @@ The following usage patterns are common:
 \end{example}
 
 \begin{nonnormative}
-External libraries can built with the \filename{ModelicaUtilities.h} included within the Modelica Standard Library.
+When building external libraries, it is not possible to rely on the tool mechanism that provides \filename{ModelicaUtilities.h} for \lstinline!Include! annotations.
+Instead, an external library project may contain its own instance of \filename{ModelicaUtilities.h}, for example obtained by making a copy of the instance included within the Modelica Standard Library.
 \end{nonnormative}
 
 
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -2660,9 +2660,7 @@ The following usage patterns are common:
 \end{example}
 
 \begin{nonnormative}
-There is no standardized way to build external libraries using the functions in \filename{ModelicaUtilities.h}.
-For example, there is no standardized way of obtaining a directory where the tool shall provide the \filename{ModelicaUtilities.h}.
-There is also no standardized library to link with to get a tool-specific implementation of the functions.
+External libraries can built with the \filename{ModelicaUtilities.h} included within the Modelica Standard Library.
 \end{nonnormative}
 
 
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -2660,7 +2660,7 @@ The following usage patterns are common:
 \end{example}
 
 \begin{nonnormative}
-When building external libraries, it is not possible to rely on the tool mechanism that provides \filename{ModelicaUtilities.h} for \lstinline!Include! annotations.
+When building external libraries independently of Modelica tools, it is not possible to rely on the tool mechanism that provides \filename{ModelicaUtilities.h} for \lstinline!Include! annotations.
 Instead, an external library project may contain its own instance of \filename{ModelicaUtilities.h}, for example obtained by making a copy of the instance included within the Modelica Standard Library.
 \end{nonnormative}
 
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -2645,6 +2645,27 @@ z1 = myfoo(2.4, 3, &i2);
 
 This section describes the utility functions declared in \filename{ModelicaUtilities.h}, which can be called in external Modelica functions written in C.
 
+The tool must ensure that the header is found by \lstinline[language=C]!#include "ModelicaUtilities.h"! within an \lstinline!Include! annotation (see \cref{annotations-for-external-libraries-and-include-files}); no \lstinline!IncludeDirectory! annotation is needed.
+
+\begin{example}
+The following usage patterns are common:
+\begin{itemize}
+\item
+  The \lstinline!Include! annotation may first \lstinline[language=C]!#include "ModelicaUtilities.h"!, and then contain the external function definition where the utility functions may be used.
+\item
+  Like above, but moving the annotation content above to, say, \filename{myExtFun.c}, and then just do \lstinline[language=C]!#include "myExtFun.c"! in the annotation.
+\item
+  After \lstinline[language=C]!#include "ModelicaUtilities.h"!, the \lstinline!Include! annotation can also define a wrapper around a function in a linked library that does the real job, where the wrapper forwards the arguments as well as passes function pointers for a selection of functions from \filename{ModelicaUtilities.h}.
+\end{itemize}
+\end{example}
+
+\begin{nonnormative}
+There is no standardized way to build external libraries using the functions in \filename{ModelicaUtilities.h}.
+For example, there is no standardized way of obtaining a directory where the tool shall provide the \filename{ModelicaUtilities.h}.
+There is also no standardized library to link with to get a tool-specific implementation of the functions.
+\end{nonnormative}
+
+
 \subsubsection{Error Reporting Utility Functions}\label{utility-functions-for-reporting-errors}\label{error-reporting-utility-functions}
 
 The functions listed below produce a message in different ways.


chapters/functions.tex
```diff
@@ -2645,6 +2645,27 @@ \subsection{Utility Functions}\label{utility-functions}
...
+  The \lstinline!Include! annotation may first \lstinline[language=C]!#include "ModelicaUtilities.h"!, and then contain the external function definition where the utility functions may be used.
+\item
+  Like above, but moving the annotation content above to, say, \filename{myExtFun.c}, and then just do \lstinline[language=C]!#include "myExtFun.c"! in the annotation.
+\item
+  After \lstinline[language=C]!#include "ModelicaUtilities.h"!, the \lstinline!Include! annotation can also define a wrapper around a function in a linked library that does the real job, where the wrapper forwards the arguments as well as passes function pointers for a selection of functions from \filename{ModelicaUtilities.h}.
+\end{itemize}
+\end{example}
+
+\begin{nonnormative}
+There is no standardized way to build external libraries using the functions in \filename{ModelicaUtilities.h}.
```
HansOlsson, line 2663: The term "standardized" is a bit ambiguous, would it be possible to be clearer? 

As a library developer ideally one would want a tool-independent way of building external libraries using them; having a standardized tool-dependent way would be messy, and we have neither (that's why wrappers are needed even if building libraries for Modelica - but on the other hand making the library independent of Modelica has benefits).

HansOlsson: Hmm... Having tested a few things within ModelicaStandardLibrary I'm not sure how necessary and correct this is. 

The current build-projects for resources just includes ModelicaUtilities.h; and doesn't do any special tricks, so it seems there is at least a tool-dependent way of building external libraries using the function in ModelicaUtilities.h; and the resulting library may even be portable between tools (I haven't checked).

henrikt-ma: > The term "standardized" is a bit ambiguous, would it be possible to be clearer?

Yes, it would be possible, but I'd also like to avoid introducing new ways to talk about what is determined by the document itself.  For example, we currently have formulations such as
> The license file is standardized.
or 
> Obfuscation and encryption are not standardized.

Hence, I'm somewhat skeptical to use a formulation like this instead:
```suggestion
The Modelica specification does not describe any way to build external libraries using the functions in \filename{ModelicaUtilities.h}.
```

henrikt-ma: > The current build-projects for resources just includes ModelicaUtilities.h; and doesn't do any special tricks, so it seems there is at least a tool-dependent way of building external libraries using the function in ModelicaUtilities.h; and the resulting library may even be portable between tools (I haven't checked).

The MSL is cheating by including its own copy of _ModelicaUtilities.h_.  We don't want a solution where there's a copy of this file in every library; it would make it impossible to develop the contents of the file.  Instead, there should be a mechanism for tools to tell (CMake projects à la #1668) where they provide this file.

HansOlsson: But even if MSL is "cheating" - it seems odd to specify that the approach that is currently used shouldn't work. I agree that we need something better, but requiring a change before we have a the better approach seems problematic.

henrikt-ma: That we state that there is no standardized way is not the same as saying that there is nothing that could work in non-standardized ways.

beutlich: > The MSL is cheating by including its own copy of _ModelicaUtilities.h_.

Which no longer is true as of https://github.com/modelica/ModelicaStandardLibrary/pull/3871.

henrikt-ma: Well, this just shifts the discussion to whether what is done inside _.CI_ could be considered a standardized non-cheating solution for building the binaries.  I still think it isn't; the CI is like its own tool environment, providing its own _ModelicaUtilities.h_ (as every tool environment should) and building the binaries in a way which hasn't been standardized (and probably won't be either, as everything points in the direction of going with CMake for standardization).

HansOlsson: This is getting messier, as we now we have https://github.com/modelica/ModelicaStandardLibrary/pull/4487 restoring the `ModelicaUtilities.h` and the separate discussion on building libraries depending `ModelicaUtilities.h` without including it.

I see some possibilities for the future:
- Have a standardized `ModelicaUtilities.h` as proposed in that PR - in that case that part(*) of this PR isn't needed
- Don't have a standardized `ModelicaUtilities.h`, but then it might be that we should describe in more detail how to work around that (will be a bit more text, but it is still possible).

I think we need to coordinate with/wait for https://github.com/modelica/ModelicaStandardLibrary/pull/4487 and thus haven't added it to the current milestone.

(*) However, this PR also explains how include could define either inline code, or include some code. That part might be generally useful even without `ModelicaUtilities.h`.

HansOlsson: Will add a more complete rewrite based on MSL 4.1.0.

chapters/functions.tex
```diff
@@ -2645,6 +2645,27 @@ \subsection{Utility Functions}\label{utility-functions}
...
+For example, there is no standardized way of obtaining a directory where the tool shall provide the \filename{ModelicaUtilities.h}.
+There is also no standardized library to link with to get a tool-specific implementation of the functions.
```
HansOlsson, line 2665: Based on https://github.com/modelica/ModelicaStandardLibrary/pull/4487
```suggestion
External libraries can built with the \filename{ModelicaUtilities.h} included within the Modelica Standard Library.
```
Since the resulting simulation model including the library will be linked with the tool-specific library by the tool I'm not sure how to mention that part.

chapters/functions.tex
```diff
@@ -2645,6 +2645,27 @@ \subsection{Utility Functions}\label{utility-functions}
 
 This section describes the utility functions declared in \filename{ModelicaUtilities.h}, which can be called in external Modelica functions written in C.
 
+The tool must ensure that the header is found by \lstinline[language=C]!#include "ModelicaUtilities.h"! within an \lstinline!Include! annotation (see \cref{annotations-for-external-libraries-and-include-files}); no \lstinline!IncludeDirectory! annotation is needed.
+
+\begin{example}
+The following usage patterns are common:
```
beutlich, line 2651: This is too much text where real examples seem more useful to me. Might be a good idea to have all three patterns available as proper examples, e. g. in the ModelicaCompliance library or ModelicaReference library or any other dedicated library.

chapters/functions.tex
```diff
@@ -2645,6 +2645,26 @@ \subsection{Utility Functions}\label{utility-functions}
...
+  The \lstinline!Include! annotation may first \lstinline[language=C]!#include "ModelicaUtilities.h"!, and then contain the external function definition where the utility functions may be used.
+\item
+  Like above, but moving the annotation content above to, say, \filename{myExtFun.c}, and then just do \lstinline[language=C]!#include "myExtFun.c"! in the annotation.
+\item
+  After \lstinline[language=C]!#include "ModelicaUtilities.h"!, the \lstinline!Include! annotation can also define a wrapper around a function in a linked library that does the real job, where the wrapper forwards the arguments as well as passes function pointers for a selection of functions from \filename{ModelicaUtilities.h}.
+\end{itemize}
+\end{example}
+
+\begin{nonnormative}
+When building external libraries, it is not possible to rely on the tool mechanism that provides \filename{ModelicaUtilities.h} for \lstinline!Include! annotations.
```
beutlich, line 2663: But I could add `IncludeDirectory="modelica://Modelica/Resources/C-Sources"` to my annotations of my external library project such that ModelicaUtilities.h is found on the include path.

henrikt-ma: This remark is not about building code in the annotations of a Modelica external function.  This is about building an external library with no Modelica code or concepts involved, with the intent of later linking with the library using the `Library`-annotation of a Modelica external function.

HansOlsson, line 2663:
```suggestion
When building external libraries independently of Modelica tools, it is not possible to rely on the tool mechanism that provides \filename{ModelicaUtilities.h} for \lstinline!Include! annotations.
```
Would this make the point clearer?

HansOlsson: I have tried to make that point clearer, see below.

chapters/functions.tex
```diff
@@ -2645,6 +2645,26 @@ \subsection{Utility Functions}\label{utility-functions}
...
+Instead, an external library project may contain its own instance of \filename{ModelicaUtilities.h}, for example obtained by making a copy of the instance included within the Modelica Standard Library.
```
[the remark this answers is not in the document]
henrikt-ma: The external include path idea doesn't work since there is no standardization of how a build system should figure out the location.  The MSL submodule approach seems like an utterly inefficient and perplexing way of getting hold of this one file.  This is why I proposed saying _may contain its own instance_; it describes one way of doing it, without saying it is the only way.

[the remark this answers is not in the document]
HansOlsson: > The external include path idea doesn't work since there is no standardization of how a build system should figure out the location.

Ok, I just thought that by specifying "use" we didn't have to explain any of that; whether you update the path in build-configurations in some way, copy the file, or do something else you just "use" the file.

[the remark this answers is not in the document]
HansOlsson: But I can live with this as well. Just giving beutlich a few more days to comment.

HansOlsson (approved):
LGTM

chapters/functions.tex
HansOlsson: skipping further wait.

chapters/functions.tex
HansOlsson: Real examples are useful, but adding them takes more time so I prefer to have that as separate PR if needed and merge this now.

PR #3468, opened 2024-01-29
Update 'constsize' example listings

henrikt-ma:
Addresses the issue with `protected` being present in the listings, pointed out by @casella in https://github.com/modelica/ModelicaSpecification/pull/3162#issuecomment-1908727141.

Should we also fix it in a maintenance-version of the conference paper?


henrikt-ma:
> Wrong target. It should be from MCP/0031 right?

Thanks!  Fixed.

HansOlsson (approved):
Ok! (For some reason previous comment was kept.)

PR #3470, opened 2024-01-29
Update example with over-determined spanning tree figure

henrikt-ma:
This is a planned follow-up to https://github.com/modelica/ModelicaSpecification/pull/3385, described in https://github.com/modelica/ModelicaSpecification/pull/3385#issuecomment-1889481739 (which I believe was the decision made in the MAP-Lang phone meeting on 2014-01-09).


HansOlsson (approved):
Ok.

eshmoylova (approved):

PR #3477, opened 2024-02-08
The Figure.group is a figure group, not a plot group

henrikt-ma:
Small cleanup, loosely related to #3475 and #3473.


HansOlsson (approved):
Looks ok, but one might also rephrase it as "Name of the group of figures"; but I guess that will involve more updates

henrikt-ma:
> Looks ok, but one might also rephrase it as "Name of the group of figures"; but I guess that will involve more updates

To me, changing to _Name of the group of figures_ is only a more verbose way of saying the same thing as today, and would suggest that we use the more verbose way of writing descriptions everywhere.  To me that seems mostly like a waste of space (not least in view of the line length restrictions in the PDF), so I prefer the style of short description strings combined with clear explanations in the text.

PR #3478, opened 2024-02-08
use vector fonts when generating the PDF

### LINKED ISSUE #3476: Pixelated fonts in Language Specification PDF
The official PDF of the MLS uses pixel-fonts. This makes readability sub-optimal. 

For a standalone PDF viewer (Acrobat Reader and PDF XChange) this results in fonts like (first words of the abstract):  
grafik

Browsers do usually do a bit of smoothing, resulting in
grafik

The solution for me was to install the LaTeX-package `cm-super` on the computer generating the PDF, as mentioned here: https://tex.stackexchange.com/questions/1291/why-are-bitmap-fonts-used-automatically/546882#546882. 

According to my tests, this generates a document with vector-based fonts and virtually unchanged file-size.

dietmarw:
Fixes #3476


henrikt-ma (approved):
Looks like the right thing to do, and now I see why I get nice results when building the document locally.

PR #3483, opened 2024-02-21
Define order within legend and plotting order

### LINKED ISSUE #3475: Plotting order

henrikt-ma:
Fixes #3475.

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -254,6 +254,7 @@ When the \lstinline!unit! of an \lstinline!Axis! is non-empty, it is an error if
 When \lstinline!legend! is not provided, the tool produces a default based on \lstinline!x! and/or \lstinline!y!.
 Providing the empty string as \lstinline!legend! means that the curve shall be omitted from the plot legend.
 Variable replacements, as described in \cref{variable-replacements}, can be used in \lstinline!legend!.
+The order of presentation within the plot legend corresponds to order of appearance in the \lstinline!curves! of a \lstinline!Plot!.
 
 \subsubsection{Escape Sequences}\label{text-markup-escape-sequences}
 
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -653,8 +653,8 @@ annotation(
 \end{lstlisting}
 \end{example}
 
-The graphics is specified as an ordered sequence of graphical primitives, which are described below.
-First, base class contents are drawn according to the order of the \lstinline!extends!-clauses, and then graphical primitives are drawn according to the order such that later objects can cover earlier ones.
+The graphics is specified as an ordered sequence of graphical primitives described below.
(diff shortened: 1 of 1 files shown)


chapters/annotations.tex
```diff
@@ -254,6 +255,10 @@ \subsubsection{Plot Curves}\label{plot-curves}
 When \lstinline!legend! is not provided, the tool produces a default based on \lstinline!x! and/or \lstinline!y!.
 Providing the empty string as \lstinline!legend! means that the curve shall be omitted from the plot legend.
 Variable replacements, as described in \cref{variable-replacements}, can be used in \lstinline!legend!.
+The order of presentation within the plot legend corresponds to order of appearance in the \lstinline!curves! of a \lstinline!Plot!.
+
+The \lstinline!zOrder! gives control over drawing order, with higher values corresponding to closer to front.
```
eshmoylova, line 260: I would prefer tying the order to the order in which they are drawn instead of the order in which they appear. So, saying, "lower values are drawn first." That way you do not need to worry about what is front and it is consistent with the way it is presented for graphical objects:
> First, base class contents are drawn according to the order of the extends-clauses, and then graphical primitives are drawn according to the order such that later objects can cover earlier ones.

DagBruck: While I agree, the usual terms when talking about z-order is front/back.

DagBruck (approved):
I understand. That's good enough for me.

chapters/annotations.tex
henrikt-ma: The relation to graphical objects is a relevant point, but for me the use of front/back is too established elsewhere to give up on easily.  For example, who would not understand the commands _Send to Back_ or _Send to Front_ in a menu of operations on any kind of 2D graphics object?

We could also consider aligning the quoted text:

> Base class contents are drawn behind the graphical primitives of the current class, ordered from back to front according to the order of the `extends`-clauses.  Graphical primitives are ordered from back to front according to order of appearance in the annotation.

Thus, with also @DagBruck's comment in mind, I'll await a more clear request for change from the language group before rewriting in terms of drawing order.  If @HansOlsson finds it relevant, this could be settled with a poll.

eshmoylova: OK, resolving.

eshmoylova (approved):

henrikt-ma:
Requesting new reviews after addressing @eshmoylova's comment regarding lack of symmetry with the description of graphical annotations.

eshmoylova (approved):

DagBruck (approved):

PR #3519, opened 2024-05-29
Improve CC-BY-SA-4.0 license visibility at GitHub

beutlich:
File by-sa.markdown was taken from https://github.com/idleberg/Creative-Commons-Markdown/releases/tag/v0.4.1

See https://github.com/beutlich/ModelicaSpecification/blob/improve-license-visibility/LICENSE.md for how it looks.

grafik

grafik

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,171 @@
-Copyright (C) 1998-2023 Modelica Association Project "Modelica Language"
-              All rights reserved.
+# Creative Commons Attribution-ShareAlike 4.0 International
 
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Having the license in the about-dialog is a good change.

Having a copy of the entire text of the license seems a bit messy - but after reading https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository that seems like the recommended way, and the license-detection mechanism guarantees that it is not modified.

PR #3533, opened 2024-06-18
Don't allow array subscripts in the argument to 'guess'

henrikt-ma:
As agreed in today's meeting.

The rationale is that `guess` is used to handle `start` attributes, and that it is not possible to have a `start` modification that only sets some of the entries of an array.


henrikt-ma:
@qlambert-pro that we need a similar restriction for `prioritize`, so I added that as well to this PR.

PR #3534, opened 2024-06-18
Update ReadMe.md

casella:
Added a short explanation in the Roadmap on how to actually read the proposal, which is not obvious at first sight.


henrikt-ma (changes requested):
Two things:
- This does not belong to the roadmap.
- I'm not sure it's a great idea to recommend starting with the extremely dry grammar.md.  I'd rather recommend directing readers to differences.md first, but if we can't agree on which one is better we could suggest them as equally good places to start.

henrikt-ma:
Perhaps a better place to put the new text is under _Summary_, right before _In a sentence (or two)_?

If placed there, it would also be good to recommend reading the rest of the _Summary_ and at least skimming the roadmap items before proceeding to the other documents.

PR #3542, opened 2024-06-24
Add full Modelica TestCase annotation also to Base Modelica

henrikt-ma:
During test implementation, we have quickly arrived at a point where we need to build Base Modelica test models, so even if we are not going to be using the `TestCase` annotation ourselves, this indicates that the annotation is just as relevant for Base Modelica as it is for full Modelica.


casella (approved):
LGTM

HansOlsson (commented):
Why is it in the package not the model?

HansOlsson (changes requested):
As indicated I don't see why it is on the package not model

casella:
> Why is it in the package not the model?

Sorry, I missed that. Of course it should be in the model

henrikt-ma:
> Why is it in the package not the model?

Looks like a mistake to not place it side by side with `experiment`.  Fixed.

PR #3546, opened 2024-06-28
Fix typo "uses" vs "use"

eshmoylova:
"classes... uses" -> "classes... use"


henrikt-ma (approved):
Looks good.

HansOlsson (approved):
Ok, must have been mislead by the second "class".

PR #3575, opened 2024-09-19
Format 'record' and 'type' as code

henrikt-ma:
Please see second commit for actual change.


henrikt-ma:
> However, the latter argument also apply to the "record or type".

To me, it makes a big difference that these are preceded by _specialized class_, and that this section of the specification is already generally using the inline code format of the class specialization concepts which are being related to each other.  The first occurrence of _record_ in the sentence matches the style of this nearby sentence:
> The elements of a record shall…

HansOlsson (approved):

PR #3581, opened 2024-10-03
Avoid scare quotes in text

henrikt-ma:
I just happened to stumble upon this due to the bad use of straight quotes in text.  Anyway, it is not one of the situations where we want the nuance of scare quoting.


HansOlsson (approved):
Ok, but we should likely have a better name. One might think that a _default connection_, would be similar to a default value - but it's more a generic default connection.

PR #3582, opened 2024-10-09
Require that ranges and conditions around 'connect' are evaluable

henrikt-ma:
I don't see how the connect set handling would work unless it is known whether a `connect`-equation is present or not, so I'm hoping that this isn't a controversial restriction of variability.


HansOlsson (approved):
Must be left-over from pre-evaluable times.

PR #3594, opened 2024-10-28
Fix cases of declaration assignment in function listings

henrikt-ma:


HansOlsson (approved):
Ok. I could not find any other such cases

PR #3606, opened 2024-11-07
Break down example of quoted identifier

henrikt-ma:
This is just a small enhancement of #3592.  I had it in mind when reviewing that PR, but must have forgotten to write it down.

The idea is that when we speak of both quoted identifiers and complex names, the example for quoted identifiers should illustrate just quoted identifiers without at the same time illustrating complex names.


henrikt-ma:
Realizing that this is a duplicate of #3593.  Closing.

PR #3619, opened 2024-11-29
Update "parameter" to "evaluable" in connectors.tex

eshmoylova:
Changes in #3582 affected only equations.tex. I thought we should check connectors.tex as well. I believe those are all instances where "parameter" should be "evaluable." I believe it is ok to use "parameter" on line 854, so did not include it. But let me know if this should also be evaluable.


HansOlsson (approved):
Looks good to me

PR #3623, opened 2024-12-06
Add missing 'constant' for members of License

henrikt-ma:
Just a minor thing I stumbled upon when trying to make sense of these annotations.


HansOlsson (approved):
Ok

PR #3630, opened 2024-12-19
Generalize vendor-specific markup in figures

henrikt-ma:
We have discovered the need for choice of display unit for variable replacements in figures.  While this could eventually become standardized, what is needed in the short term is a way to express this using vendor-specific information.  The currently allowed used of vendor-specific markup would require awkward nesting of markup constructs, so we propose to rather equip both variable replacements and links with vendor-specific markup.

By not only proposing a solution for variable replacements, we want to ensure that we come up with a reasonably consistent design that covers vendor-specific handling of:
- alternative content
- fine control over variable replacements
- fine control over links

The proposed syntax for links, `%[text]__AVendor(data)(url)` may not seem obvious at first, but please note that `%__AVendor(data)[text](url)` is already defined to be parsed as `%__AVendor(data)[text]` (vendor-specific alternative content) followed by `(url)`.

In summary, this PR proposes that there should be a total of three ways to attach vendor-specific markup (in order corresponding to the list above):
- `%__AVendor(data)[text]` (already standardized)
- `%__AVendor(data){variable}`
- `%[text]__AVendor(data)(url)` or `%__AVendor(data)(url)`

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -343,7 +343,7 @@ Ties are resolved using order of appearance in the \lstinline!curves! of a \lsti
 \subsubsection{Escape Sequences}\label{text-markup-escape-sequences}
 
 In an attribute inside a figure where the variable replacements of \cref{variable-replacements} or the text markup of \cref{text-markup-in-captions} can be used, the following use of \firstuse[text markup escape sequence]{text markup escape sequences}\index{escape sequence!text markup} applies.
-These escape sequences are applied after the application of other markup, and is not applied at all inside some of the other markup, see details for the respective markup.
+These escape sequences are applied after the application of other markup, and are not applied at all inside some of the other markup, see details for the respective markup.
 
 The percent character `\%' shall be encoded \lstinline!%%!.
 The following are all the recognized escape sequences:
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -363,6 +363,14 @@ The following are all the recognized escape sequences:
 With the percent character being encoded as \lstinline!%%!, the behavior of \lstinline!%! appearing in any other way than the escape sequences above, for variable replacement (\cref{variable-replacements}), or for the text markup (\cref{text-markup-in-captions}) is undefined, and thus possible to define in the future without breaking backward compatibility.
 \end{nonnormative}
 
+\subsubsection{Vendor-Specific Markup}\label{vendor-specific markup}
+
+\firstuse{Vendor-specific markup} takes the form \lstinline!__$\mathit{nameOfVendor}_{1}$($\mathit{data}_{1}$)$\ldots$__$\mathit{nameOfVendor}_{n}$($\mathit{data}_{n}$)!, where $n \geq 1$.
+It is only allowed as part of constructs described in \cref{variable-replacements} and \cref{text-markup-in-captions}, where it will be denoted by $\mathit{vendorSpecificMarkup}$.
+The $\mathit{nameOfVendor}$ consists of only digits and letters, and shall only convey the name of the vendor defining the meaning of the associated $\mathit{data}$.
+Text markup escape sequences don't apply inside the $\mathit{data}$, implying that it cannot contain the closing parenthesis, `)'.
+A tool which does not understand any of the vendor-specific meanings will always be able to safely ignore all vendor-specific markup.
+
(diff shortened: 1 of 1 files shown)


henrikt-ma:
@casella, I am adding you as a reviewer primarily due to your role as a representative of MAP-Lib, since I find it important that MAP-Lib is following the development of the figure features of Modelica in order that it can be adopted (more broadly than a single example) by the MSL.

henrikt-ma:
Regarding the awkward alternative of nested constructs for display unit selection, this is what it could look like, not forbidden by the current specification:
```
%__AVendor(?displayUnit=mm)[%{integrator.y}]
```

Related, it seems unfortunate to me that the current specification does not forbid putting a link in the text of another link:
```
%[foo %[bar](url2) baz](url1)
```

If we introduce a dedicated way for vendor-specific handling of variable replacements, perhaps we should generally forbid the use of nested markup for all combinations of variable replacements, links, and alternative content?  Of course, it would be backwards a backwards incompatible change, but I see a good chance that nobody has put nested constructs like this in any library yet.  The benefit would be that we lower the barrier for tools to achieve correct implementation, and we resolve the question about how to present nested links.

If there is popular demand in the future, we can relax the unnecessary restrictions then, like allowing variable replacements inside the other constructs.

henrikt-ma:
> These are not easy to read for me, and I'm a bit hesitant to add something for future changes.

Please feel free to suggest alternative syntax.

The new example about future changes was just a continuation of the example we already have for vendor-specific content.  Of course, it can be omitted if it is considered too obvious what vendor-specific markup for variable replacement could be used for.

chapters/annotations.tex
```diff
@@ -363,6 +363,14 @@ \subsubsection{Escape Sequences}\label{text-markup-escape-sequences}
 With the percent character being encoded as \lstinline!%%!, the behavior of \lstinline!%! appearing in any other way than the escape sequences above, for variable replacement (\cref{variable-replacements}), or for the text markup (\cref{text-markup-in-captions}) is undefined, and thus possible to define in the future without breaking backward compatibility.
 \end{nonnormative}
 
+\subsubsection{Vendor-Specific Markup}\label{vendor-specific markup}
+
+\firstuse{Vendor-specific markup} takes the form \lstinline!__$\mathit{nameOfVendor}_{1}$($\mathit{data}_{1}$)$\ldots$__$\mathit{nameOfVendor}_{n}$($\mathit{data}_{n}$)!, where $n \geq 1$.
+It is only allowed as part of constructs described in \cref{variable-replacements} and \cref{text-markup-in-captions}, where it will be denoted by $\mathit{vendorSpecificMarkup}$.
+The $\mathit{nameOfVendor}$ consists of only digits and letters, and shall only convey the name of the vendor defining the meaning of the associated $\mathit{data}$.
+Text markup escape sequences don't apply inside the $\mathit{data}$, implying that it cannot contain the closing parenthesis, `)'.
```
eshmoylova, line 371: There is a quote and a backtick round ) at the end. Is that intentional?

chapters/annotations.tex
```diff
@@ -404,6 +412,16 @@ \subsubsection{Variable Replacements}\label{variable-replacements}
 (Titles making use of \lstinline!%class! or \lstinline!%name! would then only lead to ugly duplication of this information.)
 \end{nonnormative}
 
+Vendor-specific markup can be added to a variable replacement in the form \lstinline!%$\mathit{vendorSpecificMarkup}${$\mathit{variable}$}!.
+The vendor-specific markup must not fundamentally alter the meaning of the variable replacement, in order to ensure that a tool can safefly ignore all vendor-specific markup and still obtain a result that fits the current context.
```
eshmoylova, line 416: Safefly
```suggestion
The vendor-specific markup must not fundamentally alter the meaning of the variable replacement, in order to ensure that a tool can safely ignore all vendor-specific markup and still obtain a result that fits the current context.
```

chapters/annotations.tex
eshmoylova: Seems intentional based on the pdf. Resolving.

DagBruck (approved):
It's ugly but probably works. It is supported by Dymola in the sense that we ignore such vendor-specific markup.

HansOlsson (approved):
Ok, I'm not happy about describing possible future changes - but that was already existing - just re-organized.

PR #3637, opened 2025-01-20
Require that external functions have explicit external call

henrikt-ma:
This addresses the problem that we didn't specify any difference in semantics of what this means in Base Modelica:
```
pure external function 'L.P.f'
  input Real 'u';
  output Real 'y';
external "C";
end 'L.P.f';
```
Clearly, the Modelica semantics makes no sense, since `'L.P.f'` is not a valid identifier in C.

The PR proposes to go with the first of the two possible solutions that come to mind:
- Don't allow leaving out the _external-function-call_.
- Specify a different way of deriving the default name of the C function, by requiring that the external function's name is a structured Base Modelica identifier, where it is the last identifier of the unmangled name that gives the name of the C function.


perost:
I agree in principle, but it seems something's gone wrong with the PR since it includes hundreds of old commits.

henrikt-ma:
> I agree in principle, but it seems something's gone wrong with the PR since it includes hundreds of old commits.

Wrong base. Thanks for pointing it out!

perost (approved):

PR #3639, opened 2025-01-22
Remove empty rest-of-line comment in listing

henrikt-ma:
Removing the comment is the simple solution.  Alternatively, we could figure out what to say in the comment.


HansOlsson (approved):
Ok, the empty comment (and rest of the example) was added in version 1.2 so I don't think it is important to figure out why after all of these years.

PR #3644, opened 2025-01-29
Fix typos

maltelenz:


HansOlsson (approved):
I don't understand how that was missed in the previous review.

PR #3659, opened 2025-03-25
Remove excessive semicolons

beutlich:


HansOlsson (approved):
Yes, they should be removed.
For C++11 they are legal - but not needed.
For C I don't if it is legal (if so in which version).

PR #3666, opened 2025-03-31
Add connectorSizing to document index

henrikt-ma:
On behalf of @maltelenz.  While it feels somewhat inconsistent to not add all members of all annotation records to the document index, I can see that things like `connectorSizing` which have their own dedicated sections in the text may be considered more relevant for the document index than other annotation record members.


HansOlsson (approved):
Seems ok

PR #3682, opened 2025-04-28
Sentence-based line breaks and small fixes in inheritance.tex

henrikt-ma:
This is a follow-up to #3679, where the new content accidentally contained non-sentence-based line breaks due to a copy/paste mistake.

As I was inspecting the generated PDF after the whitespace cleanup, I discovered two minor issues which were addressed in separate commits.


HansOlsson (approved):
LGTM; no actual changes

PR #3686, opened 2025-05-05
Don't use Markdown code quotes in LaTeX source

henrikt-ma:


HansOlsson (approved):
Ok, due to https://github.com/modelica/ModelicaSpecification/pull/3488

PR #3688, opened 2025-05-07
Unitful literals

henrikt-ma:
This is the design for attaching units to literals that was discussed at the most language design meeting in Linköping held in March this year.  The proposal is an alternative to #3668, with emphasis on simple design, ease of use, and human readability.

This design was originally implemented in SimulationX, and later given two parallel implementations in System Modeler - one for testing it in regular Modelica expressions, and one for (undocumented) use in vendor-specific annotations.


henrikt-ma:
The ease of use and readability provided by this proposal is important for the acceptance of this as a correct way to address the unit error in a situation like this:
```
Real distance(unit = "m") = 100;
Real v(unit = "m/s") = distance / 9.58; // Unit error
Real v(unit = "m/s") = distance / 9.58's'; // Correct
```

This PR makes no changes to unit checking; the following should be rejected as inconsistent:
```
Real distance(unit = "m") = 1000'cm';
```

HansOlsson:
> This PR makes no changes to unit checking; the following should be rejected as inconsistent:
> 
> ```
> Real distance(unit = "m") = 1000'cm';
> ```

The latter isn't even close to **beginning** to meaningfully adress the actual problems this would introduce in practice.
Please look at actual models as in MSL.

HansOlsson (changes requested):
To avoid merging by mistake:
As repeatedly indicated this makes it easy to introduce non-coherent units, which requires a detailed analysis.

henrikt-ma:
It is hard to think of a programming language where units and unit checking matter more than in Modelica. It is just silly that the language is lacking this basic construct of constructing unitful literals.

Applications are everywhere:
- Making expressions and equations with literals unit consistent without the need to introduce protected helper variables just to attach the missing units.
- Avoiding dangers of the wildcard effect, see example below.
- Last but not least, clarity; even if an expression or equation with literals is unit consistent according to the unstated unit checking rules of Modelica, making the implicit units of the literals explicit is a great enhancement of clarity.

Example of the day:
```
model AcuteAngle
  parameter Modelica.Units.SI.Angle a = 0;
equation
  assert(abs(a) < 90, "The angle must be acute.");
end AcuteAngle;
```
Due to the lack of clarity and the wildcard effect, the error above is easy to make, and the model developer who knows that the model only works for acute angles might fail to check that the assertion actually works.

If the developer would have used a unitful literal instead, a unit inconsistency would have revealed the problem:
```
  assert(abs(a) < 90'deg', "The angle must be acute.");
```
When corrected, the attached unit adds clarity and removes any uncertainty related to the wildcard effect:
```
  assert(abs(a) < 1.57'rad', "The angle must be acute.");
```
If the modeler decides to use `Angle_deg` instead (maybe it is the angle of a `FixedRotation`), the need to also update the assert would be detected.

henrikt-ma:
> As repeatedly indicated this makes it easy to introduce non-coherent units, which requires a detailed analysis.

Let's not pretend that we can't perform a decent amount of unit checking, and let's not assume that the feature would be massively misused to exploit any gaps we still have in the unit checking semantics.

The alternative, using numeric literals without unit and relying on the wildcard effect is so much worse than the tiny risk of introducing a unitful literal with non-coherent unit in a place where it wouldn't be caught by unit checking.

HansOlsson:
> The alternative, using numeric literals without unit and relying on the wildcard effect is so much worse than the tiny risk of introducing a unitful literal with non-coherent unit in a place where it wouldn't be caught by unit checking.

The original design was to have equations in coherent units; and have non-coherent units in the user interface at the boundary.
That design doesn't have any of those issues, and was made exactly because units matter.

Apart from that the previously raised issues with temperatures, and rotations are still unresolved.

HansOlsson:
> I think we need external input to make progress here. 

I don't see it as meaningful to involve more people when there are still no answers regarding how possibly non-coherent units in equations would actually work, especially for temperatures and rotations.

HansOlsson:
> > I think we need external input to make progress here.
> 
> I don't see it as meaningful to involve more people when there are still no answers regarding how possibly non-coherent units in equations would actually work, especially for temperatures and rotations.

It's also misleading to tell others that the issue is 'unitful literals' when the problem is that it's suddenly easy to add possibly non-coherent units like `2 'kW'`, `50 'Hz', `12 'rad/s'`, or `200 'degC'` inside equations, and then rely on some yet-to-be-specified unit-check to figure out that it wasn't the right unit.

I have still not seen anything meaningful addressing those issues, or how it would work in practice.

henrikt-ma:
> I think that, for consistency, we should then restrict unitful literals to only use SI-units (including the listed 7 non-SI units that are allowed in the SI system), as we do with unit attributes. @henrikt-ma do you agree?

Yes, the same sort of _recommendation_ makes sense for unitful literals.  I'll add it.

henrikt-ma:
> Also, should we add the three example cases listed in the minutes of our meeting as non-normative text for further clarification?

Personally I would have preferred to not include a bigger example like that for a "simple" construct like this, but considering the wild discussions i Rotkreutz I realize that it might be unrealistic to think that the average specification reader would be annoyed by such a "trivial" example.  I'll add the example, and then we can take the discussion from there.

chapters/unitexpressions.tex
```diff
@@ -104,3 +104,34 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+An acceleration of 9.8 meters per square second can be expressed as \lstinline!9.8'm/s2'!.
+\end{example}
+
+\begin{example}
+Adding 2~centimeters to a length.
+A user has set \lstinline!x! to 10~centimeters with the use of \lstinline!displayUnit!, and intends to make \lstinline!y! 2~centimeters bigger than \lstinline!x!:
+\begin{lstlisting}[language=modelica]
+  parameter Real x(unit = "m", displayUnit = "cm") = 0.1;
+equation
+  y = x + 2; /* Intention: Add 2 centimeters to x. */
```
cbu-hw: I think it could be clearer in the code example by itself that this intention is _not_ fulfilled, similar to how many other code examples in the spec point out errors or legal/illegal character in a comment right at the line of the problem.

```modelica
  y = x + 2; // Intended: Add 2 centimeters to x; actual: adds 2 meters
```
If the problem should then repeated in the line below is debatable.

Note: this suggestion also matches the comment style `//` that I perceive as predominant in short MLS code examples.

chapters/syntax.tex
```diff
@@ -405,8 +405,11 @@ \subsection{Expressions}\label{expressions1}
    | "{" array-arguments "}"
    | end
 
-UNSIGNED-NUMBER :
-   UNSIGNED-INTEGER | UNSIGNED-REAL
+numeric :
+   ( UNSIGNED-INTEGER | UNSIGNED-REAL ) [ unit-of-measurement ]
+
+unit-of-measurement :
+  Q-IDENT
```
cbu-hw: Has it been considered to introduce the unit of measurement in the fashion of Julia , i.e. https://docs.julialang.org/en/v1/manual/strings/#non-standard-string-literals (string literals prefixed by an identifier)? 

This is used to great effect in [Unitful.jl](http://juliaphysics.github.io/Unitful.jl/stable/#Usage-examples) with a `u` prefix to ensure that unit strings are distinct from other strings: 
```julia
1u"kg" == 1000u"g"
```
This opens the door to later introduction of other specific string types (like for regular expressions,...).

Concretely, if I got this correct, this part of the grammar would turn into
```
unit-of-measurement :
  "u" Q-IDENT
```

cbu-hw:
I like the proposal (saw it first at the conference in your talk) and the extended example (in my book, more illustrative examples are typically better than fewer)!

chapters/unitexpressions.tex
henrikt-ma: The difficulty is that we don't have a specification of how unit checking actually works, so in principle we can only speculate what this code actually does in different tools.  Explaining this in the comment would require too much space, so I gave the explanation in the example text below instead.

Regarding the style of comment, we have a mix of styles in the specification, and similar to different styles of whitespace around binary operators I think the current plan is to keep it this way.

chapters/syntax.tex
henrikt-ma: With that idea, it would have been tempting to use a string instead of a quoted identifier:
```
Real x = 1u"cm";
```

You may already have thought about this, but let me elaborate a bit for those who haven't. The problem I see is that this gets ugly if we imagine a general mechanism for string prefixes, as this means that the intention is that `1` followed by the unit string literal `u"cm"` means to attach the unit `"cm"` to the numeric literal.  I would then expect that `u"cm"` is valid syntax for expressing a prefixed string, but this breaks down for the same reason we couldn't use the syntax `1"cm"`:
```
String u = "hello";
String myUnitString = u"cm"; // Currently parsing as u in declaration equation, and "cm" as description string.
```

Now, if we would have to avoid string syntax for this reason and go for quoted literals instead – as you suggested – I am afraid that we would start developing too much string syntax around literals that are not strings to start with.  Also remember that quoted literals don't support `S-CHAR`, and that there isn't a standardized way to interpret escape sequences in quoted identifiers.

To summarize:
- Introducing prefixed strings would break backwards badly due to the description string syntax.
- Prefixed quoted identifiers would probably come with too many restrictions to be a worthy Modelica incarnation of prefixed strings.

cbu-hw: Ah, the description strings, right, I didn't think of that. :-| That's a concern that Julia doesn't have, of course.

henrikt-ma: Do you still want to keep this thread open, or can we resolve?

cbu-hw: You can resolve (I don't have the button, otherwise I would have already), also my thread on the example comment.

HansOlsson:
Based on the discussion in Lucern it seems this is the preferred syntax, and that the preferred semantics is that it is automatically converted to suitable units (normally base units). 

The current proposal intends to have a more restrictive semantics, that may possibly work.
However, it doesn't clearly the define semantics - and that must be improved; and I have added most of the needed changes at the end.

**Background**
As far as I understand SimulationX already has the semantics that `Real(unit="m")=1 'cm';` roughly means: `Real(unit="m", displayUnit="cm")=0.01;` - and Dymola has similar semantics in the GUI for a slightly different syntax.

We might delay conversions (and/or restrict them - e.g., to parameter bindings with unit) provided that:
- We do not later change the new semantics for valid models
- We have some indication that this is the direction

I have some doubts whether completely delaying it will actually work as tools will naturally attempt to support conversions, in some way. (Those ways should then converge.) 

In order to simplify the discussion I find it useful to only consider unitful literals in parameter bindings as other cases can be reduced to that, e.g. the example code `y = x + 2'cm';` could be handled as:
```
  parameter Real temp=2'cm';
equation
 y=x+temp;
```
I'm not saying that we should restrict the unitful literals to these cases, merely that this is useful for understanding the general issues as any issue for unitful literals in expressions can be reproduced using only unitful literals in parameter bindings. It also shows that automatically converting unitful literals in general isn't more complicated than only converting them in parameter bindings.

Specifically considering parameter bindings show a number of issues related to the following areas:
- If we don't have a unit for the parameter we don't know whether it is valid (without conversion) or which value to use (with conversion)
- If it is a temperature we also need absoluteValue
- Some issues related to angular frequencies (etc)
- Some units should not be mixed even if they are compatible

( Regarding conversion: if models only had used SI-base-units one could have automatically added a conversion to SI-base-units. However, since that is not the case one needs to check whether that is the case first, and that check seems as complicated as inferring the actual unit. )

**Proposed semantics**
However, as long as we admit that there are issues I believe that it would be possible to handle most cases in some pragmatic way:
- Only allow unitful literals when the unit can be inferred (based on other information) and the unitful literal is consistent with the inferred unit; I assume that is the intention of the PR but it is not explicitly written
- Only allow unitful literals that may involve offsets (i.e., temperatures) if absoluteValue can be inferred (especially when we add conversions, but even otherwise); see #3748
- And some specific restrictions for angular frequencies (and related quantities); will detail later (especially when we add conversions, but even otherwise)
- Allow diagnostics if the unit of the unitful literal and the inferred unit shouldn't be mixed (like "N.m" for an energy variable)

The key part is that goal isn't to handle everything, but instead that many problematic constructs should generate errors (or warnings for the last case).

This is important, as users (and tools) can currently ignore/disable unit checking and inference (or at least ignore the diagnostics), but once we start considering unit conversion one cannot unit checking - as unit conversion and ignored unit checking will give different results for the same model (and unit conversion for models with unit errors will be a mess.)

The implication is that:
```
model MValid
  Real x(unit="m"), y;
equation
 y=x+0.02'm';  //Or using temp as above.
end MValid;
```
would be legal, whereas
model MMixing1
  Real x(unit="m"), y;
equation
 y=x+2'cm'; 
end MMixing1;

model MMixing2
  Real x(unit="cm"), y;
equation
 y=x+0.02'm'; 
end MMixing;
```
would currently both be in-valid (they require conversion), and
```
model MNonValid
  Real x, y;
equation
 y=x+0.02'm'; 
end MNonValid;
```
would not be valid as we couldn't infer the unit for x (so it could be MMixing2 or MValid).

Additionally, this implies:
- Unit inference can start small, and add more rules - without breaking unitful literals in valid models
- Automatic unit conversion of unitful literals can be added safely with this restriction
- Connecting such components with other components will not invalidate the unitful literals

Or in summary: we can safely extend both models and the language.

( An interesting consequence is that some ambiguous unitful literals could be disambiguated by the inferred unit (so "T" could mean Tesla or Tonne depending on the unit; and similarly for the different pounds), but I don't think that's a good idea. )

File: chapters/unitexpressions.tex [hunk 3]
```diff
@@ -104,3 +104,34 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
 
 The unit expression \lstinline!"T"! means tesla, but note that the letter \lstinline!T! is also the symbol for the prefix tera which has a multiplier value of 10\textsuperscript{12}.
 \end{example}
+
+
+\section{Unitful Literals}
+
+The \lstinline[language=grammar]!numeric! in the grammar allows a unit to be attached to a numeric value.
+In this case, the unit is given as a \lstinline[language=grammar]!Q-IDENT! rather than a string, since the use of a string could be ambiguity interpreted as \lstinline[language=grammar]!description-string! in some contexts.
```
HansOlsson:
```suggestion
In this case, the unit is given as a \lstinline[language=grammar]!Q-IDENT! rather than a string, since the use of a string could be ambiguously interpreted as \lstinline[language=grammar]!description-string! in some contexts.
```
To me "ambiguity interpreted" looks odd, and I prefer "ambiguously interpreted"; (possibly with "be" in the middle)

https://books.google.com/ngrams/graph?content=ambiguity+interpreted%2C+ambiguously+interpreted&year_start=1800&year_end=2022&corpus=en&smoothing=3

chapters/unitexpressions.tex
```diff
@@ -104,3 +104,34 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+equation
+  y = x + 2; /* Intention: Add 2 centimeters to x. */
+\end{lstlisting}
+The result is that 2~meters instead of 2~centimeters is being added to \lstinline!x!.
+
+Using a unitful literal makes the intention clear:
+\begin{lstlisting}[language=modelica]
+  y = x + 2'cm';
+\end{lstlisting}
+It is currently not expected that tools will perform an implicit unit conversion to make the addition possible, in which case the addition should be rejected as a unit inconsistency.
```
HansOlsson:
```suggestion
Tools must infer the unit information of the unitful literal, including absolueValue when applicable, from the rest of the model, and verify that the given unit is compatible.
It is an error if the needed unit information cannot be inferred.
Diagnostics may be given for consistent units that shouldn't be mixed (like \lstinline!"N.m"! and \lstinline!"J"!).
Special care is needed to handle rotational frequencies.
It is currently not expected that tools will perform an implicit unit conversion to make the addition possible, in which case the addition should be rejected as a unit inconsistency.
```
So, something like this.

HansOlsson:
> > Based on the discussion in Lucern it seems this is the preferred syntax, and that the preferred semantics is that it is automatically converted to suitable units (normally base units).
> 
> Let's not allow anything involving conversion to creep into the scope of this PR. It is crucial that this PR is only about the syntax, so that we don't need to make any assumptions about the rules for unit checking which are still under development.
> 
> Then, once we have rules for unit checking, it will be a much better time to talk about implicitly converting these literals to units that don't lead to unit inconsistency.

I didn't add the semantics of unit conversion, I only made the semantic restriction that you are only able to use the construct if you can infer the unit (and the unit of the unitful literal is compatible with that) **fully clear**.

As far as I understand SimulationX users can write `parameter Modelica.Units.SI.Length len=1'cm';` and it is interpreted in the natural way as `0.01 m`.  That's also what users seem to expect.

If we don't specify any semantics some could follow the normal idea that units can be ignored (or disable unit checking) and think that it means the same as `parameter Modelica.Units.SI.Length len=1;`

The current proposal sort of suggests that the model should be invalid due to unit inconsistency, which requires that the unit is inferred. The specification shouldn't merely _suggest_ semantics for the value of parameters (that so significantly impact the result), and thus I proposed to explicitly add it to the proposal - and in particular clarify that if the unit cannot be inferred it is an error; so that tools cannot say "we don't infer units so it's obviously valid". (Even if the conversion-check is the hard part, and the actual conversion is the easy part.)

That unclarity for what a model means (is the value of `len` 1, 0.01, or is the model just invalid?) would be **disastrous** for users, or in other words syntax without semantics is literal non-sense.

(If we didn't allow `Real len(unit="cm")` one could always interpret it as converted to SI-base units, but that's not the case. It would also theoretically be possible to limit the syntax to SI-base-units, but that seems unlikely to work in practice. Adding `inUnit(..., "m")` makes it more verbose, without removing the problem, as there's the possibility of  `Real len(unit="cm")=inUnit(1'cm', "m");` - or users just skipping that part.)

Importantly, it is easy to determine the unit in this case (it's part of the type for the variable) and it would be silly to ignore it, but as I understand the intended goal is to allow them in places where the unit inference is less straightforward - with the same problems for the unit and which value to use or whether it is invalid.


Therefore, if we want to _just add the syntax without problems_ we could say that _it is currently only allowed in simple bindings for variables with unit_ and it is then automatically converted. Alternatively, we could state that they are only allowed in those cases and it is in an error if the units would require a conversion - but as previously stated the actual conversion is the easy part.

As far as I understand that possibility:
- Introduces the syntax
- Corresponds to the SimulationX case (AFAIK)
- It avoids rules for unit checking/inference etc; as the unit is already explicitly given
- Is compatible with my proposed changes above, as e.g., temperatures require absoluteValue - and it is a bad idea to write `parameter Modelica.Units.SI.Torque t = 1'W'`

chapters/unitexpressions.tex
```diff
@@ -104,3 +104,34 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+  parameter Real x(unit = "m", displayUnit = "cm") = 0.1;
+equation
+  y = x + 2; /* Intention: Add 2 centimeters to x. */
+\end{lstlisting}
+The result is that 2~meters instead of 2~centimeters is being added to \lstinline!x!.
+
+Using a unitful literal makes the intention clear:
+\begin{lstlisting}[language=modelica]
+  y = x + 2'cm';
+\end{lstlisting}
```
HansOlsson, line 233:
```suggestion
\end{lstlisting}
Tools must infer the unit information of the unitful literal, including absolueValue when applicable, from the rest of the model, and verify that the given unit is compatible.
It is an error if the needed unit information cannot be inferred.
Diagnostics may be given for consistent units that shouldn't be mixed (like \lstinline!"N.m"! and \lstinline!"J"!).
Special care is needed to handle rotational frequencies.
```

HansOlsson: To make it clearer that it is just adding lines to clarify, and that the following lines are unchanged.

HansOlsson:
To be concrete consider:
```
model Ok
  Modelica.Units.SI.Length len=1'm';
  Modelica.Units.SI.TemperatureDifference Td=1'degC';
end Ok;

model Bad
   Modelica.Units.SI.Length len=1'degC';
end Bad;

model NeedConversion
   Modelica.Units.SI.Length len=1'cm';
   Modelica.Units.SI.Temperature T=20'degC';
end NeedConversion;

model Ambiguous
   Real x=1'cm';
   Real y(unit="K")=1'degC';
end Ambiguous;
```
Either we agree that Ok is ok while Bad, and Ambiguous should be errors, and NeedConversion is also an error (unless we add conversions) - and then I see no harm in clarifying that. Or we are adding something without knowing what it means.

gkurzbach:
Because we have no unit checking in SimulationX, the numbers with units in quotes are always converted to the SI base unit, so 1'cm' is equal to 0.01'm' but the unit is dropped then, so 0.01 remains. If there would be unit checking, the unit could be kept and checked for consistency with the inferred unit. Further, if the unit are not the same, but have the same SI base unit, the result could be converted to the inferred unit. So, if the inferred unit would be 'mm', 0.01'm' would be converted to 10 'mm' and the unit then could be dropped, and 10 remains.

chapters/unitexpressions.tex
```diff
@@ -104,3 +104,40 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+\begin{example}
+An acceleration of 9.8~meters per square second can be expressed as \lstinline!9.8'm/s2'!.
+\end{example}
+
+A unitful literal with a temperature unit represents an absolute value of temperature.
+
+\begin{example}
+The temperature \lstinline!20'degC'! converted to kelvin would be 293.15~kelvin, not 20~kelvin.
+\end{example}
+
```
HansOlsson:
```suggestion
```
The temperature change is:
- A step backwards, especially as Modelica does have temperature differences
- Inconsistent with the rest of the proposal that doesn't do unit conversions, so `1'cm'` isn't converted to 0.01 meter, and `1'mK'` isn't converted to 0.001 kelvin, but suddenly `20'degC'` is converted to 293.15 kelvin?

(The existing alternative is to do it safely for all cases.)

henrikt-ma: No, there is no sudden conversion of `20'degC'` to any other unit.  An _extremely_ central feature of this PR is that it by itself does not come with unit conversion!

Implicit unit conversion is a delicate topic which will be much more sensible to approach once we have defined what unit consistency means in Modelica, but if we are a little crazy we could sneak it in in special situations even before we have unit checking rules. But not in this PR.

What the example wants to say is just that this is valid and will display `t` as 293.15 K (using the `absoluteValue` annotation for extra clarity):
```
Real t(unit = "degC", displayUnit = "K) = 20'degC' annotation(absoluteValue = true);
```

This is a unit error since we don't do implicit unit conversions yet:
```
Real t(unit = "K") = 20'degC';
```

Anyway, I agree that it is a cool feature to automatically infer whether `20'degC'` is absolute or relative, so I must have misunderstood when I thought that something more simple-minded was requested.  I'll replace it with the fancy thing.

henrikt-ma: The old formulation has now been removed.  Resolving.

henrikt-ma:
> So, instead of making a non-normative unclear anti-cheating rule (which to me reeks of "enumerating badness" or blacklisting) my proposal was simply to state the tool must be able to infer the unit of the literals to use them (i.e., only use trusted values).

I'll make an initial attempt, and then we can attach the rest of this discussion to the added lines.

chapters/unitexpressions.tex
henrikt-ma: I am aiming for a rule that doesn't forbid this:
```
Real pressure(unit = "Pa") = 5'N' / 20'm2';
```
So instead of speaking of inferring the unit of the unitful literal from the context, I am requiring that unit consistency can be verified.

I find saying something about this model outside the scope of this PR:
```
Real force(unit = "N") = 3;
Real distance(unit = "m") = 5;
Real work(unit = "J") = force * distance;
```

Similarly, I find statements regarding rotational frequencies to be a separate topic.

HansOlsson: It wasn't for the case above, or even for:
```
Real force(unit = "N") = 3;
Real distance(unit = "m") = 5;
Real lever(unit="m")=2;
Real work(unit = "J") = force * distance;
Real torque(unit="N.m") = force * lever;
```
but to _allow_ diagnostics for the subtly different:
```
Real torque2(unit="N.m") = work;
```
Similarly, for rotational frequencies.

chapters/unitexpressions.tex
```diff
@@ -104,3 +104,68 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+\section{Unitful Literals}
+
+The \lstinline[language=grammar]!numeric! in the grammar allows a unit to be attached to a numeric value.
+In this case, the unit is given as a \lstinline[language=grammar]!Q-IDENT! rather than a string, since the use of a string could be ambiguously interpreted as \lstinline[language=grammar]!description-string! in some contexts.
+
+\begin{example}
+An acceleration of 9.8~meters per square second can be expressed as \lstinline!9.8'm/s2'!.
+\end{example}
+
+A unitful literal may only be used where unit consistency can be verified.
```
HansOlsson:
```suggestion
A unitful literal may only be used where unit consistency, including \lstinline!absoluteValue! where applicable, can be verified.
```
Even with the other changes it still wasn't clear that `absoluteValue` was part of the unit consistency.

chapters/unitexpressions.tex
HansOlsson: Or more precisely to give strong warnings for:
```
Real torque(unit="N.m") = 12 'J';
Real work(unit="J")= 20 'N.m';
```
And rotational frequencies are still unresolved.

HansOlsson:
I would also recommend that other tools check this @casella @MarkusOlssonModelon

My experience is also that it takes order(s) of magnitude longer to implement than #3668 - but after the remaining few issues are corrected it would be possible to add it. Given that it should have been part of the unit MCP (or a separate one) here are some test-cases:

[UnitfulTestCases.zip](https://github.com/user-attachments/files/23238836/UnitfulTestCases.zip)

The corresponding MSL with it used in a few places is https://github.com/HansOlsson/Modelica/tree/UnitTest2

Note that the test-cases also contain conversions; I don't see any reason to delay that; the time to implement conversion is basically insignificant compared to the time to implement the unit consistency.

chapters/unitexpressions.tex
henrikt-ma: I am trying to not say too much and not too little in order to make progress for all the units where `absoluteValue` doesn't matter, without opening the Pandora's box of `absoluteValue` reasoning.

Does the _wasn't_ in your comment mean that this PR is now acceptably clear regarding `absoluteValue`?

chapters/unitexpressions.tex
henrikt-ma: > Or more precisely to give strong warnings for:
> 
> ```
> Real torque(unit="N.m") = 12 'J';
> Real work(unit="J")= 20 'N.m';
> ```

I would expect the above to equally valid as when using a helper variable:
```
constant Real(unit = "J") x = 12;
Real torque(unit="N.m") = x;
constant Real y(unit = "N.m") = 20;
Real work(unit="J")= 20 y;
```

What sort of warnings that should be given for the above is a topic for unit checking, not a topic for this PR.

> And rotational frequencies are still unresolved.

Just as with the `"J"` vs `"N.m"`, one can run into the same problems when using helper variables instead of unitful literals, so I don't see why we would need to solve this messy problem as part of this PR.

HansOlsson:
Simplified variant for 3.7

henrikt-ma:
The PR is now updated according to the decisions at today's language group meeting.

chapters/unitexpressions.tex
henrikt-ma: Let's start over after all the discussions at the language group meeting.  Resolving.

chapters/unitexpressions.tex
```diff
@@ -131,44 +131,59 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
-// Requires unit arithmetic to verify consistency:
-parameter Real z(unit = "m") = 2'm/s' * 2's';
+// Implicit unit conversion resolves inconsistent units:
+Real x(unit = "m") = -5'cm';        // OK, same as -0.05'm'
+// No implicit unit conversion since not (signed) literal:
+Real y(unit = "m") = 0'cm' - 5'cm'; // Unit conflict
+// No implicit unit conversion since not (signed) literal:
+Real z(unit = "cm") = x;            // Unit conflict
+// No implicit unit conversion since w has no unit-attribute:
+Real w = 5'cm';                     // OK
```
gkurzbach: > `Real w = 5'cm';                     // OK`

change to 
`Real w = 5'cm';                     // OK, w = 5`

qlambert-pro (approved):
lgtm, as long as everybody understands that the "Unit conflicts" are not expected to be reported yet.

MarkusOlssonModelon (approved):

chapters/unitexpressions.tex
```diff
@@ -127,3 +127,96 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+\end{lstlisting}
+
+There is no implicit unit conversion in equations:
+\begin{lstlisting}[language=modelica]
+  Real x(unit = "m");
+equation
+  x = -5'cm'    // Unit conflict
+  x = -0.05'm'  // OK
+\end{lstlisting}
```
HansOlsson: This was not what I thought we decided. We must have have misunderstood each other.
My impression was that these cases would be illegal, and only the case above would be handled.

I know that is a lot more restrictive, so that it completely postponed the problematic cases, instead of introducing them now and adding their semantic checks later.

henrikt-ma: That sounds like a tremendous misunderstanding.  Without the ability to use unitful literals with consistent units in equations (just like when introducing a helper constant with the same unit), we can't start fixing unit issues in libraries.

The only problematic part about unitful literals is the implicit conversion, which we have carefully restricted so that it is only allowed where it is completely obvious what to do.

henrikt-ma:
@HansOlsson this is getting urgent.

henrikt-ma:
@gkurzbach could you please confirm that the current state of this PR corresponds to what we agreed upon at the recent language meeting in Lund?

We desperately need this for the 3.7 release so that libraries can start getting trivial cases out of the way of future discussions of the unit consistency rules.

henrikt-ma:
@dzimmer and @casella, maybe you can recall when we discussed this in Lucerne last September?  Compared to what we talked about then, the current proposal is that unitful literals may also enjoy implicit unit conversion in binding equations where the target unit is available directly in the `unit`-attribute.  With this information, would you be able to help us wrap this up for the upcoming language release?  (It is a small amount of normative text, and plenty of examples.)

HansOlsson:
> @HansOlsson this is getting urgent.

Unfortunately, I find that it such a major misunderstanding that we need more time.

The problem was always that adding unitful literals in equations in general with so many details to be added later would cause major problems in the future.

Restricting unitful literals to bindings (with conversions) provides a (lesser) real benefit without any downsides, and gives a clear path forward; and that's what I thought we agreed to as a compromise that would work for this release.

HansOlsson:
> @dzimmer and @casella, maybe you can recall when we discussed this in Lucerne last September? Compared to what we talked about then, the current proposal is that unitful literals may also enjoy implicit unit conversion in binding equations where the target unit is available directly in the `unit`-attribute. With this information, would you be able to help us wrap this up for the upcoming language release? (It is a small amount of normative text, and plenty of examples.)

The most important is that both variants give the clear future direction that unitful literals will use implicit unit conversion.

The other alternative I thought we agreed on was:

Unitful literals enjoy implicit unit conversion in binding equations where the target unit is available directly in the `unit`-attribute, and they can only be used in such places.

They thus have well-defined user-friendly semantics (you write `1'mV'` and it works the same as `1e-3'V'`) in all places where they can be used; without relying on yet to be specified unit inference (that we will work on in the future). It also seems fairly straightforward to add for tools that currently don't check and infer units, with low risk of mistakes.

henrikt-ma:
> Unfortunately, I find that it such a major misunderstanding that we need more time.

Then 3.7 should wait.

> The problem was always that adding unitful literals in equations in general with so many details to be added later would cause major problems in the future.

It seems clear to everyone else that it will not.  What we are proposing now is something where a unitful literal in an equation will play the same numeric role in the equation as a literal without unit, but tools with unit checking will see that the equation is unit consistent.  If a literal with the wrong unit is used in an equation, tools with unit checking will be able to reject it, while tools without unit checking that simply strip the units of all unitful literal in equations will still get the same behavior as before when there was no unit on the literal.

There is nothing to be afraid of.

henrikt-ma:
> They thus have well-defined user-friendly semantics (you write `1'mV'` and it works the same as `1e-3'V'`) in all places where they can be used; without relying on yet to be specified unit inference (that we will work on in the future). It also seems fairly straightforward to add for tools that currently don't check and infer units, with low risk of mistakes.

But this is not helping us solve a single one of the unit inconsistencies we have encountered in libraries.  What we need is unitful literals in equations, and there is no urgent need for implicit conversion there.  The implicit conversion in declaration equations was only added to this PR to make you happy and to align better with the current behavior in SimulationX.

HansOlsson:
My concern is, as always, that end-users want to actually use the feature (not just advanced developers of MSL) - and if you ask them what `1'mm'` means they normally expect it to be the same as `1e-3'm'`, and that it just works.
So the issue isn't whether  `1'mm'` is _needed_ in equations, it's whether users will _write_ it and what happens.

I'm aware that one can introduce variables that do the same currently, but unitful literals makes it more convenient - and we should ensure that it provides a good convenience - not a convenient foot-gun.

If we really want unitful literals outside of bindings with units we should strongly recommend that outside of bindings with units only use them with base-SI-units and only in equations in base-SI-units (even if equations could be in other units); as that is what is needed in MSL (and hopefully in most other libraries).

henrikt-ma:
> If we really want unitful literals outside of bindings with units we should strongly recommend that outside of bindings with units only use them with base-SI-units and only in equations in base-SI-units (even if equations could be in other units); as that is what is needed in MSL (and hopefully in most other libraries).

Then I'll add this sort of recommendation to get you on board.

henrikt-ma:
@HansOlsson, how about the added text?

chapters/unitexpressions.tex
```diff
@@ -127,3 +127,104 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+Implicit unit conversion in binding equations:
+\begin{lstlisting}[language=modelica]
+// Implicit unit conversion resolves inconsistent units:
+Real x(unit = "m") = -5'cm';        // OK, same as -0.05'm'
+// No implicit unit conversion since not (signed) literal:
+Real y(unit = "m") = 0'cm' - 5'cm'; // Unit conflict
+// No implicit unit conversion since not (signed) literal:
+Real z(unit = "cm") = x;            // Unit conflict
+// No implicit unit conversion since w has no unit-attribute:
+Real w = 5'cm';                     // OK
```
HansOlsson: I would write something like:
```suggestion
Real w = 5'cm';                     // Not recommended, but accepted
```

chapters/unitexpressions.tex
```diff
@@ -127,3 +127,104 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+
+\begin{example}
+An acceleration of 9.8~meters per square second can be expressed as \lstinline!9.8'm/s2'!.
+\end{example}
+
+When the modified component of a binding equation has a \lstinline!unit!-attribute, a binding equation consisting of a unitful literal or a negated unitful literal is implicitly converted to the component's unit.
+
+\begin{nonnormative}
+The predominant use of unscaled SI units in \lstinline!unit!-attributes in libraries (as this is a good way to avoid unit inconsistency without tool assistance to verify unit consistency) should be kept in mind when using unitful literals where there is no implicit unit conversion.
+That is, using units which are not equivalent to unscaled SI units for those unitful literals is likely to be a mistake.
```
HansOlsson: We should use the traditional terminology from BIPM, especially as scaling is a bit odd for mass.

Something like:
```suggestion
The predominant use of coherent SI units in \lstinline!unit!-attributes in libraries (as this is a good way to avoid unit inconsistency without tool assistance to verify unit consistency) should be kept in mind when using unitful literals where there is no implicit unit conversion.
Coherent SI units are SI base units and coherent derived units.
Coherent indicate that they are without any scaling (e.g., including newton meter, but excluding centimeter), and ensures that the equations between the numerical values of quantities take exactly the same form as the equations between the quantities themselves.
Note that the SI base unit for mass is kilogram.
That is, using units which are not coherent units for those unitful literals is likely to be a mistake.
```
Part of a sentence is based on the BIPM quote:
> The word “coherent” here means that equations between the numerical values of quantities take exactly the same form as the equations between the quantities themselves.

chapters/unitexpressions.tex
henrikt-ma: Not a big deal.  Applying.

chapters/unitexpressions.tex
```diff
@@ -127,3 +127,107 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+\begin{example}
+An acceleration of 9.8~meters per square second can be expressed as \lstinline!9.8'm/s2'!.
+\end{example}
+
+When the modified component of a binding equation has a \lstinline!unit!-attribute, a binding equation consisting of a unitful literal or a negated unitful literal is implicitly converted to the component's unit.
+
+\begin{nonnormative}
+The predominant use of coherent SI units in \lstinline!unit!-attributes in libraries (as this is a good way to avoid unit inconsistency without tool assistance to verify unit consistency) should be kept in mind when using unitful literals where there is no implicit unit conversion.
+Coherent SI units are SI base units and coherent derived units.
+Coherent indicate that they are without any scaling (e.g., including newton meter, but excluding centimeter), and ensures that the equations between the numerical values of quantities take exactly the same form as the equations between the quantities themselves.
```
henrikt-ma: I would formulate it a bit differently to not make it sound as giving the definition of _coherent_:
```suggestion
A coherent system of units ensures that the equations between the numerical values of quantities take exactly the same form as the equations between the quantities themselves.
In particular, scaled SI units are not part of the coherent system, since, e.g., the product of centimeter and newton is not equivalent to joule.
```

chapters/unitexpressions.tex
```diff
@@ -127,3 +127,107 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+In this case, the unit is given as a \lexicalunitref{Q-IDENT} rather than a string, since the use of a string could be ambiguously interpreted as \productionref{description-string} in some contexts.
+
+\begin{example}
+An acceleration of 9.8~meters per square second can be expressed as \lstinline!9.8'm/s2'!.
+\end{example}
+
+When the modified component of a binding equation has a \lstinline!unit!-attribute, a binding equation consisting of a unitful literal or a negated unitful literal is implicitly converted to the component's unit.
+
+\begin{nonnormative}
+The predominant use of coherent SI units in \lstinline!unit!-attributes in libraries (as this is a good way to avoid unit inconsistency without tool assistance to verify unit consistency) should be kept in mind when using unitful literals where there is no implicit unit conversion.
```
henrikt-ma, line 153: We can even be kind and give the definition before explaining the consequences:

```suggestion
The predominant use of coherent SI units in \lstinline!unit!-attributes in libraries (as this is a good way to avoid unit inconsistency without tool assistance to verify unit consistency) should be kept in mind when using unitful literals where there is no implicit unit conversion.
A unit system is \emph{coherent} when every derived unit can be built from the base units purely by multiplying and dividing powers of them, with no numerical factor other than 1 anywhere in the construction.
```

chapters/unitexpressions.tex
henrikt-ma: With the definition given above, it is better two swap the order:
```suggestion
For example, scaled SI units such as centimeter are not part of the coherent system, since they are derived units where the scaling relative to the base units is not 1.
A coherent system of units ensures that the equations between the numerical values of quantities take exactly the same form as the equations between the quantities themselves.
```

chapters/unitexpressions.tex
```diff
@@ -127,3 +127,107 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+Coherent SI units are SI base units and coherent derived units.
+Coherent indicate that they are without any scaling (e.g., including newton meter, but excluding centimeter), and ensures that the equations between the numerical values of quantities take exactly the same form as the equations between the quantities themselves.
+Note that the SI base unit for mass is kilogram.
+That is, using units which are not coherent units for those unitful literals is likely to be a mistake.
+The mistake can be detected by tools that perform (tool-dependent) analysis of unit consistency, while tools that don't analyze unit consistency are likely to just drop the unit and produce unexpected results, see example below.
```
HansOlsson:
```suggestion
The mistake can be detected by tools that perform (tool-dependent) analysis of unit consistency, while tools that don't analyze unit consistency are likely to just drop the unit and produce incorrect results, see example below.
```
I find that "unexpected" is a bit vague, and "incorrect" more clearly captures the issue.

chapters/unitexpressions.tex
HansOlsson: Looks good.

chapters/unitexpressions.tex
henrikt-ma: Apply this as well, now that we have a clear definition right before this?

henrikt-ma: I take that for a yes.

henrikt-ma:
After many iterations there is now (in my opinion) a very clear explanation of why the use of non-coherent units in unitful literals without implicit conversion is likely to be a mistake.  However, the recommendation is not very concise since the clarity requires some terminology to be introduced before we get to the main paragraph at the end.

I considered moving the first two paragraphs to section 1.3 _Some Definitions_, but it felt wrong to move definitions so far away when they are only relevant for this non-normative part of 19.2.  I also considered having a subsection with definitions somewhere in chapter 19, but I couldn't think of a natural place to insert it, and it seemed like something non-normative growing out of proportion.

So, in the end I have no better idea than to keep the definitions as part of the non-normative section.

Any final thoughts?

chapters/unitexpressions.tex
```diff
@@ -127,3 +127,113 @@ \section{The Syntax of Unit Expressions}\label{the-syntax-of-unit-expressions}
...
+\begin{lstlisting}[language=modelica]
+  parameter Real x(unit = "m", displayUnit = "cm") = 0.1;
+equation
+  y = x + 2; /* Intention: Add 2 centimeters to x. */
+\end{lstlisting}
+The result is that 2~meters instead of 2~centimeters is being added to \lstinline!x!.
+
+Using a unitful literal makes the intention clear:
+\begin{lstlisting}[language=modelica]
+  y = x + 2'cm';
```
HansOlsson:
```suggestion
  y = x + 2'cm'; // Detectable error
```
We normally have comments on the lines of each incorrect line (even if it looks redundant), and don't rely on text succeeding or preceding it.

I'm not sure if `y = x + 2; /* Intention: Add 2 centimeters to x. */` could be clarified as well (it is a non-detectable error).

HansOlsson:
> After many iterations there is now (in my opinion) a very clear explanation of why the use of non-coherent units in unitful literals without implicit conversion is likely to be a mistake. However, the recommendation is not very concise since the clarity requires some terminology to be introduced before we get to the main paragraph at the end.

I really hope this recommendation is strong enough for users.

chapters/unitexpressions.tex
henrikt-ma: I addressed this with the standard comment `Unit conflict` instead.  Please resolve if OK.

HansOlsson: Ok

HansOlsson (approved):
Ok.

PR #3718, opened 2025-08-13
Apply suggested workaround for LaTeXML issue

henrikt-ma:
This workaround is a temporary solution while waiting LaTeXML 0.8.9 or newer, so an alternative to applying the workaround would be to upgrade.


HansOlsson (approved):
Looks good.
However, it might be that LaTeXML 0.8.9 is close to being released based on previous releases.

henrikt-ma:
Sure, but when they release is one thing, and when we upgrade could be another.

PR #3724, opened 2025-08-27
Constrain use of 'delay' with zero delay time

henrikt-ma:
This is a new take on #3245, picking up where it was left with the comment  https://github.com/modelica/ModelicaSpecification/issues/3245#issuecomment-1366255020.  (I am revisiting this now because of my ongoing test implementation of `delay` with event generation.)

I think this solves the issue with an acceptable tradeoff between backwards compatibility and feasibility of actually implementing according to specification:
- As @MarkusOlssonModelon seems to have suggested in [phone meeting](https://github.com/modelica/ModelicaSpecification/issues/3287#issuecomment-1351574969), `delayTime = 0` is forbidden for time-varying `delayTime`.
- In the two argument form `delay(u, delayTime)`, the "new" evaluable variability is used to constrain `delayTime`, making it possible for tools to determine at translation time whether `delayTime` has been set to zero, but not forcing tools to evaluate.
- By allowing `delayTime = 0` in the two argument form, we avoid the need to give complicated rules for when it is an error to have `delayTime = 0` in guarded expressions such as `if tau > 0 then delay(u, tau) else u`.
- In principle, the funky technique of only evaluating `delayTime` when the value would be zero can be avoided by generating different equation systems for the two cases, and make an initialization time decision of which case to use.  (Not expecting any tool to actually do it this way.)

### DIFF
--- a/chapters/operatorsandexpressions.tex
+++ b/chapters/operatorsandexpressions.tex
@@ -945,7 +945,8 @@ For variable step size integrators, the buffer size is dynamic during integratio
 
 In order to avoid extrapolation in the delay buffer, the maximum step size of the integrator is limited by the delay time, making delay times close to zero a potential source of poor simulation performance.
 
-In the form where $\mathit{delayTime}$ is required to be evaluable, a tool can decide to evaluate $\mathit{delayTime}$ only when the value would be zero, and introduce a runtime check that it remains non-zero otherwise.
+It is a quality of implementation to avoid evaluating $\mathit{delayTime}$ when possible.
+A pragmatic approach is to only evaluate $\mathit{delayTime}$ in case the value would be zero, and to check at runtime that it is positive in case it was not evaluated.
 \end{nonnormative}
 
 \subsubsection{spatialDistribution}\label{spatialdistribution}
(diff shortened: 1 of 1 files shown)


HansOlsson (commented):
Hans:
* Should add QoI for run-time check.
* Would prefer that if parameter delayTime is non-evaluable it is still valid, but does not support delayTime=0.

HansOlsson:
Language meeting:
Seems an improvement to current work-around: `if dt>0 then delay(x, dt) else x`.

Has an implication for whether delay breaks algebraic loops; so decision must be made early (slightly complicated to not evaluate it if non-zero).

(Some people strongly want delay to be able to handle delay-time=0.)

HansOlsson:
Language group: the request is currently unclear - waiting for more information.

HansOlsson:
Will require detailed investigation.

HansOlsson:
Language meeting:
Conclusion is that we don't want this (especially not for 3.7 as it involves complicated tentative evaluation)  - the plan is:
* Close this PR.
* Create new PR for restricting delayTime to be >0, and merge for 3.7 (subject to review)
* Introduce a boolean parameter for by passing the delay-part in relevant delay-blocks (in MSL and other libraries).

PR #3756, opened 2025-10-02
Fix grammar and hyperref, remove unused defines

phannebohm:
I suppose those `\def\S` are obsolete, since they are no longer used anywhere.

### DIFF
--- a/preamble.tex
+++ b/preamble.tex
@@ -191,13 +191,6 @@
 % Like \annotationindex, but also outputting the formatted name of the annotation.
 \newcommand{\fmtannotationindex}[1]{\lstinline{#1}\annotationindex{#1}}
 
-\def\S0#1#2{\hypertarget{#2}\chapter{#1}\label{#2}}
-\def\S1#1#2{\hypertarget{#2}\section{#1}\label{#2}}
-\def\S2#1#2{\hypertarget{#2}\subsection{#1}\label{#2}}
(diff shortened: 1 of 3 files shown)


preamble.tex
```diff
@@ -191,13 +191,6 @@
 % Like \annotationindex, but also outputting the formatted name of the annotation.
 \newcommand{\fmtannotationindex}[1]{\lstinline{#1}\annotationindex{#1}}
 
-\def\S0#1#2{\hypertarget{#2}\chapter{#1}\label{#2}}
-\def\S1#1#2{\hypertarget{#2}\section{#1}\label{#2}}
-\def\S2#1#2{\hypertarget{#2}\subsection{#1}\label{#2}}
-\def\S3#1#2{\hypertarget{#2}\subsubsection{#1}\label{#2}}
-\def\S4#1#2{\hypertarget{#2}\paragraph{#1}\label{#2}}
-\def\S5#1#2{\hypertarget{#2}\subparagraph{#1}\label{#2}}
-
```
henrikt-ma, line 200: Are you sure these are not redefining the behavior of macros being used by packages outside the sources in our repository?

henrikt-ma: I checked it myself by turning all of them into `\ref{hello-hello-hello}`, and it didn't lead to any reports of undefined references, so I suppose they are OK to remove.

henrikt-ma (approved):
All changes look good to me, although we usually try to avoid mixing diverse changes like this in the same PR.

PR #3760, opened 2025-10-15
Normatively describe trade-off at delay times close to zero

### LINKED ISSUE #3726: Is delayTime approaching zero really that problematic?
The [specification](https://specification.modelica.org/master/operators-and-expressions.html#delay) non-normatively explains:
> Furthermore, the maximum step size of the integrator is limited by this minimum delay time in order to avoid extrapolation in the delay buffer.

My interpretation is that the following model is expected to get stuck at time π/2:
```
model DelayTimeApproachingZero
  Real x = time;
  Real w(unit = "rad/s") = 1.0;
  Real delayTime(unit = "s") = cos(w * time)^4 + Modelica.Constants.small;
  Real pastTime = time - delayTime;
  Real y = delay(x, delayTime, 2.0);
  annotation(experiment(StopTime = 6.3));
end DelayTimeApproachingZero;
```

But would it really be so bad to make it through periods of very small `delayTime` by doing a tiny bit of extrapolation between the most recent accepted step and the current intermediate step of the integration method?

I would be interested in hearing how different tools deal with the model above.

henrikt-ma:
Fixes #3726.

As a nice side-effect, I think that what remains of the `delay` section makes more sense as a separate section when it can be presented as just containing implementation notes.


HansOlsson:
@MarkusOlssonModelon : Intended to be interpolation, not extrapolation.

Not for breaking loops; even though users may think it is made for that.

PR #3765, opened 2025-10-20
Use consistent language when talking about scheduling clock ticks

henrikt-ma:
This is a follow-up to #3754.


HansOlsson (approved):
Seems clearer.
"Triggered at" very much seems as if the event happens at that time, not that time into the future.

PR #3770, opened 2025-10-27
Generalize modern table presentation with application to Protection

henrikt-ma:
This PR aims to make the presentation of the `Protection` annotation more similar to the presentation of other annotations, even though `Protection` only acts as a container for sub-annotations which are defined in more detail one by one.  If we can work this out for `Protection`, it should be straightforward to treat `Documentation` in the same manner.

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -1096,12 +1096,11 @@ The default for \lstinline!iconTransformation! is to be the same as \lstinline!t
 \end{semantics}
 \end{annotationdefinition}
 
-%\begin{nonnormative}
-%A connector can be shown in both an icon layer and a diagram layer of a class.
-%Since the coordinate systems typically are different, placement information needs to be given using two different coordinate systems.
-%More flexibility than just using scaling and translation is needed since the abstraction views might need different visual placement of the connectors.
-%The attribute \lstinline!transformation! gives the placement in the diagram layer and \lstinline!iconTransformation! gives the placement in the icon layer.
-%\end{nonnormative}
+\begin{nonnormative}
+While it is possible to have independent placements of a connector in the icon and diagram layers, this freedom should be used with care.
+A user who wants to find out what hides behind the icon of a component by opening up the component's diagram view will benefit if the relative positioning of connectors are similar between icon and diagram.
+This way, intuitive understanding of the diagram layer is enabled without the need to pay attention to connector names.
+\end{nonnormative}
 
 The \lstinline!transformation! and \lstinline!iconTransformation! of the \lstinline!Placement! annotation have the following type:
 \begin{lstlisting}[language=modelica]


chapters/annotations.tex
```diff
@@ -2366,29 +2380,39 @@ \section{Access Control to Protect Intellectual Property}\label{annotations-for-
...
-record Protection
-  /*literal*/ constant Access access;
-  $\ldots$
-end Protection;
-type Access = enumeration(hide, icon, documentation, diagram,
+This section describes the part of the \lstinline!Protection! annotation used to define protection.
+
+\begin{annotationdefinition*}[access]\label{modelica:Protection.access}\annotationindex{access}\index{Protection@\robustinline{Protection}!\robustinline{access} sub-annotation}
+\begin{synopsis}\begin{lstlisting}
+/*literal*/ constant Access Protection.access;
```
```suggestion
/*literal*/ constant Access access;
```
Making it consistent with the other ones.

henrikt-ma: You are right that the PR is currently inconsistent and that this needs to be fixed.

However, I am not sure it's such a good idea to present the sub-annotations in the same way as the annotations residing directly under `annotation(…)`.

HansOlsson: Well, many sub-annotations are defined as part of a record https://specification.modelica.org/master/annotations.html#annotations-for-documentation https://specification.modelica.org/master/annotations.html#annotations-for-figures https://specification.modelica.org/master/annotations.html#modelica:Dialog

So, the alternative would be to have "record Protection" preceding this declaration as in the current specification, but it would seem odd with just one element, and normally the entry would then be called "Protection".

henrikt-ma: I see some similarities with those examples, but as you already noted they are not quite the same:
- The `Documentation` is not a sub-annotation, so it could be converted to the new style if we really wanted to.  However, there is not that much to say about `Documentation` as a whole, so it makes sense to me that we don't present it in the same way as an annotation which we are able to cover in more detail without only referring to other sections.
- The `Figure` is a type and not the name of an annotation, so it looks correct to me that we don't present it as an annotation by itself, and that we don't include it in any summary table.  However, if there was a way to present `Documentation.figures` as a sub-annotation, it would have been nice to do so at the top of https://specification.modelica.org/master/annotations.html#annotations-for-figures, something like this:
```
Figure[:] Documentation.figures;

record Figure
  /*literal*/ constant String title = "" "Title meant for display";
  /*literal*/ constant String identifier = "" "Identifier meant for programmatic access";
  /*literal*/ constant String group = "" "Name of figure group";
  /*literal*/ constant Boolean preferred = false "Automatically display figure after simulation";
  Plot[:] plots "Plots";
  /*literal*/ constant String caption = "" "Figure caption";
end Figure;
```
- I find the `Dialog` properly presented already.  With `loadSelector` and friends being described together with the rest of `Dialog`, I don't see a need to change anything regarding them.  (Having separate annotation definitions for `Dialog.loadSelector` etc would just look odd to me since they are such an integral part of `Dialog`.)

HansOlsson: Well, one thing to consider is that not only is it different, it also breaks the idea that the annotations are described by legal Modelica code. 

E.g., we use `/*literal*/` to avoid having a special keyword, whereas `Protection.access` is not a valid name for a component - and the most obvious way of handling it: `'Protection.access'` would just create more confusion.

So, having `record Protection` in front of may seem a bit excessive, but at least we have _some_ precedence and it is legal Modelica.

henrikt-ma: The problems I see with `record Protection` are:
- It no longer fits nicely on a line.
- We need to indicate that this is just one out of several members of the record.

I mean, this becomes a very cumbersome way to say that the `access` member a `Protection` is a `/*literal*/ constant Access`:
```
record Protection
  /*literal*/ constant Access access;
  $\ldots$
end Protection;
```

While I would still much prefer the non-valid Modelica `Protection.access` notation, one could also consider some sort of class modification notation:
```
Protection(/*literal*/ constant Access access);
```

HansOlsson: Well, the text already has a link to the rest of the contents of Protection, and master already has the
```
record Protection
  /*literal*/ constant Access access;
  $\ldots$
end Protection;
```
So, why not keep it for now?

HansOlsson: Various variants were considered, including having Protection in the text and having the full Protection-annotation followed by simplified form.

Open a new issue for resolving this.

henrikt-ma: I have now changed the presentation according to https://github.com/modelica/ModelicaSpecification/pull/3770#discussion_r2585258915, and also applied the same style to `Documentation`.

chapters/annotations.tex
```diff
@@ -976,9 +1060,54 @@ \subsubsection{Graphical Properties}\label{graphical-properties}
...
+\end{semantics}
+\end{annotationdefinition}
+
+%\begin{nonnormative}
+%A connector can be shown in both an icon layer and a diagram layer of a class.
+%Since the coordinate systems typically are different, placement information needs to be given using two different coordinate systems.
+%More flexibility than just using scaling and translation is needed since the abstraction views might need different visual placement of the connectors.
+%The attribute \lstinline!transformation! gives the placement in the diagram layer and \lstinline!iconTransformation! gives the placement in the icon layer.
+%\end{nonnormative}
+
```
HansOlsson: Why was this explanation commented out?
I can understand that the explanation may seem a bit out of place in the specification, but just commenting it out seems a bit harsh.

henrikt-ma: Good question. The reason was that I found the reorganized normative text so clear that the non-normative paragraph seemed completely redundant.  Any specific part of it that you miss particularly?

Edit: I have now improved the structure of `\cref{component-instance}` further, avoiding giving normative rules right below the section heading instead of in the annotation definition proper.  The two roles of `Transformation` should also be so clear in the annotation definition now, that what goes below the annotation definition only needs to give the rules for how the `Translation` attributes are applied.  With these improvements, I find it even harder to see the purpose of the non-normative paragraph, or where it would fit in.

HansOlsson: To me the purpose of that paragraph is to explain why you can give an iconTransformation. The new text still doesn't say you might need to place the icon differently in the two views due to possibly different coordinate systems.

However, looking at the use in MSL I realized that:
- It is often used when it isn't needed (just duplicating the transformation, or possibly more complicated with the same result).
- When it is used "correctly" as in `Modelica.Mechanics.Rotational.Components.AngleToTorqueAdaptor` it is still "similar".

So, additionally, 
The iconTransformation should be used with care, since it is expected that the connectors are placed the similarly in the two views.

henrikt-ma: > So, additionally,
> The iconTransformation should be used with care, since it is expected that the connectors are placed the similarly in the two views.

OK.  For me this works better as an excuse for the non-normative paragraph.  I tried writing something based on this.  Better?

chapters/annotations.tex
HansOlsson: Ok.

chapters/annotations.tex
HansOlsson: Yes.

PR #3780, opened 2025-11-07
List changes

HansOlsson:
Starting the process of listing semantic changes.

### DIFF
--- a/RationaleMCP/Changes/3_7/3610.md
+++ b/RationaleMCP/Changes/3_7/3610.md
@@ -14,7 +14,7 @@ Making it possible to create functions that generate events similarly as the bui
 function bar
   input Real x;
   output Integer i = integer(x);
-  annotation(GenerateEvents=false);
+  annotation(GenerateEvents=false); // Redundant, but added for clarity
 end bar;
 
 function foo
@@ -27,3 +27,4 @@ Real x=foo(time);
 ```
 
 The call of `bar` in `foo` is now illegal, similarly as in a model.
+Note that in practice `foo` might contain both something generating event and the call to `bar`.


henrikt-ma:
How about naming the directory so that one can actually see that separation of major and minor version?  Like `3_7` instead of `37`?

RationaleMCP/Changes/Readme.md
```diff
@@ -0,0 +1,8 @@
+This list the semantic changes in the various versions.
```
henrikt-ma: Wording:
```suggestion
This directory contains descriptions of semantic changes introduced in the various versions.
```

RationaleMCP/Changes/Readme.md
```diff
@@ -0,0 +1,8 @@
...
+
+Clarifications, examples, reformulations, and changes in layout or other external resources will not be included in this list.
+
+
+
```
henrikt-ma: Redundant:

```suggestion
```

RationaleMCP/Changes/Readme.md
```diff
@@ -0,0 +1,8 @@
...
+The are separate files for each release to make it easy to maintain during development, but may later be merged.
+
```
henrikt-ma, line 4: Redundant:

```suggestion
```

RationaleMCP/Changes/Readme.md
```diff
@@ -0,0 +1,8 @@
+This list the semantic changes in the various versions.
+
+Clarifications, examples, reformulations, and changes in layout or other external resources will not be included in this list.
+
+
+
+The are separate files for each release to make it easy to maintain during development, but may later be merged.
```
henrikt-ma: Typo and clarification:

```suggestion
There are separate files for each release to make it easy to maintain during development, but the files may later be merged.
```

RationaleMCP/Changes/Readme.md
```diff
@@ -0,0 +1,8 @@
+This list the semantic changes in the various versions.
+
+Clarifications, examples, reformulations, and changes in layout or other external resources will not be included in this list.
```
henrikt-ma: I thought the plan was to only describe backwards incompatible changes, that is, to not add descriptions of backwards compatible new features?

RationaleMCP/Changes/3_7/3610.md
```diff
@@ -0,0 +1,28 @@
+Discrete-time variability based on GenerateEvents #3610
+
+# Main change
+
+Variability of functions with GenerateEvents.
```
henrikt-ma: Formatting:
```suggestion
Variability of functions with `GenerateEvents`.
```

RationaleMCP/Changes/3_7/3610.md
```diff
@@ -0,0 +1,28 @@
+Discrete-time variability based on GenerateEvents #3610
```
henrikt-ma: Formatting:
```suggestion
# Discrete-time variability based on `GenerateEvents` #3610
```

RationaleMCP/Changes/3_7/3610.md
```diff
@@ -0,0 +1,28 @@
...
+
+# Main change
```
henrikt-ma: Push down:
```suggestion
## Main change
```

RationaleMCP/Changes/3_7/3610.md
```diff
@@ -0,0 +1,28 @@
...
+
+Variability of functions with GenerateEvents.
+
+# Reason
```
henrikt-ma: Same as above:
```suggestion
## Reason
```

RationaleMCP/Changes/3_7/3610.md
```diff
@@ -0,0 +1,28 @@
...
+
+Making it possible to create functions that generate events similarly as the built-in functions, and have rules consistent with them just being inlined.
+
+# Breaks
```
henrikt-ma: Same as above:
```suggestion
## Breaks
```

RationaleMCP/Changes/Readme.md
HansOlsson: New features should be described _separately_; but I think it would still be good to include them somehow. But let's first focus on the backwards incompatible ones.

henrikt-ma: So if that is the focus for now, then I think this is what line 1 should say.

RationaleMCP/Changes/Readme.md
```diff
@@ -0,0 +1,5 @@
+This directory contains descriptions of semantic changes introduced in the various versions.
+
+Clarifications, examples, reformulations, and changes in layout or other external resources will not be included in this list.
```
HansOlsson:
```suggestion
This directory contains descriptions of semantic changes, particularly backwards incompatible changes, introduced in the various versions.

Clarifications, new or modified examples, reformulations, and changes in layout or other external resources will not be included in this list.
```

RationaleMCP/Changes/3_7/3610.md
```diff
@@ -0,0 +1,28 @@
...
+  output Integer i=bar(x)+(if x>0 then 1 else 0);
+  annotation(GenerateEvents=true);
+end foo;
+
+function bar
+  input Real x;
+  output Integer i=integer(x);
+  annotation(GenerateEvents=false);
+end bar;
+Integer x=foo(time);
```
henrikt-ma: In order to be relevant, this line should be valid according to the old semantics:
```suggestion

Real x=foo(time);
```

RationaleMCP/Changes/3_7/3610.md
```diff
@@ -0,0 +1,28 @@
...
+  input Real x;
+  output Integer i=bar(x)+(if x>0 then 1 else 0);
+  annotation(GenerateEvents=true);
+end foo;
+
+function bar
+  input Real x;
+  output Integer i=integer(x);
+  annotation(GenerateEvents=false);
+end bar;
```
henrikt-ma: With the call to `foo` being last in the listing, I think the listing should consistently define functions before they are used:

Also, I am not sure whether the `(if x>0 then 1 else 0)` really helps pinpointing the problem…

```suggestion
function bar
  input Real x;
  output Integer i = integer(x);
  annotation(GenerateEvents=false);
end bar;

function foo
  input Real x;
  output Integer i = bar(x);
  annotation(GenerateEvents=true);
end foo;
```

HansOlsson: True, the `(if x>0 then 1 else 0)` was intended to make the example more realistic in some sense. (As the GenerateEvents=true was intended to generate events for that part, but realism might not be required here.)

RationaleMCP/Changes/3_7/3610.md
```diff
@@ -0,0 +1,29 @@
...
+
+Making it possible to create functions that generate events similarly as the built-in functions, and have rules consistent with them just being inlined.
+
+## Breaks
+
+``` modelica
+function bar
+  input Real x;
+  output Integer i = integer(x);
+  annotation(GenerateEvents=false);
```
henrikt-ma: Clarify that this is redundant, or remove:
```suggestion
```

HansOlsson: I believe it is clearer to keep it, but then with explanation.

henrikt-ma: OK.  Resolving.

PR #3786, opened 2025-11-20
Clarify

### LINKED ISSUE #3785: Partial derivative for several inputs
I can't make sense of the partial derivative definition in https://specification.modelica.org/master/functions.html#partial-derivatives-of-functions.

Based on the example, it seems clear that the output type must be the same (up to units) as for the differentiated function, since `T` is a scalar:
```
  h := Gibbs(p, T) - T * Gibbs_T(p, T);
```

Then, however, I see that the `der(f, …)` allows one or more input arguments to be listed, and then I would have expected that the result of `Gibbs_T` would be an array of the same size as the number of listed inputs, meaning that it should be used like this:
```
  h := Gibbs(p, T) - {T} * Gibbs_T(p, T);
```

Am I missing something?

HansOlsson:
Closes #3785
(I know that the mathematical notation could be clarified further, and that the partial derivatives that are different inputs commute, but I don't want to get too carried away.)

### DIFF
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -1474,16 +1474,21 @@ Therefore {\lstinline!h!} indirectly includes the derivative with respect to {\l
 
 \subsection{Partial Derivatives of Functions}\label{partial-derivatives-of-functions}\index{partial derivative}\index{der@\robustinline{der}!partial derivative}
 
-A class defined as:
+A function class defined as follows is a partial derivative of another function:
 \begin{lstlisting}[language=grammar]
 IDENT "=" der "(" name "," IDENT { "," IDENT } ")" comment
 \end{lstlisting}
-is the partial derivative of a function, and may only be used as declarations of functions.
 
-The semantics is that a function (and only a function) can be specified in this form, defining that it is the partial derivative of the function to the right of the equal sign (looked up in the same way as a short class definition, and the looked up name must be a function), and partially differentiated with respect to each {\lstinline!IDENT!} in order (starting from the first one).
-Each {\lstinline!IDENT!} must be a scalar {\lstinline!Real!} input to the function, and corresponds mathematically to prepending $\frac{\partial}{\partial\text{\lstinline!IDENT!}}$ to the function call.
+In \lstinline!$f$ = der($g$, $u_{1}, \ldots$)!, the function being defined is named $f$, and the function being differentiated is $g$.
+The name $g$ is looked up in the same way as a in short class definition, and the referenced class must be a function.
+Each $u_{i}$ must be a scalar {\lstinline!Real!} input to the function, and corresponds mathematically to prepending $\frac{\partial}{\partial u_{i}}$ to the function call.
+The $u_{i}$ are applied in increasing order of $i$ (although the partial derivatives commute for a broad class of functions).
 
-The comment allows a user to comment the function (in the info-layer and as one-line description, and as icon).
+\begin{nonnormative}
+In mathematical notation, the order of partial differentiation is reversed compared to the function definition; \lstinline!der($g$, $x$, $y$)! means $\frac{\partial}{\partial y}\frac{\partial}{\partial x}g$.
+\end{nonnormative}
+
+The \lstinline[language=grammar]!comment! has the same semantics as in a short class definition, for instance allowing the function to be given a description string, as well as \lstinline!Documentation! and \lstinline!Icon! annotations.
 
 \begin{example}
 The specific enthalpy can be computed from a Gibbs-function as follows:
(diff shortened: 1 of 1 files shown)


chapters/functions.tex
```diff
@@ -1481,7 +1481,7 @@ \subsection{Partial Derivatives of Functions}\label{partial-derivatives-of-funct
 is the partial derivative of a function, and may only be used as declarations of functions.
 
 The semantics is that a function (and only a function) can be specified in this form, defining that it is the partial derivative of the function to the right of the equal sign (looked up in the same way as a short class definition, and the looked up name must be a function), and partially differentiated with respect to each {\lstinline!IDENT!} in order (starting from the first one).
-Each {\lstinline!IDENT!} must be a scalar {\lstinline!Real!} input to the function.
+Each {\lstinline!IDENT!} must be a scalar {\lstinline!Real!} input to the function, and mathematically correspond to prepending $\frac{\partial}{\partial\text{\lstinline!IDENT!}}$ to the function call.
```
henrikt-ma: Avoid it sounding as it it must correspond:
```suggestion
Each {\lstinline!IDENT!} must be a scalar {\lstinline!Real!} input to the function, and corresponds mathematically to prepending $\frac{\partial}{\partial\text{\lstinline!IDENT!}}$ to the function call.
```

henrikt-ma: Please ignore this in favor of the complete proposal above.

chapters/functions.tex
```diff
@@ -1481,7 +1481,7 @@ \subsection{Partial Derivatives of Functions}\label{partial-derivatives-of-funct
 is the partial derivative of a function, and may only be used as declarations of functions.
```
henrikt-ma: I found it best to start from the beginning.  Here is a proposal for the entire normative part of the section:
```suggestion
A function class defined as follows is a partial derivative of another function:
\begin{lstlisting}[language=grammar]
IDENT "=" der "(" name "," IDENT { "," IDENT } ")" comment
\end{lstlisting}

In \lstinline!$f$ = der($g$, $u_{1}, \ldots$)!, the function being defined is named $f$, and the function being differentiated is $g$.
The name $g$ is looked up in the same way as a in short class definition, and the referenced class must be a function.
Each $u_{i}$ must be a scalar {\lstinline!Real!} input to the function, and corresponds mathematically to prepending $\frac{\partial}{\partial u_{i}}$ to the function call.
The $u_{i}$ are applied in increasing order of $i$ (although the partial derivatives commute for a broad class of functions).

\begin{nonnormative}
In mathematical notation, the order of partial differentiation is reversed compared to the function definition; \lstinline!der($g$, $x$, $y$)! means $\frac{\partial}{\partial y}\frac{\partial}{\partial x}g$.
\end{nonnormative}

The \lstinline[language=grammar]!comment! has the same semantics as in a short class definition, for instance allowing the function to be given a description string, as well as \lstinline!Documentation! and \lstinline!Icon! annotations.
```

chapters/functions.tex
```diff
@@ -1502,6 +1502,8 @@ \subsection{Partial Derivatives of Functions}\label{partial-derivatives-of-funct
   h := Gibbs(p, T) - T * Gibbs_T(p, T);
 end specificEnthalpy;
 \end{lstlisting}
+Thus \lstinline!der(Gibbs, T)! corresponds to $\frac{\partial \text{\lstinline!Gibbs!}}{\partial \text{\lstinline!T!}}=\frac{\partial}{\partial \text{\lstinline!T!}}\text{\lstinline!Gibbs!}$.
+For multiple arguments \lstinline!der(Gibbs, T, T, p)! corresponds to $\frac{\partial}{\partial \text{\lstinline!p!}}\frac{\partial}{\partial \text{\lstinline!T!}}\frac{\partial}{\partial \text{\lstinline!T!}}\text{\lstinline!Gibbs!}$.
```
henrikt-ma: With the more extensive normative test, I think this sentence can be dropped from the example:

```suggestion
```

chapters/functions.tex
```diff
@@ -1502,6 +1502,8 @@ \subsection{Partial Derivatives of Functions}\label{partial-derivatives-of-funct
   h := Gibbs(p, T) - T * Gibbs_T(p, T);
 end specificEnthalpy;
 \end{lstlisting}
+Thus \lstinline!der(Gibbs, T)! corresponds to $\frac{\partial \text{\lstinline!Gibbs!}}{\partial \text{\lstinline!T!}}=\frac{\partial}{\partial \text{\lstinline!T!}}\text{\lstinline!Gibbs!}$.
```
henrikt-ma: When `Gibbs` is in the numerator, shouldn't it appear with the formal parameters written out?

```suggestion
Thus \lstinline!der(Gibbs, T)! corresponds to $\frac{\partial \text{\lstinline!Gibbs(p, T)!}}{\partial \text{\lstinline!T!}} = \frac{\partial}{\partial \text{\lstinline!T!}}\text{\lstinline!Gibbs!}$.
```

HansOlsson: Sure, it is slightly clearer.

chapters/functions.tex
HansOlsson: Now done.

PR #3787, opened 2025-11-26
Add BOM to document index

henrikt-ma:
On behalf of @maltelenz, who I guess needed a reference for the deprecation to give to some library author.


HansOlsson (approved):
Ok

PR #3789, opened 2025-12-04
Clarify restrictions on fill patterns.

### LINKED ISSUE #1826: What is fillPattern=FillPattern.Sphere on Rectangle or Polygon?
**Modified by dietmarw on 9 Nov 2015 14:17 UTC**
How should a rectangle/polygon with fillPattern=FillPattern.Sphere be drawn? Currently tools behave different here.

See attached test model.

----

**Reported by dietmarw on 9 Nov 2015 14:09 UTC**
How should a rectangle/polygon with fillPattern=FillPattern.Spheree be drawn? Currently tools behave different here.

See attached test model.

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1826

HansOlsson:
Closes #1826

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -925,7 +925,7 @@ The \lstinline!FillPattern! attributes \lstinline!Horizontal!, \lstinline!Vertic
 
 The attributes \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder! and \lstinline!Sphere! specify gradients that represent a horizontal cylinder, a vertical cylinder and a sphere, respectively.
 The gradient goes from line color to fill color.
-For polygons the gradients are defined on the enclosing rectangle (for \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder!) or circle (for \lstinline!Sphere!).
+For polygons the gradients are defined on the enclosing rectangle (for \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder!) or ellipse (for \lstinline!Sphere!).
 For ellipses the gradients are defined such that the perimeter has line color.
 The gradients colors are defined considering full rectangles (ignoring rounded corners) and full ellipses (ignoring arcs).
 
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -925,7 +925,7 @@ The \lstinline!FillPattern! attributes \lstinline!Horizontal!, \lstinline!Vertic
 
 The attributes \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder! and \lstinline!Sphere! specify gradients that represent a horizontal cylinder, a vertical cylinder and a sphere, respectively.
 The gradient goes from line color to fill color.
-The \lstinline!HorizontalCylinder! and \lstinline!VerticalCylinder! are only defined for rectangles and ellipses (not polygons), and \lstinline!Sphere! is only defined for ellipses.
+For polygons the gradients are defined on the enclosing rectangle (for \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder!) or circle (for \lstinline!Sphere!).
 
 The border pattern attributes \lstinline!Raised!, \lstinline!Sunken! and \lstinline!Engraved! represent frames which are rendered in a tool-dependent way --- inside the extent of the filled shape.
 
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -926,7 +926,7 @@ The \lstinline!FillPattern! attributes \lstinline!Horizontal!, \lstinline!Vertic
 The attributes \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder! and \lstinline!Sphere! specify gradients that represent a horizontal cylinder, a vertical cylinder and a sphere, respectively.
 Gradients are defined for the geometry of a \lstinline!GraphicItem! before its \lstinline!rotation! is applied.
 The coloring is defined for an enclosing shape, and then clipped to the actual shape of the \lstinline!GraphicItem!.
-For \lstinline!HorizontalCylinder! and \lstinline!VerticalCylinder!, the minimal enclosing axis-parallel rectangle is used, while \lstinline!Sphere! uses the smallest enclosing ellipse.
+For \lstinline!HorizontalCylinder! and \lstinline!VerticalCylinder!, the minimal enclosing axis-parallel rectangle is used, while \lstinline!Sphere! uses the smallest area enclosing ellipse.
 The enclosing shapes are defined considering full rectangles (ignoring rounded corners) and full ellipses (ignoring arcs).
 For \lstinline!HorizontalCylinder!, the gradient attains the fill color along the horizontal line dividing the enclosing rectangle at the center, and the line color at the two horizontal sides of the rectangle.
 The \lstinline!VerticalCylinder! is defined analogously to \lstinline!HorizontalCylinder!, but using vertical lines.
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -924,10 +924,9 @@ The \lstinline!LinePattern! attribute \lstinline!Solid! indicates a normal line,
 The \lstinline!FillPattern! attributes \lstinline!Horizontal!, \lstinline!Vertical!, \lstinline!Cross!, \lstinline!Forward!, \lstinline!Backward! and \lstinline!CrossDiag! specify fill patterns drawn with the line color over the fill color.
 
 The attributes \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder! and \lstinline!Sphere! specify gradients that represent a horizontal cylinder, a vertical cylinder and a sphere, respectively.
-The gradient goes from line color to fill color.
-For polygons the gradients are defined on the enclosing rectangle (for \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder!) or circle (for \lstinline!Sphere!).
-For ellipses the gradients are defined such that the perimeter has line color.
 The gradients colors are defined considering full rectangles (ignoring rounded corners) and full ellipses (ignoring arcs).
+For polygons the gradients are defined on the enclosing rectangle (for \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder!) or circle (for \lstinline!Sphere!).
+The color gradient attains the fill color at the center, and attains the line color at two of the sides of the rectangle or the border of the ellipse.
 
 The border pattern attributes \lstinline!Raised!, \lstinline!Sunken! and \lstinline!Engraved! represent frames which are rendered in a tool-dependent way --- inside the extent of the filled shape.
 
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -924,9 +924,13 @@ The \lstinline!LinePattern! attribute \lstinline!Solid! indicates a normal line,
 The \lstinline!FillPattern! attributes \lstinline!Horizontal!, \lstinline!Vertical!, \lstinline!Cross!, \lstinline!Forward!, \lstinline!Backward! and \lstinline!CrossDiag! specify fill patterns drawn with the line color over the fill color.
 
 The attributes \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder! and \lstinline!Sphere! specify gradients that represent a horizontal cylinder, a vertical cylinder and a sphere, respectively.
-The gradients colors are defined considering full rectangles (ignoring rounded corners) and full ellipses (ignoring arcs).
-For polygons the gradients are defined on the enclosing rectangle (for \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder!) or ellipse (for \lstinline!Sphere!).
-The color gradient attains the fill color at the center, and attains the line color at two of the sides of the rectangle or the border of the ellipse.
+Gradients are defined for the geometry of a \lstinline!GraphicItem! before its \lstinline!rotation! is applied.
+The coloring is defined for an enclosing shape, and then clipped to the actual shape of the \lstinline!GraphicItem!.
+For \lstinline!HorizontalCylinder! and \lstinline!VerticalCylinder!, the minimal enclosing axis-parallel rectangle is used, while \lstinline!Sphere! uses the smallest enclosing ellipse.
+The enclosing shapes are defined considering full rectangles (ignoring rounded corners) and full ellipses (ignoring arcs).
+For \lstinline!HorizontalCylinder!, the gradient attains the fill color along the horizontal line dividing the enclosing rectangle at the center, and the line color at the two horizontal sides of the rectangle.
+The \lstinline!VerticalCylinder! is defined analogously to \lstinline!HorizontalCylinder!, but using vertical lines.
+For \lstinline!Sphere!, the gradient attains the fill color at the center of the enclosing ellipse, and the line color along the entire border of the ellipse.
 
 The border pattern attributes \lstinline!Raised!, \lstinline!Sunken! and \lstinline!Engraved! represent frames which are rendered in a tool-dependent way --- inside the extent of the filled shape.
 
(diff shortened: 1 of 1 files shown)


maltelenz (changes requested):
There are libraries out there that use this, and it is clearly useful, so removing this feature is not reasonable.

Example usage of `HorizontalCylinder` on a `Polygon`:

Image

```
model Package
  annotation(
    Icon(
      coordinateSystem(extent = {{-100, -100}, {100, 100}}, grid = {10, 10}),
      graphics = {
        Polygon(origin = {0.248, 0.044}, lineColor = {56, 56, 56}, fillColor = {128, 202, 255}, fillPattern = FillPattern.Solid, points = {{99.752, 100}, {99.752, 59.956}, {99.752, -50}, {100, -100}, {49.752, -100}, {-19.752, -100.044}, {-100.248, -100}, {-100.248, -50}, {-90.248, 29.956}, {-90.248, 79.956}, {-40.248, 79.956}, {-20.138, 79.813}, {-0.248, 79.956}, {19.752, 99.956}, {39.752, 99.956}, {59.752, 99.956}}, smooth = Smooth.Bezier),
        Polygon(origin = {0, -13.079}, lineColor = {192, 192, 192}, fillColor = {255, 255, 255}, pattern = LinePattern.None, fillPattern = FillPattern.HorizontalCylinder, points = {{100, -86.921}, {50, -86.921}, {-50, -86.921}, {-100, -86.921}, {-100, -36.921}, {-100, 53.079}, {-100, 103.079}, {-50, 103.079}, {0, 103.079}, {20, 83.079}, {50, 83.079}, {100, 83.079}, {100, 33.079}, {100, -36.921}}, smooth = Smooth.Bezier),
        Polygon(origin = {0, -10.704}, lineColor = {113, 113, 113}, fillColor = {255, 255, 255}, points = {{100, -89.296}, {50, -89.296}, {-50, -89.296}, {-100, -89.296}, {-100, -39.296}, {-100, 50.704}, {-100, 100.704}, {-50, 100.704}, {0, 100.704}, {20, 80.704}, {50, 80.704}, {100, 80.704}, {100, 30.704}, {100, -39.296}}, smooth = Smooth.Bezier)
      }
    )
  );
end Package;
```

d-hedberg:
System Modeler
image

chapters/annotations.tex
```diff
@@ -925,6 +925,7 @@ \subsubsection{Graphical Properties}\label{graphical-properties}
 
 The attributes \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder! and \lstinline!Sphere! specify gradients that represent a horizontal cylinder, a vertical cylinder and a sphere, respectively.
 The gradient goes from line color to fill color.
+The \lstinline!HorizontalCylinder! and \lstinline!VerticalCylinder! are only defined for rectangles and ellipses (not polygons), and \lstinline!Sphere! is only defined for ellipses.
```
```suggestion
For polygons the gradients are defined on the enclosing rectangle (for \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder!) or circle (for \lstinline!Sphere!).
```
Trying to shortly explain how the gradients work for polygons, based on feedback.

As far as I understand "enclosing" generally means "smallest enclosing".
Note that it is subtly different that basing it on vertices (in case of splines) - as long as you don't do any weird extrapolation users will not notice.

henrikt-ma:
Please don't take https://github.com/modelica/ModelicaSpecification/pull/3789#issuecomment-3618298772 as a concrete proposal; it has several drawbacks:
- The idea of using the local origin will break appearance badly for existing use if the origin isn't near "the center".
- There are many more centers of mass that could be sensible choices, even without adding the more numerically demanding ones to the list, so picking one of them would become an arbitrary design decision.
- It would be disruptive for System Modeler users.
- It would require implementation effort for Wolfram to comply with the specification.

One thing that I believe might help us reach agreement would be to introduce a `gradientCenter` attribute in `FilledShape`, which would be optional for `Rectangle` and `Ellipse`, defaulting to being at the obvious center of the shape.  For `Polygon` it should be non-optional, but with a deprecated semantics of tool-specific behavior when being unspecified.

However, when thinking more about how to define the "gradient length", I realized that it wasn't clear to me how `Sphere` should be applied even to a non-circular `Ellipse`.  Should it be A or B below?
ellipses

chapters/annotations.tex
```diff
@@ -925,6 +925,9 @@ \subsubsection{Graphical Properties}\label{graphical-properties}
 
 The attributes \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder! and \lstinline!Sphere! specify gradients that represent a horizontal cylinder, a vertical cylinder and a sphere, respectively.
 The gradient goes from line color to fill color.
+For polygons the gradients are defined on the enclosing rectangle (for \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder!) or circle (for \lstinline!Sphere!).
```
henrikt-ma: I think you meant _ellipse_?
```suggestion
For polygons the gradients are defined on the enclosing rectangle (for \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder!) or ellipse (for \lstinline!Sphere!).
```

Also, it might be more to the point to speak of smallest horizontal or vertical band, than to speak of a rectangle where two of the sides are relevant?

chapters/annotations.tex
```diff
@@ -925,6 +925,9 @@ \subsubsection{Graphical Properties}\label{graphical-properties}
...
+For ellipses the gradients are defined such that the perimeter has line color.
```
henrikt-ma: Is this a more clear way of expressing it?
```suggestion
The color gradient attains the fill color at the center of the band or the center of the enclosing ellipse, and attains the line color at the two borders of the band or the border of the ellipse.
```

chapters/annotations.tex
```diff
@@ -925,6 +925,9 @@ \subsubsection{Graphical Properties}\label{graphical-properties}
...
+The gradients colors are defined considering full rectangles (ignoring rounded corners) and full ellipses (ignoring arcs).
```
henrikt-ma: Something should be done about the grammar of this sentence:
```suggestion
The enclosing band or ellipse is defined considering full rectangles (ignoring rounded corners) and full ellipses (ignoring arcs).
```

henrikt-ma: It would also be better to put this sentence before the sentence above, so that we first complete the specification of the enclosing band/ellipse, before speaking of the coloring.

chapters/annotations.tex
HansOlsson: Using a rectangular bounding-box seems so standardized that I prefer to use it, even if we only care about two of its sides.

chapters/annotations.tex
henrikt-ma: Will start over based on new text.  Resolving.

chapters/annotations.tex
```diff
@@ -924,7 +924,9 @@ \subsubsection{Graphical Properties}\label{graphical-properties}
 The \lstinline!FillPattern! attributes \lstinline!Horizontal!, \lstinline!Vertical!, \lstinline!Cross!, \lstinline!Forward!, \lstinline!Backward! and \lstinline!CrossDiag! specify fill patterns drawn with the line color over the fill color.
 
 The attributes \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder! and \lstinline!Sphere! specify gradients that represent a horizontal cylinder, a vertical cylinder and a sphere, respectively.
-The gradient goes from line color to fill color.
+The gradients colors are defined considering full rectangles (ignoring rounded corners) and full ellipses (ignoring arcs).
+For polygons the gradients are defined on the enclosing rectangle (for \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder!) or ellipse (for \lstinline!Sphere!).
```
henrikt-ma: (Github has messed up the mechanism for adding comments, so bear with me.)

Now that we have a sentence about the colors at the end of the paragraph, we should only speak of the geometry here, and it needs to be done so that it describes all combinations of shape and color gradient.

```suggestion
Gradients are defined for the geometry of a \lstinline!GraphicItem! before its \lstinline!rotation! is applied.
The coloring is defined for an enclosing shape, and then clipped to the actual shape of the \lstinline!GraphicItem!.
For \lstinline!HorizontalCylinder! and \lstinline!VerticalCylinder!, the minimal enclosing axis-parallel rectangle is used, while \lstinline!Sphere! uses the smallest enclosing ellipse.
The enclosing shapes are defined considering full rectangles (ignoring rounded corners) and full ellipses (ignoring arcs).
```

chapters/annotations.tex
```diff
@@ -924,7 +924,9 @@ \subsubsection{Graphical Properties}\label{graphical-properties}
...
+The color gradient attains the fill color at the center, and attains the line color at two of the sides of the rectangle or the border of the ellipse.
```
henrikt-ma: To clearly rule out one of the gradients we have seen in images posted in this PR, this needs to be more clear for the cylinders:

```suggestion
For \lstinline!HorizontalCylinder!, the gradient attains the fill color along the horizontal line dividing the enclosing rectangle at the center, and the line color at the two horizontal sides of the rectangle.
The \lstinline!VerticalCylinder! is defined analogously to \lstinline!HorizontalCylinder!, but using vertical lines.
For \lstinline!Sphere!, the gradient attains the fill color at the center of the enclosing ellipse, and the line color along the entire border of the ellipse.
```

chapters/annotations.tex
henrikt-ma: Resolving outdated conversation.

henrikt-ma (commented):
Now this looks very reasonable to me, but I'd like to leave it to others working closer to the graphical user interfaces to provide approving reviews.

d-hedberg:
Sounds good to me, but wouldn't it make sense to include some visual examples complementing the text? It would decrease the risk of different (mis)interpretations of the text.

chapters/annotations.tex
```diff
@@ -924,7 +924,13 @@ \subsubsection{Graphical Properties}\label{graphical-properties}
 The \lstinline!FillPattern! attributes \lstinline!Horizontal!, \lstinline!Vertical!, \lstinline!Cross!, \lstinline!Forward!, \lstinline!Backward! and \lstinline!CrossDiag! specify fill patterns drawn with the line color over the fill color.
 
 The attributes \lstinline!HorizontalCylinder!, \lstinline!VerticalCylinder! and \lstinline!Sphere! specify gradients that represent a horizontal cylinder, a vertical cylinder and a sphere, respectively.
-The gradient goes from line color to fill color.
+Gradients are defined for the geometry of a \lstinline!GraphicItem! before its \lstinline!rotation! is applied.
+The coloring is defined for an enclosing shape, and then clipped to the actual shape of the \lstinline!GraphicItem!.
+For \lstinline!HorizontalCylinder! and \lstinline!VerticalCylinder!, the minimal enclosing axis-parallel rectangle is used, while \lstinline!Sphere! uses the smallest enclosing ellipse.
```
henrikt-ma: It should be clarified that we don't mean the less computationally tractable smallest-perimeter ellipse:

```suggestion
For \lstinline!HorizontalCylinder! and \lstinline!VerticalCylinder!, the minimal enclosing axis-parallel rectangle is used, while \lstinline!Sphere! uses the smallest area enclosing ellipse.
```

henrikt-ma: Pinging @HansOlsson.

henrikt-ma:
In case there are concerns regarding the implementation effort for the current proposal, I have some notes and observations to share:
- For the geometries of `Ellipse` and `Rectangle`, it is trivial to find the minimal enclosing ellipse or axis-parallel rectangle (especially considering that these are determined before applying any rotation).
- For the less trivial `Polygon`, the minimal enclosing axis-parallel rectangle is very easy to compute also when using `smooth = Bezier`, as the $x$ and $y$ coordinates can be treated separately, only requiring the maximization of a quadratic polynomial for each spline segment.  (A less ambitious tool could say that this is a quality-of-implementation-thing, and ignore `smooth`, but why not do it properly when it is so easy?)
- The remaining case is finding the minimal area ellipse enclosing a `Polygon`, and I have explored this problem with the help of Claude Code, see below.

About the minimal area ellipse enclosing a `Polygon`:
- For `smooth = None`, this is a well-studied problem, and there exists an incremental method, Welzl's algorithm, for tracking the minimal area ellipse while adding one point at a time, with O(1) complexity for each addition.  It seems to have been originally developed for finding the minimal circle, but has been generalized for finding the minimal area ellipse.
- For `smooth = Bezier`, the following simple strategy turns out to converge very fast:
  * By only considering the Bezier spline end points, one obtains a "lower bound ellipse" using the same method as for `smooth = None`.
  * For each spline segment, compute the maximum value of the quadratic function defining the ellipse.
  * Drop Bezier splines found to be inside the lower bound ellipse.
  * An upper bound for the minimum area is obtained from the maximum value of the quadratic, providing a very nice stopping criterion.
  * Split the Bezier segment where the maximum value of the quadratic was obtained, resulting in a new spline segment endpoint which will increase the lower bound ellipse, and keep iterating.
- The "hard work" is to get the well known incremental method for `smooth = None` up and running; extending it to `smooth = Bezier` is just a thin layer on top.
- Claude Code was able to give me a C++ implementation for the above, completely free of external dependencies.
- The algorithm above was verified by comparing it to a few completely independent methods based on convex optimization, and found to give consistent results.
- The running time was in the order of a millisecond even with hundreds of points defining the shape.
- If a millisecond still seems expensive, note that the computation is only needed once per actual `Polygon` primitive using `Sphere`; the result can be cached, and when the `Polygon` is rotated or otherwise transformed, one only needs to transform the ellipse.

HansOlsson:
> I don't think we have talked about the fact that also `Polygon` is used with `Sphere` in the MSL.
> 
> This is what the `VoluminousWheel` would look like according to the current proposal (there are three `Polygon` items stacked here, where the `Sphere` gradient is only used for the bottom item, while the top item uses `HorizontalCylinder`):

That looks ok-ish.

Compared to HorizontalCylinder it goes to dark also at the outer edge of the wheel; so it more a balloon tire where that part is in shadow than a more usual cylindrical car tire with a sharp shadow. That is consistent with the polygon-outline that is rounded at the top.

henrikt-ma:
> > Regarding @HansOlsson's [comment](https://github.com/modelica/ModelicaSpecification/pull/3802#issuecomment-3717885019), I wouldn't find this PR a satisfactory fix for #1826 if didn't clarify how to draw the combination of `Rectangle` and `Sphere` (already in use in `Modelica.Mechanics.Translational.Components.Mass`, as noted above). I would also find it weird to specify how to draw `Sphere` on a `Polygon`, while leaving the more important application to `Rectangle` unspecified.
> 
> I still think the current state of the proposal is still a major step forward, and the intent wasn't to leave it unspecified - but only to leave room for different gradients **for the time being**, and for rectangle with spherical gradient allow both:

A big problem with this would be that the only supported way of getting somewhat predictable results for a rectangle would be to use `Polygon` instead of `Rectangle`.  It reminds me of some strange examples I've seen in the MSL, where a `Polygon` with `smooth = Bezier` is used to draw an ellipse (`PipeWithScalarField` is a prime example of this).  I think we should move away from such non-canonical shape representations, not leave the specification in a state where it also pushes library authors to change `Rectangle` into `Polygon`.

henrikt-ma:
> That looks ok-ish.
> 
> Compared to HorizontalCylinder it goes to dark also at the outer edge of the wheel; so it more a balloon tire where that part is in shadow than a more usual cylindrical car tire with a sharp shadow. That is consistent with the polygon-outline that is rounded at the top.

I am not saying it is a great-looking wheel, I am just showing what the current proposal would mean for this icon in its current state.  In order to work on the looks – if anyone would feel the urge – a first step must be to improve the specification so that the icon designer at least can have a rough idea of what to expect when using different combinations of shapes and fills.

HansOlsson:
Pinging @maltelenz

maltelenz (approved):
If the language group consensus is to move forward with the current state, I won't stand in the way, even though I agree with @henrikt-ma about still underspecified parts.

The current proposal is still a large improvement to the status quo.

PR #3791, opened 2025-12-04
Also mention evaluable in list of expression variabilities

henrikt-ma:
Just something we forgot when the evaluable variability was introduced.


HansOlsson (approved):
Ok

PR #3795, opened 2025-12-09
Describe spatialDistribution as event triggering

henrikt-ma:
We decided to make both `delay` and `spatialDistribution` event triggering operators, but so far the specification has only been updated for `delay`.  This PR does it also for `spatialDistribution`.

At least to me, trying to capture the intricacies of event triggering in the pseudo-code for `spatialDistribution` doesn't sound like a good idea.  Instead, I'd like us to consider the option of throwing out the pseudo-code, now that it is becoming even more remote from an actual implementation.

Note that we haven't test-implemented this in System Modeler, so I am counting on others to verify that the few things I have added regarding event triggering is in agreement with what has already been "test-implemented" in other tools.


chapters/operatorsandexpressions.tex
```diff
@@ -739,6 +758,130 @@ \subsubsection{delay}\label{delay}
...
+where $z(\xi, t)$ is the transported quantity, $\xi$ is the normalized spatial coordinate ($0 \le \xi \le 1$), $t$ is the time, $v(t)$ is the normalized transport velocity and the boundary conditions are set at either $\xi = 0$ or $\xi = 1$, depending on the sign of the velocity.
+(When discontinuities are present, the differentiation operators above need to be understood in a distributive sense.)
+
+The calling syntax is:
+\begin{lstlisting}[language=modelica]
+(out0, out1) = spatialDistribution(in0, in1, x, positiveVelocity,
+                                   initialPoints = {0.0, 1.0},
+                                   initialValues = {0.0, 0.0});
+\end{lstlisting}
+where \lstinline!in0!, \lstinline!in1!, \lstinline!out0!, \lstinline!out1!, and \lstinline!x! are all subtypes of \lstinline!Real!, \lstinline!positiveVelocity! is a \lstinline!Boolean!, and \lstinline!initialPoints! and \lstinline!initialValues! are arrays of subtypes of \lstinline!Real!.
```
henrikt-ma: I realize that I forgot to also lift the restriction to `Real`, so that we keep it as similar to `delay` as possible.  This was our intention, right?

HansOlsson:
@MarkusOlssonModelon and @gkurzbach state that implemented in Modelon Impact and SimulationX.
(Events if significant discontinuous change in input.)

gkurzbach (approved):
ok

MarkusOlssonModelon (approved):

chapters/operatorsandexpressions.tex
```diff
@@ -739,6 +758,134 @@ \subsubsection{delay}\label{delay}
...
+(When discontinuities are present, the differentiation operators above need to be understood in a distributive sense.)
+
+The calling syntax is:
+\begin{lstlisting}[language=modelica]
+(out0, out1) = spatialDistribution(in0, in1, x, positiveVelocity,
+                                   initialPoints = {0.0, 1.0},
+                                   initialValues = {0.0, 0.0});
+\end{lstlisting}
+The expressions \lstinline!in0!, \lstinline!in1!, \lstinline!out0!, and \lstinline!out1! shall all be subtypes of a common type $T$, where $T$ is either \lstinline!Real!, \lstinline!Integer!, \lstinline!Boolean!, or an enumeration type.
+Further, \lstinline!x! shall be a subtype of \lstinline!Real!, \lstinline!positiveVelocity! is a \lstinline!Boolean!, and \lstinline!initialPoints! and \lstinline!initialValues! are arrays of subtypes of $T$ and \lstinline!Real!, respectively.
```
```suggestion
The expressions \lstinline!in0!, \lstinline!in1!, \lstinline!out0!, and \lstinline!out1! shall all be subtypes of\lstinline!Real!.
Further, \lstinline!x! shall be a subtype of \lstinline!Real!, \lstinline!positiveVelocity! is a \lstinline!Boolean!, and \lstinline!initialPoints! and \lstinline!initialValues! are arrays of subtypes of \lstinline!Real!.
```
I should have realized this earlier, but non-Real spatialDistribution will cause problems for the initialValues.

Basically we normally interpolate between the initial points, but that doesn't work if the values are non-Reals.

I'm sure that it could be handled in some way, but I'm not confident that we will find all the issues before the release - and since there wasn't that much push for non-Real spatialDistribution I think it can wait.

chapters/operatorsandexpressions.tex
```diff
@@ -739,6 +758,134 @@ \subsubsection{delay}\label{delay}
...
+\end{equation*}
+This allows the direct computation of the solution based on interpolating the boundary conditions.
+
+When a \lstinline!spatialDistribution!-expression is discrete-time (see \cref{discrete-time-expressions}), events will be generated in order to allow the value to change at the correct points in time, and it is an error if both \lstinline!x! and \lstinline!in0! or \lstinline!in1! change at the same event.
+Further, when a \lstinline!spatialDistribution!-expression is non-discrete-time and event generation is enabled (not appearing inside \lstinline!noEvent!), discontinuities in \lstinline!in0! and \lstinline!in1! may be preserved in the distribution $z$, and trigger events when appearing in any of the operator outputs.
+It is a quality of implementation to avoid excessive generation of events by only preserving significant discontinuities.
+
+Disregarding the event-triggering handling of discontinuities, \lstinline!spatialDistribution! for $T$ being \lstinline!Real! can be described in terms of the pseudo-code given as a block:
```
HansOlsson, line 804:
```suggestion
Disregarding the event-triggering handling of discontinuities, \lstinline!spatialDistribution! can be described in terms of the pseudo-code given as a block:
```
Similarly.

HansOlsson (changes requested):
Overall good, but some issue with non-Real initial values.
Since that case seems a bit esoteric I think it is better to delay that and get the other changes.

chapters/operatorsandexpressions.tex
```diff
@@ -739,6 +758,134 @@ \subsubsection{delay}\label{delay}
...
+
+The solution, $z$, can be described in terms of characteristics:
+\begin{equation*}
+z(\xi+\int_{t}^{t+\beta}\! v(\alpha) \mathrm{d}\alpha,\, t+\beta) = z(\xi, t), \quad\text{for all $\beta$ as long as staying inside the domain}
+\end{equation*}
+This allows the direct computation of the solution based on interpolating the boundary conditions.
+
+When a \lstinline!spatialDistribution!-expression is discrete-time (see \cref{discrete-time-expressions}), events will be generated in order to allow the value to change at the correct points in time, and it is an error if both \lstinline!x! and \lstinline!in0! or \lstinline!in1! change at the same event.
+Further, when a \lstinline!spatialDistribution!-expression is non-discrete-time and event generation is enabled (not appearing inside \lstinline!noEvent!), discontinuities in \lstinline!in0! and \lstinline!in1! may be preserved in the distribution $z$, and trigger events when appearing in any of the operator outputs.
```
HansOlsson, line 797: I also realized that this line suggest that one can write:
`(, out1) = noEvent(spatialDistribution(in0, 0, x, true, {0,1}, {0,0}));`
to disable event generation for spatialDistribution, similarly as for other operators.

That makes sense, but two points:
* That should preferably be clarified, as `noEvent` was previously described for "expressions", but multi-returning functions aren't really "expressions".
* If `in0` is an integer expression it would seem that we generate a non-discrete-time integer - is that intentional? This might indicate that there might be additional problems for non-Real `spatialDistribution`.

HansOlsson:
After thinking some more I think I have a solution for the type issue (the noEvent-part seems fairly straightforward in comparison).

Consider `spatialDistribution(0, 1, x, posVec, initialPoints={0,1}, initialValues={0,1})`, previously this used interpolation for the initial points and generated a Real output. With the proposed $T$-type it seems it would generate an integer output.

I saw the following bad solutions:
- Have it generate an integer output and then fail during start-up as the initial values are not piece-wise constant (as required for Integers, right?).
- Switch to piece-wise constant interpolation since the values are integers (different simulation result)
- Deprecate this specific case
- Evaluate the initial values, and since they are not piece-wise constant keep it as real (weird)!

However, I then realized that there's a simple solution:
1. Keep the type of the current spatialDistribution as is.
2. Add a variant allowing `spatialDistribution(0, 1, x, posVec, initialPoints={0,1}, initialDiscrete={0,1})` (name initialDiscrete to be discussed) that generate piece-wise constant initial values, and with the proposed $T$-type for inputs and output. It seems clear how to handle non-Reals for this, and I would say that this variant shouldn't be legal in noEvent.

We have precedence for different over-loads in that sense for String.
Note that it isn't just using a different name to say that it's a different type - but it also has different semantics.

The benefit I see:
- Easy to use the new variant for Integer/Boolean etc (no need for duplicate points)
- Clearly backwards compatible
- Easy to distinguish the new variant; and give diagnostics

There are some minor unclear points for Reals (is it only for discrete-time Reals or also for continuous-time inputs? and the fact that discrete-time Reals are tied to when-semantics).

chapters/operatorsandexpressions.tex
henrikt-ma: Fortunately, we have deliberately decided to allow repeated time points in `initialPoints`, because we saw the need to represent discontinuities coming.  The fix is trivial: just require that a change in `initialValues` for a non-`Real` distribution requires a repeated point in `initialPoints`.  This has now been clarified.

I think we should hear whether @MarkusOlssonModelon or @gkurzbach can see any remaining issue.

HansOlsson: > Fortunately, we have deliberately decided to allow repeated time points in `initialPoints`, because we saw the need to represent discontinuities coming. The fix is trivial: just require that a change in `initialValues` for a non-`Real` distribution requires a repeated point in `initialPoints`.

As indicated in https://github.com/modelica/ModelicaSpecification/issues/3795#issuecomment-4294161129 it's more complicated, as "non-Real" is more complicated, but there's also a solution.

chapters/operatorsandexpressions.tex
henrikt-ma: > * That should preferably be clarified, as `noEvent` was previously described for "expressions", but multi-returning functions aren't really "expressions".

Done.

> * If `in0` is an integer expression it would seem that we generate a non-discrete-time integer - is that intentional? This might indicate that there might be additional problems for non-Real `spatialDistribution`.

Yes, that is intentional and worth explaining.  I have added a non-normative section.

henrikt-ma:
> Consider `spatialDistribution(0, 1, x, posVec, initialPoints={0,1}, initialValues={0,1})`, previously this used interpolation for the initial points and generated a Real output. With the proposed T -type it seems it would generate an integer output.

I think one should look at the entire equation, where $T$ can be determined based on the variables on the left-hand side.  This has now been clarified.

henrikt-ma:
> 2\. Add a variant allowing `spatialDistribution(0, 1, x, posVec, initialPoints={0,1}, initialDiscrete={0,1})`

We have already carefully prepared for allowing repeated `initialPoints` values to allow discontinuities in the buffer initialization.  Sticking to this design seems like the sensible thing to do, as this is how it will still have to be done for `Real` where there is a mix of interpolation and discontinuities.

henrikt-ma:
> There are some minor unclear points for Reals (is it only for discrete-time Reals or also for continuous-time inputs? and the fact that discrete-time Reals are tied to when-semantics).

The type $T$ is determined based on the output variables, while the _expression_ variability is determined based on the `spatialDistribution` expression on the right-hand side.  That the variability design is flawed and that we wouldn't be able to leverage that the outputs are discrete-time is not the fault of `spatialDistribution`; if we fix this more generally, also `spatialDistribution` will be able to benefit.

henrikt-ma:
> We have precedence for different over-loads in that sense for String.

Don't compare to `String`.  The relevant comparison is with `delay`.  This entire PR is about getting the same improvements to `spatialDistribution` that we have already made for `delay`.  When I implemented the new semantics for `delay`, it was the support for maintaining discontinuities that was the challenge; once this was done it was only icing on the cake to also support the non-`Real` types.

HansOlsson:
> > We have precedence for different over-loads in that sense for String.
> 
> Don't compare to `String`. The relevant comparison is with `delay`. This entire PR is about getting the same improvements to `spatialDistribution` that we have already made for `delay`. 

The delay part for spatialDistribution is similar to delay. 

However, the initialization of the buffer lacks a counter-part for delay (at the moment), and thus I found that additional named argument as for String may be a solution.

> When I implemented the new semantics for `delay`, it was the support for maintaining discontinuities that was the challenge; once this was done it was only icing on the cake to also support the non-`Real` types.

I agree that most of the implementation work is in maintaining the discontinuities - and I think that adding that for Reals bring real benefits. However, this proposed solution to the Integer problem would cause the left-hand-side to impact how the function in the right-hand side works, i.e., sort of overloading an operation based on intended result type, which is a new concept (and something I understand other languages like Java and C++ deliberately avoid).

It's also problematic that these issues were detected this late. I would much prefer if we postpone the non-Real arguments instead of adding some new overloading variant this close to the release.

henrikt-ma:
> I agree that most of the implementation work is in maintaining the discontinuities - and I think that adding that for Reals bring real benefits. However, this proposed solution to the Integer problem would cause the left-hand-side to impact how the function in the right-hand side works, i.e., sort of overloading an operation based on intended result type, which is a new concept (and something I understand other languages like Java and C++ deliberately avoid).

Sure, but that design was very short-lived in this PR.  With the parallel clarification of multi-output coercion rules in #3888, it became clear that $T$ can be determined from the right-hand side alone, and this is how the rule is currently formulated:
> Regarding standard type coercion (\cref{standard-type-coercion}), $T$ is \lstinline!Real! if and only if at least one of \lstinline!in0!, \lstinline!in1! and \lstinline!initialValues! is a subtype of \lstinline!Real!.

> It's also problematic that these issues were detected this late. I would much prefer if we postpone the non-Real arguments instead of adding some new overloading variant this close to the release.

As long as we are only making obvious clarifications that don't invalidate the two test implementations, I see no strong reason to hesitate. I am also hoping that there will be a third test implementation within soon (but I am not the one working on it).

HansOlsson:
> Sure, but that design was very short-lived in this PR. 

Oh good, I didn't get that from the note.

> With the parallel clarification of multi-output coercion rules in #3888, it became clear that T can be determined from the right-hand side alone, and this is how the rule is currently formulated:
> 
> > Regarding standard type coercion (\cref{standard-type-coercion}), 
> >   T
> >  is \lstinline!Real! if and only if at least one of \lstinline!in0!, \lstinline!in1! and \lstinline!initialValues! is a subtype of \lstinline!Real!.

To me that in combination with "When $T$ is not \lstinline!Real!, two adjacent values in \lstinline!initialValues! may only differ where the associated \lstinline!initialPoints! values are equal." implies that a currently valid model using:
```
   Real out, x;
equation
  (,out)=spatialDistribution(0,1,x,true,{0,1},{0,1});
```
is suddenly invalid, as $T$ is inferred to be Integer, and the initial part isn't step-wise constant.

I can see that it is esoteric and easy to patch.

HansOlsson:
Thus the minimum is to add a RationaleMCP\Changes\3_7\3795.md explaining the incompatibility.

I give others until Monday to see if the agree with the consequences. 
(Hopefully GitHub will also have restored searching for all pull requests by then.)

henrikt-ma:
> To me that in combination with "When T is not \lstinline!Real!, two adjacent values in \lstinline!initialValues! may only differ where the associated \lstinline!initialPoints! values are equal." implies that a currently valid model using:
> 
> ```
>    Real out, x;
> equation
>   (,out)=spatialDistribution(0,1,x,true,{0,1},{0,1});
> ```
> 
> is suddenly invalid, as T is inferred to be Integer, and the initial part isn't step-wise constant.
> 
> I can see that it is esoteric and easy to patch.

On the model side, yes.  I would argue that this is a backwards incompatible change we should accept for the sake of clean design.  I don't think it deserves an explanation in the specification; it is better to just let tools detect the problem so that the user can make one or more of the literals a `Real` literal instead.

henrikt-ma:
> Thus the minimum is to add a RationaleMCP\Changes\3_7\3795.md explaining the incompatibility.

Done.

HansOlsson (approved):
Ok

PR #3798, opened 2025-12-16
Non real solvability

### LINKED ISSUE #3779: Solvability requirement for non-Real if-equations
Separated out from: #3763 

Consider:
```
model A
  Boolean b=time<2;
  Integer i, j;
equation 
  if b then
    i=2;
  else
    j=4;
  end if;
  i+j=5;
end A;
```
This illustrates that equation for non-Real shouldn't only require that the variables are in "almost solved form", but also if-equations with non-evaluable conditions must have the same variables in "almost solved form" in order to easily handle the equation.

The problem is exactly how to formulate the requirement.
```
model B
  Boolean b=time<2;
  Integer i, j, k, m;
equation 
  k=2;
  m=4;
  if b then
    j=k+m;
  else
    j=4-m;
  end if;
  j=i+5;
end B;
```
Doesn't have the same set of variables solvable in all branches, but it has a set of variables (`j`,`m`) solvable in all branches. (If the last equation were `j=i*2+5` the overall system wouldn't be solvable, but that is a global check, not a local one.)

One simple approach to handle if-equations is to merge the branches to construct if-expressions. That cannot be done _uniquely_ here as it would either be` j=if b then k+m` else 4-m; or `m=if b then j-k else 4+j;`, but it is not possible to have both `j` and `m` in almost solved form in the same merged equation.

If there are multiple equations with multiple variables it may even be difficult to find all possibilities.

So, it seems quite complicated.

One simple restriction would be to restrict them even further and state that for non-Real equation in non-evaluable if-equations there must appear unique variables in the left-hand-sides of the equations in all branches, as the "almost solved form" was primarily to handle connection-sums and similar cases that aren't needed inside if-equations. If that is too restrictive we can later extend it.

Note that this is just for non-Real equations, excluding e.g., `j=x*0.5;`

HansOlsson:
According to decision (as far as I can see).
Closes #3779

### DIFF
--- a/chapters/equations.tex
+++ b/chapters/equations.tex
@@ -195,7 +195,7 @@ The \lstinline!if!-equations which do not have exclusively evaluable expressions
 Additional restrictions apply in combination with \lstinline!when!-equations, see \cref{where-a-when-equation-may-occur} and \cref{equations-within-when-equations}.
 
 \begin{nonnormative}
-If first condition is violated, the single assignment rule would not hold, because the number of equations may change during simulation although the number of unknowns remains the same.
+If the first condition is violated, the single assignment rule would not hold, because the number of equations may change during simulation although the number of unknowns remains the same.
 The second condition provides the generalization of \eqref{eq:dae-discrete-valued} to \lstinline!if!-equations.
 \end{nonnormative}
 


chapters/equations.tex
```diff
@@ -181,11 +181,20 @@ \subsection{If-Equations}\label{if-equations}
 In an equation section, the equations in the body are seen as equations that must be satisfied.
 The bodies that are not selected have no effect on that model evaluation.
 
-The \lstinline!if!-equations in equation sections which do not have exclusively parameter expressions as switching conditions shall have the same number of equations in each branch (a missing else is counted as zero equations and the number of equations is defined as in \cref{local-equation-size}).
+The \lstinline!if!-equations in equation sections which do not have exclusively parameter expressions as switching conditions shall satisfy the following:
```
henrikt-ma: Let's take the opportunity to get rid of some redundancy, and avoid a potentially confusing reference (equation sections don't have switching conditions):

```suggestion
The \lstinline!if!-equations which do not have exclusively parameter expressions as switching conditions shall satisfy the following:
```

chapters/equations.tex
```diff
@@ -181,11 +181,20 @@ \subsection{If-Equations}\label{if-equations}
...
+\begin{itemize}
+\item
+  Have the same number of equations in each branch (a missing else is counted as zero equations and the number of equations is defined as in \cref{local-equation-size}).
```
henrikt-ma: Formatting, and it's not really _missing_:

```suggestion
  Have the same number of equations in each branch, where the number of equations is defined as in \cref{local-equation-size}.
  Absence of an \lstinline!else!-branch is treated as having a branch with zero equations.
```

chapters/equations.tex
```diff
@@ -181,11 +181,20 @@ \subsection{If-Equations}\label{if-equations}
...
+\item
+  If the equations are non-Real they shall have component-references (or a list of them) as their left-hand-side.
```
henrikt-ma: Formatting:

```suggestion
  If the equations are non-\lstinline!Real!, they shall have component-references (or a list of them) as their left-hand-side.
```

chapters/equations.tex
```diff
@@ -181,11 +181,20 @@ \subsection{If-Equations}\label{if-equations}
...
+  Any indices for such component-references must be evaluable.
+  All branches shall have the same set of variables in the non-Real equations.
```
henrikt-ma: Formatting:

```suggestion
  All branches shall have the same set of variables in the non-\lstinline!Real! equations.
```

chapters/equations.tex
```diff
@@ -181,11 +181,20 @@ \subsection{If-Equations}\label{if-equations}
...
+\end{itemize}
 Additional restrictions apply in combination with \lstinline!when!-equations, see \cref{where-a-when-equation-may-occur} and \cref{equations-within-when-equations}.
 
 \begin{nonnormative}
-If this condition is violated, the single assignment rule would not hold, because the number of equations may change during simulation although the number of unknowns remains the same.
+If first condition is violated, the single assignment rule would not hold, because the number of equations may change during simulation although the number of unknowns remains the same.
+The second condition ensures that it is easy to solve the non-Real equations.
```
henrikt-ma: Formatting:

```suggestion
The second condition ensures that it is easy to solve the non-\lstinline!Real! equations.
```

henrikt-ma: Actually, we can be more specific:

```suggestion
The second condition provides the generalization of \eqref{eq:dae-discrete-valued} to \lstinline!if!-equations.
```

It would also be good to add a cross reference in the other direction:
```
\begin{itemize}
\item
  Non-\lstinline!Integer! equations at most requiring flipping sides of the equation to obtain the used assignment form.
\item
  For \lstinline!Integer! equations the solved variable must appear uniquely as a term (without any multiplicative factor) on either side of the equation, at most requiring addition or subtraction of other terms in the equation to obtain the used assignment form.
\end{itemize}
\item
  Rules for \lstinline!if!-equations are given in \cref{if-equations}.
```

chapters/equations.tex
```diff
@@ -181,11 +181,20 @@ \subsection{If-Equations}\label{if-equations}
...
 The bodies that are not selected have no effect on that model evaluation.
 
-The \lstinline!if!-equations in equation sections which do not have exclusively parameter expressions as switching conditions shall have the same number of equations in each branch (a missing else is counted as zero equations and the number of equations is defined as in \cref{local-equation-size}).
+The \lstinline!if!-equations in equation sections which do not have exclusively parameter expressions as switching conditions shall satisfy the following:
+\begin{itemize}
+\item
+  Have the same number of equations in each branch (a missing else is counted as zero equations and the number of equations is defined as in \cref{local-equation-size}).
+\item
+  If the equations are non-Real they shall have component-references (or a list of them) as their left-hand-side.
+  Any indices for such component-references must be evaluable.
```
henrikt-ma: I think we call then _subscripts_, compare `\cref{connect-equations-and-connectors}`:

```suggestion
  Any subscripts for such component-references must be evaluable.
```

chapters/equations.tex
henrikt-ma: Also, in https://github.com/modelica/ModelicaSpecification/issues/3779#issuecomment-3612632733 we said that this should be tied to evaluability, so I would have expected this:
```suggestion
The \lstinline!if!-equations which do not have exclusively evaluable expressions as switching conditions shall satisfy the following:
```

chapters/equations.tex
HansOlsson: Now also added, but obviously inside the list :-)

chapters/equations.tex
```diff
@@ -181,11 +181,21 @@ \subsection{If-Equations}\label{if-equations}
...
 The bodies that are not selected have no effect on that model evaluation.
 
-The \lstinline!if!-equations in equation sections which do not have exclusively parameter expressions as switching conditions shall have the same number of equations in each branch (a missing else is counted as zero equations and the number of equations is defined as in \cref{local-equation-size}).
+The \lstinline!if!-equations which do not have exclusively evaluable expressions as switching conditions shall satisfy the following:
+\begin{itemize}
+\item
+  Have the same number of equations in each branch, where the number of equations is defined as in \cref{local-equation-size}.
+  Absence of an \lstinline!else!-branch is treated as having a branch with zero equations.
+\item
+  If the equations are non-\lstinline!Real!, they shall have component-references (or a list of them) as their left-hand-side.
```
henrikt-ma: I see now that we can make it more clear that we are speaking of the equations inside the branches:

```suggestion
  Non-\lstinline!Real! equations in the \lstinline!if!-equation branches shall have component-references (or a list of them) as their left-hand-side.
```

HansOlsson: Sure, to be extra careful.

chapters/equations.tex
```diff
@@ -181,11 +181,21 @@ \subsection{If-Equations}\label{if-equations}
...
 The bodies that are not selected have no effect on that model evaluation.
 
-The \lstinline!if!-equations in equation sections which do not have exclusively parameter expressions as switching conditions shall have the same number of equations in each branch (a missing else is counted as zero equations and the number of equations is defined as in \cref{local-equation-size}).
+The \lstinline!if!-equations which do not have exclusively evaluable expressions as switching conditions shall satisfy the following:
+\begin{itemize}
+\item
+  Have the same number of equations in each branch, where the number of equations is defined as in \cref{local-equation-size}.
+  Absence of an \lstinline!else!-branch is treated as having a branch with zero equations.
+\item
+  Non-\lstinline!Real! equations in the \lstinline!if!-equation branches shall have component-references (or a list of them) as their left-hand-side.
```
henrikt-ma: I also realized that we need to distinguish between simple and non-simple equations:

```suggestion
  Non-\lstinline!Real! simple equality equations (see \cref{simple-equality-equations}) in the \lstinline!if!-equation branches shall have component-references (or a list of them) as their left-hand-side.
```

HansOlsson: The "list of them" was for non-simple, so if we only mention simple it would be:
```suggestion
  Non-\lstinline!Real! simple equality equations (see \cref{simple-equality-equations}) in the \lstinline!if!-equation branches shall have component-references as their left-hand-side.
```

chapters/equations.tex
```diff
@@ -181,11 +181,21 @@ \subsection{If-Equations}\label{if-equations}
...
+  Any subscripts for such component-references must be evaluable.
```
henrikt-ma, line 191: We also need something for the non-simple equations:

```suggestion
  Any subscripts for such component-references must be evaluable.
  Any \lstinline!for!- and \lstinline!if!-equations in the \lstinline!if!-equation branches shall have evaluable controlling conditions, and contain equations which fullfil these requirements recursively.
```

HansOlsson: To me the recursive part worked automatically: for-equations shall always have evaluable conditions, and nested if-equations containing non-Real equations follow from just applying the rule for that if-equation.

chapters/equations.tex
henrikt-ma: Hmm… this is one of the examples in `\cref{simple-equality-equations}`:
```
(out1, out2, out3) = function_name(inexpr1, inexpr2);
```
What am I missing?

chapters/equations.tex
henrikt-ma: Besides the requirements on these equations, I also wanted to make clear that they may qualify for the matching of branches, despite not being given in simple equality equation form.

chapters/equations.tex
HansOlsson: I missed the exact definition of "simple".

chapters/equations.tex
```diff
@@ -181,11 +181,22 @@ \subsection{If-Equations}\label{if-equations}
...
 The bodies that are not selected have no effect on that model evaluation.
 
-The \lstinline!if!-equations in equation sections which do not have exclusively parameter expressions as switching conditions shall have the same number of equations in each branch (a missing else is counted as zero equations and the number of equations is defined as in \cref{local-equation-size}).
+The \lstinline!if!-equations which do not have exclusively evaluable expressions as switching conditions shall satisfy the following:
+\begin{itemize}
+\item
+  Have the same number of equations in each branch, where the number of equations is defined as in \cref{local-equation-size}.
+  Absence of an \lstinline!else!-branch is treated as having a branch with zero equations.
+\item
+  Non-\lstinline!Real! simple equality equations (see \cref{simple-equality-equations}) in the \lstinline!if!-equation branches shall have component-references as their left-hand-side.
```
henrikt-ma: Didn't the wrong suggestion get applied here?

```suggestion
  Non-\lstinline!Real! simple equality equations (see \cref{simple-equality-equations}) in the \lstinline!if!-equation branches shall have component-references (or a list of them) as their left-hand-side.
```

HansOlsson: Likely. Will use this one.

chapters/equations.tex
```diff
@@ -181,11 +181,22 @@ \subsection{If-Equations}\label{if-equations}
...
+  Non-\lstinline!Real! simple equality equations (see \cref{simple-equality-equations}) in the \lstinline!if!-equation branches shall have component-references (or a list of them) as their left-hand-side.
+  Any subscripts for such component-references must be evaluable.
+  Any \lstinline!for!- and \lstinline!if!-equations in the \lstinline!if!-equation branches shall have evaluable controlling conditions, and contain equations which fullfil these requirements recursively.
+  All branches shall have the same set of variables in the non-\lstinline!Real! equations.
+\end{itemize}
 Additional restrictions apply in combination with \lstinline!when!-equations, see \cref{where-a-when-equation-may-occur} and \cref{equations-within-when-equations}.
 
 \begin{nonnormative}
-If this condition is violated, the single assignment rule would not hold, because the number of equations may change during simulation although the number of unknowns remains the same.
+If first condition is violated, the single assignment rule would not hold, because the number of equations may change during simulation although the number of unknowns remains the same.
```
henrikt-ma: Missing _the_:

```suggestion
If the first condition is violated, the single assignment rule would not hold, because the number of equations may change during simulation although the number of unknowns remains the same.
```

henrikt-ma (changes requested):
Basically good, just a trivial suggestion.

PR #3801, opened 2025-12-18
Update unit-overview based on comment.

HansOlsson:
As discussed at latest Modelica unit meeting.

Note that the text under Detailed rules could be added to the specification (including all the specific rules from the linked appendix, but I don't want to start editing the specification text until we have agreement).


RationaleMCP/0027/ReadMe.md
```diff
@@ -11,24 +11,91 @@ The problem with undefined unit is that it gets in the way of carrying out check
...
+Generally it correctly finds some issues in libraries, but some libraries have systematic issues.
+E.g., the buildings library uses a large number of multiplicative literals without unit for converting between different time and power units. 
 
 ## Required Patents
 To the best of our knowledge, there are no patents that would conflict with the incorporation of this MCP.
 
 ## References
-(None.)
+https://doi.org/10.3384/ecp21817 and its references.
+
```
henrikt-ma, line 52: The content from here to the end of the file does not belong in the _ReadMe.md_, but should go into a separate design document, referenced from this file.

PR #3802, opened 2025-12-24
Images with examples of gradient fills

henrikt-ma:
These are the images that I promised to provide for #3789.


henrikt-ma:
This PR only provides examples for the specification text proposed in #3789.  I think we should try to have discussions about the actual specification in #3789, and only discuss the examples as such in this PR.

HansOlsson:
I think I would prefer these images without the side-circles and triangle (intending to show the two colors), possibly having those colors somewhere else for reference. Just having a gradient between black and white instead of a copper tone may also reduce the need for reference colors.

DagBruck (approved):
I'm ok with these changes to the document, but as it says -- there may be differences in the rendering between tools.

PR #3805, opened 2026-01-07
Define a^s for matrices

### LINKED ISSUE #3800: Scalar exponentiation of Integer matrix
Reading https://specification.modelica.org/master/arrays.html#scalar-exponentiation-of-matrices, I would guess that the following is valid:
```
Integer[2, 2] m = [1, 1; 1, 1]^2;
```

However, the specification of scalar exponentiation of a matrix stands out by not giving typing rules.

HansOlsson:
Closes #3800

### DIFF
--- a/chapters/arrays.tex
+++ b/chapters/arrays.tex
@@ -1365,7 +1365,7 @@ This is a consequence of the parsing rules, i.e., since \lstinline!2.! could be 
 
 Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square numeric matrix, and \lstinline!s! is a scalar subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
 The result has the same type as \lstinline!a!.
-The exponentiation is defined mathematically as repeated multiplication, with a identity matrix of suitable type as the result for $\text{\lstinline!s!} = 0$, e.g.:
+The exponentiation is defined mathematically as repeated multiplication, with the identity matrix as the result for $\text{\lstinline!s!} = 0$, e.g.:
 \begin{lstlisting}[language=modelica]
 a^1 = a;
 a^2 = a * a;


chapters/arrays.tex
```diff
@@ -1363,14 +1363,15 @@ \subsection{Element-wise Exponentiation}\label{element-wise-exponentiation}
...
+Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square matrix of a subtype of \lstinline!Real!, and \lstinline!s! is a scalar as a subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
 The exponentiation is done by repeated multiplication, e.g.:
 \begin{lstlisting}[language=modelica]
 a^3 = a * a * a;
-a^0 = identity(size(a, 1));
+a^0 = identity(size(a, 1)) * 1.0;
 assert(size(a, 1) == size(a, 2), "Matrix must be square");
 a^1 = a;
 \end{lstlisting}
+Exponentation of a matrix of integers is defined by first converting it to reals according to \cref{standard-type-coercion}, in order to avoid integer overflow.
```
henrikt-ma: This goes without saying.  If you really feel the urge to clarify, I think it should go into the non-normative paragraph below.

chapters/arrays.tex
```diff
@@ -1363,14 +1363,15 @@ \subsection{Element-wise Exponentiation}\label{element-wise-exponentiation}
 
 \subsection{Scalar Exponentiation of Matrices}\label{scalar-exponentiation-of-square-matrices-of-numeric-elements}\label{scalar-exponentiation-of-matrices}
 
-Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square numeric matrix and \lstinline!s! is a scalar as a subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
+Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square matrix of a subtype of \lstinline!Real!, and \lstinline!s! is a scalar as a subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
```
henrikt-ma: There's an old typo we should fix too.  Further, to keep this aligned with how other operators are defined, we should also specify explicitly that the result is a `Real` matrix.

```suggestion
Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square matrix of a subtype of \lstinline!Real!, and \lstinline!s! is a scalar subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
The result is a \lstinline!Real! matrix.
```

chapters/arrays.tex
```diff
@@ -1363,14 +1363,15 @@ \subsection{Element-wise Exponentiation}\label{element-wise-exponentiation}
...
 The exponentiation is done by repeated multiplication, e.g.:
 \begin{lstlisting}[language=modelica]
 a^3 = a * a * a;
-a^0 = identity(size(a, 1));
+a^0 = identity(size(a, 1)) * 1.0;
 assert(size(a, 1) == size(a, 2), "Matrix must be square");
 a^1 = a;
```
henrikt-ma: This example is a mess.  Let's at least get rid of the assert, but possibly also the trivial `a^1` line.  Alternatively, keep the `a^1` line, but order the lines by increasing exponent.

Also, the `a^0` line obviously looked a lot better before, so maybe we should consider saying it with words instead?

chapters/arrays.tex
```diff
@@ -1363,14 +1363,15 @@ \subsection{Element-wise Exponentiation}\label{element-wise-exponentiation}
 
 \subsection{Scalar Exponentiation of Matrices}\label{scalar-exponentiation-of-square-matrices-of-numeric-elements}\label{scalar-exponentiation-of-matrices}
 
-Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square numeric matrix and \lstinline!s! is a scalar as a subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
+Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square matrix of a subtype of \lstinline!Real!, and \lstinline!s! is a scalar as a subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
 The exponentiation is done by repeated multiplication, e.g.:
```
henrikt-ma: Actually performing repeated multiplication sounds like a bad idea to me.  Perhaps we should also make an effort to avoid specifying _how_ to compute the result, in order to give room for implementations to compute something closer to a correctly rounded result?

Simply saying _defined as_ instead of _done by_ would be a step in the right direction:

```suggestion
The exponentiation is defined as repeated multiplication, e.g.:
```

HansOlsson: Yes, that makes sense. Note that repeated multiplication isn't as bad as it may seem, as the order isn't defined which should allow recursive formulations (assuming anyone actually wants a high power).

chapters/arrays.tex
HansOlsson: I re-ordered them, replaced the assert by a^2, and replaced the zero-case by text.

chapters/arrays.tex
henrikt-ma: Well, when I head "done by repeated multiplication", I only hear two possible interpretations; repeated multiplication from left, and repeated multiplication from right.  This also points at a danger with using some sort of Modelica pseudo code in the semi-example included in the definition, as the Modelica syntax actually comes with a particular choice of operator associativity.

chapters/arrays.tex
```diff
@@ -1363,14 +1363,15 @@ \subsection{Element-wise Exponentiation}\label{element-wise-exponentiation}
 
 \subsection{Scalar Exponentiation of Matrices}\label{scalar-exponentiation-of-square-matrices-of-numeric-elements}\label{scalar-exponentiation-of-matrices}
 
-Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square numeric matrix and \lstinline!s! is a scalar as a subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
-The exponentiation is done by repeated multiplication, e.g.:
+Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square matrix of a subtype of \lstinline!Real!, and \lstinline!s! is a scalar subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
+The result is a \lstinline!Real! matrix.
+The exponentiation is defined as repeated multiplication, e.g.:
```
henrikt-ma: Avoid a sentence starting with "And" below by including text here instead:

```suggestion
The exponentiation is defined as repeated multiplication, with a \lstinline!Real! identity matrix as the result for $\text{\lstinline!s!} = 0$, e.g.:
```

chapters/arrays.tex
```diff
@@ -1363,14 +1363,15 @@ \subsection{Element-wise Exponentiation}\label{element-wise-exponentiation}
...
 \begin{lstlisting}[language=modelica]
-a^3 = a * a * a;
-a^0 = identity(size(a, 1));
-assert(size(a, 1) == size(a, 2), "Matrix must be square");
 a^1 = a;
+a^2 = a * a;
+a^3 = a * a * a;
 \end{lstlisting}
+And \lstinline!a ^ 0! is a real-valued identity matrix.
```
henrikt-ma: Not needed if suggestion above is accepted:

```suggestion
```

chapters/arrays.tex
henrikt-ma: Would it also help to throw in _mathematically_ to reduce risk of interpreting the examples as some sort of reference implementation?

```suggestion
The exponentiation is defined mathematically as repeated multiplication, with a \lstinline!Real! identity matrix as the result for $\text{\lstinline!s!} = 0$, e.g.:
```

chapters/arrays.tex
henrikt-ma: Continued in another conversation.  Resolving.

chapters/arrays.tex
henrikt-ma: Looks good now.  Resolving.

HansOlsson:
Language group: 

Scalar case always give real result, but allow negative exponent so a bit different. 
However, risk that people use `scalar([2]^3)` instead of `integer(2^3)`

Reason for integer - defined by multiplication and makes sense for permutations/shift matrices.

Poll, ok with:
- Real result, Hans
- Integer result, Gerd, Hans, Henrik

So reformulate as Integer giving Integer. @HansOlsson

chapters/arrays.tex
```diff
@@ -1363,13 +1363,13 @@ \subsection{Element-wise Exponentiation}\label{element-wise-exponentiation}
 
 \subsection{Scalar Exponentiation of Matrices}\label{scalar-exponentiation-of-square-matrices-of-numeric-elements}\label{scalar-exponentiation-of-matrices}
 
-Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square numeric matrix and \lstinline!s! is a scalar as a subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
-The exponentiation is done by repeated multiplication, e.g.:
+Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square matrix of a subtype of \lstinline!Real!, and \lstinline!s! is a scalar subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
```
henrikt-ma: According to yesterday's meeting:

```suggestion
Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square numeric matrix, and \lstinline!s! is a scalar subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
The result is a matrix of the same type as \lstinline!a!.
The exponentiation is defined mathematically as repeated multiplication, with an identity matrix as the result for $\text{\lstinline!s!} = 0$, e.g.:
```

(Stupid Github doesn't allow the suggestion to also replace the two lines below because of presence of earlier conversations which have already been resolved.)

chapters/arrays.tex
```diff
@@ -1363,13 +1363,13 @@ \subsection{Element-wise Exponentiation}\label{element-wise-exponentiation}
...
+The result is a \lstinline!Real! matrix.
+The exponentiation is defined mathematically as repeated multiplication, with a \lstinline!Real! identity matrix as the result for $\text{\lstinline!s!} = 0$, e.g.:
```
henrikt-ma: Workaround for issue with Github suggestion mentioned above:

```suggestion
```

chapters/arrays.tex
```diff
@@ -1363,13 +1363,13 @@ \subsection{Element-wise Exponentiation}\label{element-wise-exponentiation}
 
 \subsection{Scalar Exponentiation of Matrices}\label{scalar-exponentiation-of-square-matrices-of-numeric-elements}\label{scalar-exponentiation-of-matrices}
 
-Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square numeric matrix and \lstinline!s! is a scalar as a subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
-The exponentiation is done by repeated multiplication, e.g.:
+Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square matrix of a subtype of \lstinline!Real!, and \lstinline!s! is a scalar subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
+The result is a \lstinline!Real! matrix.
```
henrikt-ma: Workaround for issue with Gitbug suggestion mentioned above:

```suggestion
```

HansOlsson: I should have looked here before doing the same change, but I think the result is the same.

henrikt-ma: I'll resolve this to start with.

chapters/arrays.tex
henrikt-ma: Resolving outdated suggestion.

chapters/arrays.tex
```diff
@@ -1363,13 +1363,13 @@ \subsection{Element-wise Exponentiation}\label{element-wise-exponentiation}
 
 \subsection{Scalar Exponentiation of Matrices}\label{scalar-exponentiation-of-square-matrices-of-numeric-elements}\label{scalar-exponentiation-of-matrices}
 
-Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square numeric matrix and \lstinline!s! is a scalar as a subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
-The exponentiation is done by repeated multiplication, e.g.:
+Exponentiation \lstinline!a ^ s! is defined if \lstinline!a! is a square numeric matrix, and \lstinline!s! is a scalar subtype of \lstinline!Integer! with $\text{\lstinline!s!} \geq 0$.
+The result has the same type as \lstinline!a!.
+The exponentiation is defined mathematically as repeated multiplication, with a identity matrix of suitable type as the result for $\text{\lstinline!s!} = 0$, e.g.:
```
henrikt-ma: We have already said what the type of the result shall be in the previous sentence, so I don't think we need to repeat it here (note that this also fixes the typo in _a identity_):

```suggestion
The exponentiation is defined mathematically as repeated multiplication, with the identity matrix as the result for $\text{\lstinline!s!} = 0$, e.g.:
```

chapters/arrays.tex
henrikt-ma: Resolving outdated conversation.

chapters/arrays.tex
HansOlsson: Ok, as long as we don't use the identity-call that can work.

henrikt-ma: Exactly.

PR #3806, opened 2026-01-08
Change from parameter to evaluable.

### LINKED ISSUE #3799: Non-evaluable controlling conditions around when-equations
In https://specification.modelica.org/master/equations.html#restrictions-on-where-a-when-equation-may-occur, I believe it was an omission to not require _exclusively evaluable conditions_:
> • `when`-equations can only occur within `if`-equations and `for`-equations if the controlling expressions are exclusively parameter expressions.

Correct?

(This is something I found while reviewing #3798.)

HansOlsson:
Closes #3799
Note that there's one remaing parameter that also should be changed, but it is part of #3798

### DIFF
--- a/chapters/equations.tex
+++ b/chapters/equations.tex
@@ -352,7 +352,7 @@ equation
     end if;
   end when;
 \end{lstlisting}
-whereas the restriction to evaluable-expression is intended to allow:
+whereas the restriction to an evaluable expression is intended to allow:
 \begin{lstlisting}[language=modelica]
   parameter Boolean b = true;
   parameter Integer n = 3;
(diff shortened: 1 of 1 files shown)


chapters/equations.tex
```diff
@@ -352,7 +352,7 @@ \subsubsection{Equations within When-Equations}\label{restrictions-on-equations-
     end if;
   end when;
 \end{lstlisting}
-whereas the restriction to parameter-expression is intended to allow:
+whereas the restriction to evaluable-expression is intended to allow:
```
henrikt-ma: This is written without hyphenation everywhere else:

```suggestion
whereas the restriction to an evaluable expression is intended to allow:
```

chapters/equations.tex
```diff
@@ -921,7 +921,7 @@ \section{Initialization, initial equation, and initial algorithm}\label{initiali
     der(y) = 0;
   end if;
 \end{lstlisting}
-if \lstinline!steadyState! had not been a parameter-expression both of those equations would have been illegal according to the restrictions in \cref{if-equations}.
+if \lstinline!steadyState! had not been an evaluable parameter-expression both of those equations would have been illegal according to the restrictions in \cref{if-equations}.
```
henrikt-ma: Drop _parameter_:

```suggestion
if \lstinline!steadyState! had not been an evaluable expression, both of those equations would have been illegal according to the restrictions in \cref{if-equations}.
```

henrikt-ma (changes requested):
Needs minor stylistic changes.

PR #3808, opened 2026-01-13
Define that equations for the state-variables should be inside it

### LINKED ISSUE #3768: What is the situation in a state machine where variables in the state are assigned values outside the block
```
model StateInOut4
  inner Real value(start=0);
  block State1
    outer output Real value;
    Real a;
  equation 
    a = value;
    //a = time;
  end State1;

  block State2
    outer output Real value;
  equation 
    value = 1;
  end State2;

  State1 state1;
  State2 state2;
equation 
  initialState(state1);
  transition(state1,state2,ticksInState()>=5,reset=true,immediate=true);
  transition(state2,state1,ticksInState()>=5,reset=true,immediate=true);

  state1.a=time;
  annotation (experiment(StopTime=7, __Dymola_Algorithm="Dassl"));
end StateInOut4;
```
In this model, variable `state1.a` is assigned a value on the outer layer.
I checked the code exported from Dymola for this model, and this equation is not within the state activation control if-statement：
```
when _Clocks.BaseClock_0.SubClock_1.active then
  state1.a := time;
  if (_StateMachines.state1.activeState == stateMachine_of_state1.state_state1) then 
    state1.value := state1.a;
  end if;
  if (_StateMachines.state1.activeState == stateMachine_of_state1.state_state2) then 
    state2.value := 1;
  end if;
end when;  
```
However, in the direct simulation results in Dymola, it has no value when the state is not activated:
Image
If this model is exported as FMU and then imported into Dymola for simulation, variable `state1.a` will have a value at all times:
Image
So what should the expected result be? I think even if assigning values on the outer layer, variables in the state should still only be calculated when the state is activated.

HansOlsson:
Define that equations for the state-variables should be inside the states
Closes #3768

### DIFF
--- a/chapters/statemachines.tex
+++ b/chapters/statemachines.tex
@@ -24,6 +24,7 @@ The following properties are different to Lucid Synchrone 3.0:
   Modelica blocks can be used as states.
   They might contain clocked equations.
   If the equations are discretized, they are integrated between the previous and the current clock tick, if the corresponding state is active.
+  All equations for the variables defined in the state must be inside the state (this corresponds to restriction to balanced models, \cref{balanced-models}).
 \end{itemize}
 
 \section{Transitions}\label{transitions}


henrikt-ma (commented):
I think it would be better if @gkurzbach could review this, as they have implemented the state machines in SimulationX.

HansOlsson:
Language group: Gerd: "for sure", check inner/outer

gkurzbach (approved):
sure

PR #3815, opened 2026-01-26
Disambiguate passing of constant protected component references in external call

### LINKED ISSUE #3809: Passing protected constant variable in external function call
Should the following really be allowed?
```
    pure function extFun
      output Integer y;
    protected
      constant Integer c = 5;
    external "C"
      y = extFun(c);
    end extFun;
```

What makes me think it shouldn't is that the constant is passed by non-const reference to the external function, and the following:

> The value of a protected component passed to the external function should be considered undefined (destroyed) after the external function call.

It seems weird to allow that the constant local variable may be "destroyed" in this way.

Edit: I have a better reason in https://github.com/modelica/ModelicaSpecification/issues/3809#issuecomment-3786920519 below.

### LINKED ISSUE #3810: Passing component references from outside of the function in the external function call
I don't think it is clear what to do in this case:
```
constant Integer c = 5;

pure function extFun
  output Integer y;
external "C"
  y = extFun(c);
end extFun;
```

What makes me unsure is that I expected that all component references in the external function call would be possible to classify as either input, output, or protected.

Note that making `c` an array shows that this can not be understood as a way to pass a constant argument, since only scalar constant arguments are allowed:
```
constant Integer[1] c = {5};

pure function extFun
  output Integer y;
external "C"
  y = extFun(c);
end extFun;
```

henrikt-ma:
Fixes #3809. Fixes #3810.

Based on https://github.com/modelica/ModelicaSpecification/issues/3809#issuecomment-3789919930, but with more words to make it even more clear.

### DIFF
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -1942,7 +1942,7 @@ A component reference to a component that is part of an input or output is treat
 \subsection{Argument type Mapping}\label{argument-type-mapping}
 
 The arguments of the external function are declared in the same order as in the Modelica declaration, unless specified otherwise in an explicit external function call.
-Protected variables (i.e., temporaries) are passed in the same way as outputs, whereas constants and {\lstinline!size!} calls are passed as inputs.
+Values of constant expressions (including references to constant components) and \lstinline!size! calls are passed as inputs, whereas protected non-constant variables (i.e., temporaries) are passed in the same way as outputs.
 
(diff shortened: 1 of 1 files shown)


MarkusOlssonModelon (approved):

HansOlsson (approved):
Looking good.

PR #3817, opened 2026-01-27
Fix language mistakes in "left hand side" and "similarly as"

henrikt-ma:
Fixing language mistakes encountered when working on #3763.  Making a separate PR for this, so that the PR fixing #3763 can be cleaner.

### DIFF
--- a/RationaleMCP/Changes/3_7/3610.md
+++ b/RationaleMCP/Changes/3_7/3610.md
@@ -26,5 +26,5 @@ end foo;
 Real x=foo(time);
 ```
 
-The call of `bar` in `foo` is now illegal, similar to when appearing in a model.
+The call of `bar` in `foo` is now illegal, similar to using `bar` directly in a model.
 Note that in practice `foo` might contain both something generating event and the call to `bar`.
--- a/RationaleMCP/Changes/3_7/3610.md
+++ b/RationaleMCP/Changes/3_7/3610.md
@@ -6,7 +6,7 @@ Variability of functions with `GenerateEvents`.
 
 ## Reason
 
-Making it possible to create functions that generate events similarly as the built-in functions, and have rules consistent with them just being inlined.
+Making it possible to create functions that generate events similarly to the built-in functions, and have rules consistent with them just being inlined.
 
 ## Breaks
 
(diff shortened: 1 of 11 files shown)


chapters/functions.tex
```diff
@@ -110,7 +110,7 @@ \subsection{Inheritance of Functions}\label{inheritance-of-functions}
 \end{nonnormative}
 
 A special case is defining a \lstinline!function! as a short-class definition with modifiers for inputs inside a model.
-These default values, unless overridden in the function call, will then be considered for variability similarly as if they were given in the function call, see \cref{function-variability}.
+These default values, unless overridden in the function call, will then be considered for variability similarly to if they were given in the function call, see \cref{function-variability}.
```
```suggestion
These default values, unless overridden in the function call, will then be considered for variability as if they were given in the function call, see \cref{function-variability}.
```
This change felt problematic in some way for me. Does this work?

RationaleMCP/Changes/3_7/3610.md
```diff
@@ -26,5 +26,5 @@ end foo;
 Real x=foo(time);
 ```
 
-The call of `bar` in `foo` is now illegal, similarly as in a model.
+The call of `bar` in `foo` is now illegal, similar to when appearing in a model.
```
HansOlsson:
```suggestion
The call of `bar` in `foo` is now illegal, similar to using `bar` directly in a model.
```
I didn't like the sound of "similar to when appearing". Is this clearer?

chapters/functions.tex
henrikt-ma: Sure.

RationaleMCP/Changes/3_7/3610.md
henrikt-ma: Works for me.

chapters/classes.tex
```diff
@@ -1099,7 +1099,7 @@ \section{Balanced Models}\label{balanced-models}
 \end{nonnormative}
 
 \begin{definition}[Globally balanced]\index{globally balanced}\index{balanced!globally}
-Similarly as locally balanced, but including all unknowns and equations from all components.
+Similarly to locally balanced, but including all unknowns and equations from all components.
```
```suggestion
Similar to locally balanced, but including all unknowns and equations from all components.
```
When seeing the other changes, I realized that it seems that it should be "Similar", but we can make it a separate PR otherwise.

HansOlsson (approved):
Looking good. One minor comment, but it can also wait.

PR #3818, opened 2026-01-27
Local balance per type

### LINKED ISSUE #3763: Local balance does not consider type

henrikt-ma:
Fixing #3763.  Opening as draft to make clear that this is not ready for merging, and so that a discussion about how to do this can be started already.

In particular, I am looking for input on which type concepts to use when defining _scalar type class_, that is the formal categorization used to break down the local balance counting based on type.

Note that this PR contains some language fixing commits which have also been isolated in the PR #3817, so the changes of the present PR will look much cleaner once #3817 has been merged.


chapters/classes.tex
```diff
@@ -1025,31 +1025,44 @@ \section{Balanced Models}\label{balanced-models}
...
 \begin{definition}[Local number of unknowns]\index{local number of unknowns}
-The local number of unknowns of a \lstinline!model! or \lstinline!block! class is the sum based on the components:
+Let $T$ be a scalar type class.
+The local number of $T$ unknowns of a \lstinline!model! or \lstinline!block! class is the sum based on the components:
 \begin{itemize}
 \item
-  For each declared component of specialized class \lstinline!type! (\lstinline!Real!, \lstinline!Integer!, \lstinline!String!, \lstinline!Boolean!, enumeration and arrays of those, etc.) or \lstinline!record!, or \lstinline!operator record! not declared as \lstinline!outer!, it is the number of unknown variables inside it (i.e., excluding parameters and constants and counting the elements after expanding all records, operator record, and arrays to a set of scalars of primitive types).
-\item
-  Each declared component of specialized class \lstinline!type! or \lstinline!record! declared as \lstinline!outer! is ignored.
+  For each component declared as \lstinline!outer!, it is zero.
```
HansOlsson: Ignoring all `outer` components would be a major change, as currently it only ignores `outer` for record/type variables. 
Consider MBS-models `outer World world;` currently the count include the normal count for its public connector - matching the connections to them, so it sort of evens out.

(I'm aware that during translation those connections are lifted to the place of the `inner`.)

Thus restore - possible with some further note as below:
```suggestion
  Each declared component of specialized class \lstinline!type! or \lstinline!record! declared as \lstinline!outer! is ignored; other \lstinline!outer! components are handled normally.
```

chapters/classes.tex
```diff
@@ -1099,16 +1112,16 @@ \section{Balanced Models}\label{balanced-models}
 \end{nonnormative}
 
 \begin{definition}[Globally balanced]\index{globally balanced}\index{balanced!globally}
-Similar to locally balanced, but including all unknowns and equations from all components.
-The global number of unknowns is computed by expanding all unknowns (i.e., excluding parameters and constants) into a set of scalars of primitive types.
-This should match the global equation size defined as:
+Similarly to locally balanced, but including all unknowns and equations from all components.
```
```suggestion
Similar to locally balanced, but including all unknowns and equations from all components.
```
Seems that change got lost in merging.

chapters/classes.tex
```diff
@@ -1025,31 +1025,44 @@ \section{Balanced Models}\label{balanced-models}
 The restrictions below apply after flattening -- i.e., inherited components are included -- possibly modified.
 The corresponding restrictions on connectors and connections are in \cref{restrictions-of-connections-and-connectors}.
 
+\begin{definition}[Scalar type class]\index{scalar type class}
+A \emph{scalar type class} refers to an equivalence class of equation-compatible scalar types.
+
+\textbf{TODO: Define equation-compatible, or use one of the existing compatibility definitions!}
```
HansOlsson: I guess that should be cleaned up, right?

chapters/classes.tex
```diff
@@ -1191,7 +1205,7 @@ \section{Balanced Models}\label{balanced-models}
 \end{align*}
 and 2 equations corresponding to the 2 flow variables \lstinline!p.i! and \lstinline!n.i!.
 
-These are 5 equations in 5 unknowns (locally balanced model).
+These are 5 equations in 5 unknowns (locally balanced in \lstinline!Real!).
```
HansOlsson: I find the use of "in" as part "locally balanced in Real" a bit confusing.
I would prefer some other word, but don't know exactly which ones, "locally balanced for Real", or "locally balanced in terms of Real"

HansOlsson (changes requested):
I assume it was intended to be "Ready for Review", after merging the other one, right?

Looks good overall.
However, there are some minor comments.

chapters/classes.tex
henrikt-ma: Yes, I wanted to have a discussion about the approach before proceeding with this detail.

chapters/classes.tex
henrikt-ma: I see.

chapters/classes.tex
henrikt-ma: We can give _for_ a try…

henrikt-ma: I must say I still would prefer _in_ over _for_, so maybe we should keep looking for the right word.  Candidates:
- concerning
- considering

HansOlsson: "Considering" seems the best of them for me too, but looking at a thesaurus I also found "with regard to", "with respect to".

chapters/classes.tex
HansOlsson: Check if it exists, otherwise make separate later PR for it.
Current rules state mixes integer and real in some odd way, instead of having type coercion integer->real.

HansOlsson:
Language group: Will be a clear separation integer/enumeration; so cannot write an equation with inverse for that (example to be given).
In the worst case we would need to merge these types, but hopefully not needed.

henrikt-ma:
> This model currently compiles and simulates in Modelon Impact. But it would not be valid with the proposed type separation, the only unknown variable has type `E` and the only equation has type `Integer`.

SystemModeler does not accept the model based on the following rule in  https://specification.modelica.org/master/modelica-dae-representation.html:
> The key difference between the two groups of discrete-time variables $z$ and $m$ here is how they are determined.
> The interpretation of the solved form of \eqref{eq:dae-discrete-valued} is that given values for everything else, there is a closed-form solution for $m$ in the form of a sequence of assignments to each of the variables of $m$ in turn -- there must be no cyclic dependencies between the equations used to solve for $m$.
> Further, each of the original model equations behind \eqref{eq:dae-discrete-valued} must be given in almost solved form:
> -  Non-\lstinline!Integer! equations at most requiring flipping sides of the equation to obtain the used assignment form.
> …

HansOlsson:
Dymola doesn't accept it either.

MarkusOlssonModelon:
> > This model currently compiles and simulates in Modelon Impact. But it would not be valid with the proposed type separation, the only unknown variable has type `E` and the only equation has type `Integer`.
> 
> SystemModeler does not accept the model based on the following rule in https://specification.modelica.org/master/modelica-dae-representation.html:
> 
> > The key difference between the two groups of discrete-time variables 
> >   z
> >  and 
> >   m
> >  here is how they are determined.
> > The interpretation of the solved form of \eqref{eq:dae-discrete-valued} is that given values for everything else, there is a closed-form solution for 
> >   m
> >  in the form of a sequence of assignments to each of the variables of 
> >   m
> >  in turn – there must be no cyclic dependencies between the equations used to solve for 
> >   m
> > .
> > Further, each of the original model equations behind \eqref{eq:dae-discrete-valued} must be given in almost solved form:
> > 
> > * Non-`Integer` equations at most requiring flipping sides of the equation to obtain the used assignment form.
> >   …

You are right. In that case I have no concerns, should be fine to separate integers/enumerations.

chapters/classes.tex
HansOlsson: Looking at "6.7 Type Compatible Expressions" as far as I understand the current text is working, but I see the following issues:
- That section doesn't explicitly state that it applies to equations (it is instead stated in "8.3.1 Simple Equality Equations")
- It is complicated to find the specific rules for equations, as it also handles comparison operators, division and power.
- Thus the type coercion integer->real is stated twice.
- It is less clear for algorithms and multi-returning functions.

I can see the following minor improvements of that section:
1. State that it applies to equations as well and that the type compatible expression for the equation is the type of the equation; and if necessary add non-normative text giving the exact rules.
2. At the start state the rules use "Real or Integer expression" to explicitly handle the integer->real type coercion.

If we wanted to state that it applies after that coercion we would replace:
 - If A is a Real expression then B must be a Real or Integer expression. The type of the full expression is Real, compare section 10.6.13, unless the operator is a relational operator (section 3.5) where the type of the full expression is Boolean.
- If A is an Integer expression then B must be a Real or Integer expression. For exponentiation and division the type of the full expression is Real (even if both A and B are Integer) see section 10.6.7 and section 10.6.5, for relational operators the type of the full expression is Boolean. In other cases the type of the full expression is Real or Integer (same as B), compare section 10.6.13.

by:
 - If A is a Real expression then B must be a Real expression. The type of the full expression is Real, compare section 10.6.13, unless the operator is a relational operator (section 3.5) where the type of the full expression is Boolean.
- If A is an Integer expression then B must be a Integer expression. For exponentiation and division the type of the full expression is Real (even if both A and B are Integer) see section 10.6.7 and section 10.6.5, for relational operators the type of the full expression is Boolean. In other cases the type of the full expression is Integer, compare section 10.6.13.

I don't think that would be a significant enough improvement, and in particular I see issues when combining it with the next item.

3. Add that for binding equations, algorithms, and multi-returning function calls (?), if the left-hand-side is an Integer variable then the right-hand-side must be an Integer expression, something like:

- If A is an Integer expression (which must be a component reference) as the left-hand-side of an algorithm etc then B must be an Integer expression, and the type of the expression is Integer.
- Otherwise if A ...

I think it is best to discuss that in a separate PR.

HansOlsson: So, for now add something like:
```suggestion
The type of the equation is defined by the type compatible expression for the equation in \cref{type-compatible-expressions}.
```

HansOlsson (approved):
Ok now. Should preferably clean up the type compatible expression part.

PR #3823, opened 2026-01-30
Improve return-statement example

### LINKED ISSUE #3822: Potential Error in Section 12.1.2 Example Note
In the example of Section [12.1.2](https://specification.modelica.org/master/functions.html#function-return-statements), the `return` statement within the `for` algorithm cannot be replaced with a `break` statement; otherwise, the value of `index` will remain 0 indefinitely.

Image

henrikt-ma:
Fixes #3822.

### DIFF
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -85,12 +85,12 @@ return;
 Using \lstinline!return! to terminate evaluation of a function algorithm:
 \begin{lstlisting}[language=modelica]
 function findValue "Find position of value in array, or 0 if not found"
-  input Integer elements[:];
+  input Integer x[:];
   input Integer value;
   output Integer index;
 algorithm
-  for i in 1:size(elements,1) loop
-    if elements[i] == value then
+  for i loop
+    if x[i] == value then
       index := i;
       return;
     end if;
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -82,9 +82,9 @@ return;
 \end{lstlisting}
 
 \begin{example}
-(Note that this could alternatively use \lstinline!break!:)
+Using \lstinline!return! to terminate evaluation of a function algorithm:
 \begin{lstlisting}[language=modelica]
-function findValue "Returns position of val or 0 if not found"
+function findValue "Returns position of value in array"
   input Integer x[:];
   input Integer val;
   output Integer index;
@@ -95,10 +95,10 @@ algorithm
       return;
     end if;
   end for;
-  index := 0;
-  return;
+  assert(false, "Value not found.");
 end findValue;
 \end{lstlisting}
+The same result could be obtained using \lstinline!break!, but then additional measures would be needed to avoid triggering the assert.
 \end{example}
 
 \subsection{Inheritance of Functions}\label{inheritance-of-functions}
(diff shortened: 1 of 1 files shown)


HansOlsson (changes requested):
Too many unneeded changes:

I don't like the change from returning 0 on failure to asserting, since that makes the find-function less useful.
I find the change from "val" to "value" in description as sub-optimal, if we wanted to avoid abbreviations, and not use value to avoid confusion, one possibility would be to name it "key".
And instead of writing "additional measures" it would be clearer to explain them.
Additionally there's a subtle difference - it finds the last instead of first occurrence in that case.

chapters/functions.tex
```diff
@@ -82,9 +82,9 @@ \subsection{Function Return-Statements}\label{function-return-statements}
 \end{lstlisting}
 
 \begin{example}
-(Note that this could alternatively use \lstinline!break!:)
+Using \lstinline!return! to terminate evaluation of a function algorithm:
 \begin{lstlisting}[language=modelica]
-function findValue "Returns position of val or 0 if not found"
+function findValue "Returns position of value in array"
```
```suggestion
function findValue "Returns position of val or 0 if not found"
```
Restore, or rename and use `key`

chapters/functions.tex
```diff
@@ -95,10 +95,10 @@ \subsection{Function Return-Statements}\label{function-return-statements}
       return;
     end if;
   end for;
-  index := 0;
-  return;
+  assert(false, "Value not found.");
```
```suggestion
  index := 0;
  return;
```
Restore.

chapters/functions.tex
```diff
@@ -95,10 +95,10 @@ \subsection{Function Return-Statements}\label{function-return-statements}
...
 end findValue;
 \end{lstlisting}
+The same result could be obtained using \lstinline!break!, but then additional measures would be needed to avoid triggering the assert.
```
```suggestion
The same result could be obtained using \lstinline!break!, if `index:=0;` was moved to the start of the algorithm.
```

chapters/functions.tex
henrikt-ma: But why should we give an example with `return;` at the end of an algorithm?

HansOlsson: > But why should we give an example with `return;` at the end of an algorithm?

Ok, skip that leading to:
```suggestion
  index := 0;
```

chapters/functions.tex
henrikt-ma: This is the weakness of the example; once `index := 0` is at the start of the algorithm, it works pretty much equally well for both `return` and `break`.  The example would be so much better if the function did something which couldn't be done just as easily with `break`.  I looked at the MSL for inspiration, but didn't find much:
- Lots of cases in Spice3, but I don't think that sub-library is representative of how to model with Modelica.
- `BooleanVectors.firstTrueIndex`, which is basically what we have here, but with `index := 0` already at the beginning of the algorithm.  (Moving the `index := 0` to the end of the algorithm could be considered an improvement of the implementation.)
- `Modelica.Fluid.Utilities.regRoot2` makes good use of `return`, but the function is too complicated to serve as a concise example in the specification.

HansOlsson: I agree that it's a weakness in terms of explaining the rationale for having both `break `and `return`.
However, I don't view it as a weakness in terms of explaining the semantics of `return`.

chapters/functions.tex
henrikt-ma: I'd say `key` is for maps.  I have restored the name `x`.  The main point was to avoid the non-word "val" in the description string.

HansOlsson (approved):

Issue #3816, opened 2026-01-26
Allow equality relations among Reals for parameter?


casella:
[MLS Section 3.5](https://specification.modelica.org/maint/3.6/operators-and-expressions.html#equality-relational-and-logical-operators) states

> In relations of the form `v1 == v2` or `v1 <> v2`, `v1 or v2` **shall**, unless used in a function, not be a subtype of Real. 
>
> [The reason for this rule is that relations with Real arguments are transformed to state events (see [section 8.5](https://specification.modelica.org/master/equations.html#events-and-synchronization)) and this transformation becomes unnecessarily complicated for the == and <> relational operators (e.g., two crossing functions instead of one crossing function needed, epsilon strategy needed even at event instants). Furthermore, testing on equality of Real variables is questionable on machines where the number length in registers is different to number length in main memory.]

"Shall" looks pretty prescriptive to me. Yet, if I try this model on Dymola and OpenModelica (not sure about other tools)

```modelica
model CheckInequality
  Real x(start = 0, fixed = true);
  Real y;
equation 
  der(x) = 0;
  y = if x == 0 then 2 else 4;
end CheckInequality;
```
I get some warning along the lines that "Hey, you shoulnd't really do that, but we'll compile your model anyway" and then the model is compiled. In this case it actually behaves correctly, but that's by pure chance.

Here, we have a typical scenario of a MLS rule that has been ignored for ages (at least by some tools), so that people write models ignoring it and then we get into all kind of trouble.

My first question is: should this rule really be enforced, even if it means breaking code that actually runs? If the answer is no, we should rewrite it as

> In relations of the form `v1 == v2` or `v1 <> v2`, `v1 or v2` **should**, unless used in a function, not be a subtype of Real. 
>
> [If you use such constructs, you may end up having numerically fragile simulation code]

If the answer is yes, then we should really mean it.

Comments?

HansOlsson:
It would be possible to slightly decouple the aspects by considering `noEvent` and parameter expression as follows:

```
model CheckInequality
  Real x(start = 0, fixed = true);
  Real y;
  parameter Real denom=2.0;
  final parameter Real p=1.0/denom;
  final parameter Boolean b = p*denom==1;
equation 
  der(x) = 0;
  y = noEvent(if x == 0 then 2 else 4);
end CheckInequality;
```
Note that the check above will fail for some values of the parameter `denom `like 98.
I don't see you could decouple it in the other way (generate events without having strict equality).

[...]

PR #3825, opened 2026-02-04
Generate events equality

### LINKED ISSUE #3824: GenerateEvents and Real equality testing
Our new and shiny semantics for `GenerateEvents = true` seems problematic in the following example:
```
model RealEqualityInsideFunction
  function f
    input Real u;
    output Boolean y = u == 0.5;
  end f;

  Boolean b = f(time);
end RealEqualityInsideFunction;
```

As I read the specification, this shouldn't be rejected:
- The equality testing on `Real` operands happens inside a function, so it is allowed.
- The variability of the function call is discrete-time, so `b = f(time)` fulfills variability requirements.

According to the old rules, `b = f(time)` would be rejected, so it didn't seem like a problem if the `u == 0.5` was treated without generating events.  Now, however, with the model being accepted, it looks as we have accidentally introduced the need to handle event-generating equality testing on `Real` operands.  I suppose this was not by intention?

HansOlsson:
Closes #3824
It also explains the special handling of functions, as discussed ni #3816

### DIFF
--- a/chapters/operatorsandexpressions.tex
+++ b/chapters/operatorsandexpressions.tex
@@ -284,6 +284,8 @@ The following holds for relational operators:
   \begin{nonnormative}
   The reason for this rule is that relations with \lstinline!Real! arguments are transformed to state events (see \cref{events-and-synchronization}) and this transformation becomes unnecessarily complicated for the \lstinline!==! and \lstinline!<>! relational operators (e.g., two crossing functions instead of one crossing function needed, epsilon strategy needed even at event instants).
   Furthermore, testing on equality of \lstinline!Real! variables is questionable on machines where the number length in registers is different to number length in main memory.
+  The handling in normal functions is similar to the handling in external functions.
+  Functions with \lstinline!GenerateEvents = true! can be treated as being directly inlined, and thus follow the rules for non-functions.
   \end{nonnormative}
 \item
   Relational operators can generate events, see \cref{discrete-time-expressions}.
--- a/chapters/operatorsandexpressions.tex
+++ b/chapters/operatorsandexpressions.tex
@@ -280,7 +280,7 @@ The following holds for relational operators:
 \item
   For operands of enumeration types, the order is given by the order of declaration of the enumeration literals.
 \item
-  In relations of the form \lstinline!v1 == v2 or v1 <> v2!, \lstinline!v1! or \lstinline!v2! shall, unless used in a function, not be a subtype of \lstinline!Real!.
+  In relations of the form \lstinline!v1 == v2 or v1 <> v2!, \lstinline!v1! or \lstinline!v2! shall, unless used in a function without \lstinline!GenerateEvents = true! (see \cref{modelica:GenerateEvents}), not be a subtype of \lstinline!Real!.
   \begin{nonnormative}
   The reason for this rule is that relations with \lstinline!Real! arguments are transformed to state events (see \cref{events-and-synchronization}) and this transformation becomes unnecessarily complicated for the \lstinline!==! and \lstinline!<>! relational operators (e.g., two crossing functions instead of one crossing function needed, epsilon strategy needed even at event instants).
   Furthermore, testing on equality of \lstinline!Real! variables is questionable on machines where the number length in registers is different to number length in main memory.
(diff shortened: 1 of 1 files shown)


chapters/operatorsandexpressions.tex
```diff
@@ -280,10 +280,12 @@ \section{Equality, Relational, and Logical Operators}\label{equality-relational-
 \item
   For operands of enumeration types, the order is given by the order of declaration of the enumeration literals.
 \item
-  In relations of the form \lstinline!v1 == v2 or v1 <> v2!, \lstinline!v1! or \lstinline!v2! shall, unless used in a function, not be a subtype of \lstinline!Real!.
+  In relations of the form \lstinline!v1 == v2 or v1 <> v2!, \lstinline!v1! or \lstinline!v2! shall, unless used in a function without \lstinline!GenerateEvents = true! (see \cref{modelica:GenerateEvents}), not be a subtype of \lstinline!Real!.
```
henrikt-ma: This has to be an old formatting mistake that we should fix now:

```suggestion
  In relations of the form \lstinline!v1 == v2! or \lstinline!v1 <> v2!, \lstinline!v1! or \lstinline!v2! shall, unless used in a function without \lstinline!GenerateEvents = true! (see \cref{modelica:GenerateEvents}), not be a subtype of \lstinline!Real!.
```

chapters/operatorsandexpressions.tex
```diff
@@ -280,10 +280,12 @@ \section{Equality, Relational, and Logical Operators}\label{equality-relational-
...
   \begin{nonnormative}
   The reason for this rule is that relations with \lstinline!Real! arguments are transformed to state events (see \cref{events-and-synchronization}) and this transformation becomes unnecessarily complicated for the \lstinline!==! and \lstinline!<>! relational operators (e.g., two crossing functions instead of one crossing function needed, epsilon strategy needed even at event instants).
   Furthermore, testing on equality of \lstinline!Real! variables is questionable on machines where the number length in registers is different to number length in main memory.
+  The handling in normal functions is similar to the handling in external functions.
+  Functions with \lstinline!GenerateEvents = true! can be treated as being directly inlined, and thus follow the rules for non-functions.
```
henrikt-ma: Can we try to avoid _non-functions_:

```suggestion
  Functions with \lstinline!GenerateEvents = true! can be treated as being directly inlined, and thus follow the same rules that apply outside functions.
```

chapters/operatorsandexpressions.tex
```diff
@@ -280,10 +280,12 @@ \section{Equality, Relational, and Logical Operators}\label{equality-relational-
...
 \item
-  In relations of the form \lstinline!v1 == v2 or v1 <> v2!, \lstinline!v1! or \lstinline!v2! shall, unless used in a function, not be a subtype of \lstinline!Real!.
+  In relations of the form \lstinline!v1 == v2 or v1 <> v2!, \lstinline!v1! or \lstinline!v2! shall, unless used in a function without \lstinline!GenerateEvents = true! (see \cref{modelica:GenerateEvents}), not be a subtype of \lstinline!Real!.
   \begin{nonnormative}
   The reason for this rule is that relations with \lstinline!Real! arguments are transformed to state events (see \cref{events-and-synchronization}) and this transformation becomes unnecessarily complicated for the \lstinline!==! and \lstinline!<>! relational operators (e.g., two crossing functions instead of one crossing function needed, epsilon strategy needed even at event instants).
   Furthermore, testing on equality of \lstinline!Real! variables is questionable on machines where the number length in registers is different to number length in main memory.
+  The handling in normal functions is similar to the handling in external functions.
```
henrikt-ma: I can't see that _normal functions_ is an established term.  If we pretend that `GenerateEvents` would not be allowed for external function classes, we could use that instead:

```suggestion
  Allowing the comparisons in functions without \lstinline!GenerateEvents = true! does not introduce any kind of function behavior that an external function would not be able to express anyway.
```

henrikt-ma (commented):
Basically good, just needs some iteration on how to phrase it.

HansOlsson:
Language group: Without this rule we get discrete-time Boolean based on equality testing. No objections to change.

henrikt-ma (approved):
Looks good and appears to have support in the language group as well.

PR #3827, opened 2026-02-06
Start input

### LINKED ISSUE #3821: Clarification about start values for inputs
Lets consider the following minimal model

```
model undefinedInput01 "undefinedInput01.mo"
 input Real a(start=50);
equation 
  assert(abs(a-50)<1e-5, "error");
end undefinedInput01;
```

Does the specification ensure, that `a` is actually evaluated to `50`? In OpenModelica it is done as such and the LSGreenH2Production (OpenScaling project) relies on that. From what I found, this seems to be more a recommendation (if at all?).

> **Section 4.4.2.2 Prefix Rules:**
> 
> In simulation models and blocks ...   Especially, the input prefix defines that values for such a variable have to be provided from the simulation environment and the output prefix defines that...

> **4.8.6 Attributes start, fixed, nominal, and unbounded**
> ...
> The attributes start and fixed define the initial conditions for a variable. fixed = false means an initial guess, i.e., value may be changed by static analyzer. fixed = true means a required value

> 
> **8.6 Initialization, initial equation, and initial algorithm**
> For parameters, fixed defaults to true. For other variables, fixed defaults to false.
> start-values of variables having fixed = false can be used as initial guesses, in case iterative solvers are used in the initialization phase.

Even if the initialization is regarded as simulation environment to provide the value for `a`, the last sentence indicates that it is only a guess for the case of iterative solvers?

HansOlsson:
Closes #3821 
I'm not fully satisfied with the "simulation environment" part, but I didn't want to get too deep in the FMI (and other environments) here.


chapters/classes.tex
```diff
@@ -248,7 +248,9 @@ \subsubsection{Prefix Rules}\label{prefix-rules}
...
 \item
   In \emph{simulation} \emph{models} and \emph{blocks} (i.e., on the top level of a model or block that shall be simulated), these prefixes define the interaction with the environment where the simulation model or block is used.
-  Especially, the \lstinline!input! prefix defines that values for such a variable have to be provided from the simulation environment and the \lstinline!output! prefix defines that the values of the corresponding variable can be directly utilized in the simulation environment, see the notion of \emph{globally balanced} in \cref{balanced-models}.
+  Especially, the \lstinline!input! prefix defines that values for such a variable can be provided from the simulation environment (otherwise using the start-value or fallback-value cref{def:fallback-value}) and the \lstinline!output! prefix defines that the values of the corresponding variable can be directly utilized in the simulation environment, see the notion of \emph{globally balanced} in \cref{balanced-models}.
```
JohannesKrauseGitH: The "or" sounds a bit as if both options are fine. From my understanding, the start value shall be used and the fallback value only, if there is no start value? Maybe we can make that little bit more explicit?

HansOlsson: True, ~~will do~~ have done

```suggestion
  Especially, the \lstinline!input! prefix defines that values for such a variable can be provided from the simulation environment (otherwise using the start-value, or when it is missing falling back to fallback-value cref{def:fallback-value}) and the \lstinline!output! prefix defines that the values of the corresponding variable can be directly utilized in the simulation environment, see the notion of \emph{globally balanced} in \cref{balanced-models}.
```

chapters/classes.tex
```diff
@@ -248,7 +248,9 @@ \subsubsection{Prefix Rules}\label{prefix-rules}
...
 \item
   In \emph{simulation} \emph{models} and \emph{blocks} (i.e., on the top level of a model or block that shall be simulated), these prefixes define the interaction with the environment where the simulation model or block is used.
-  Especially, the \lstinline!input! prefix defines that values for such a variable have to be provided from the simulation environment and the \lstinline!output! prefix defines that the values of the corresponding variable can be directly utilized in the simulation environment, see the notion of \emph{globally balanced} in \cref{balanced-models}.
+  Especially, the \lstinline!input! prefix defines that values for such a variable can be provided from the simulation environment (otherwise using the start-value, or when it is missing falling back to fallback-value cref{def:fallback-value}) and the \lstinline!output! prefix defines that the values of the corresponding variable can be directly utilized in the simulation environment, see the notion of \emph{globally balanced} in \cref{balanced-models}.
```
henrikt-ma: Suggesting various improvements:
- Missing `\` in `\cref`.
- Correct way to write "start value" (looks like we still have some general cleanup to do here).
- We always write _fallback value_ with a space.
- There is already a general rule which makes the fallback value kick in when `start` is not provided, so it is enough to talk about it in the next sentence, making this sentence a bit easier to digest.
- I think the use of `start` deserves a more prominent presentation than just being mentioned in a parenthesis.

```suggestion
  Especially, the \lstinline!input! prefix defines that values for such a variable can be provided from the simulation environment, and the \lstinline!output! prefix defines that the values of the corresponding variable can be directly utilized in the simulation environment, see the notion of \emph{globally balanced} in \cref{balanced-models}.
  When the simulation environment does not provide a value for an input variable, the variable's \lstinline!start!-attribute is used instead.
```

chapters/classes.tex
```diff
@@ -248,7 +248,9 @@ \subsubsection{Prefix Rules}\label{prefix-rules}
...
+  When using a fallback-value for an input of the simulation environment it is recommended to not give any diagnostic during translation, but only during simulation provided it is actually used.
```
henrikt-ma: Second part of the change above:

```suggestion
  In case there is no \lstinline!start!-attribute, the recommended diagnostic for using the fallback value (see \cref{def:fallback-value}) should not be given during translation, but only during simulation provided it is actually used.
```

henrikt-ma (approved):
In agreement with language group decision in https://github.com/modelica/ModelicaSpecification/issues/3821#issuecomment-3854359127 (with the details given in https://github.com/modelica/ModelicaSpecification/issues/3821#issuecomment-3852334514).

PR #3828, opened 2026-02-06
Limit arrays

### LINKED ISSUE #3820: Modelica array class
I have been consistently confused about array class in Modelica. If such an array class was used as the type specifier of a component, the MLS has already provided an explanation. However:

1. If base classes (such as `function`, `record`, or enumeration class) can be called, can array classes also be called? If yes, should the input and output types have an extra dimension?
```Modelica
model ArrayCallable
  record R
    Real x = 0;
  end R;
  record S = R[2](x = 1:2);
  S s = S();  // Illegal ?
  
  type E = enumeration(a, b);
  type G = E[2](quantity = {"a", "b"});
  G g = G(1); // Illegal ?
end ArrayCallable;
```

2. Can array classes be used in composite names to access members of the class? For example, an array `package`.
```Modelica
model ArrayPackage
  package P
    constant Real c = 0;
  end P;
  package Q = P[2](c = 1:2);
  Real x = Q.c;  // Is c an element of Q?
end ArrayPackage;
```

In C++, there is a also concept of array types. The constructor for such types works by sequentially constructing each array element and is unrelated to the constructor of the base class. Moreover, it cannot access types or static variables from the base class. Is the handling in Modelica the same, or will there be clear regulations on this in the future?

HansOlsson:
Closes #3820 
I think this covers all of the discussed cases.

As far as I can see it is also consistent with the previous intent, but just missed:
- Enumeration constructor is part of list that can be called vectorized.
- Record cast (but not record constructor) explicitly listed that it could be called vectorized
- Lookup of classes in components restricted it to scalar components; and lookup for classes is described as constructing a component of the class.

### DIFF
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -833,7 +833,7 @@ There are basically four groups of built-in functions in Modelica:
 
 \section{Record Constructor Functions}\label{record-constructor-functions}
 
-Whenever a record is defined, a record constructor function with the same name and in the same scope as the record class is implicitly defined according to the following rules:
+Whenever a scalar record is defined, a record constructor function with the same name and in the same scope as the record class is implicitly defined according to the following rules:
 
 % Warning: 'partial flatteing' doesn't seem to be defined.  See 'partial instantitation'.
 The declaration of the record is partially flattened including inheritance, modifications, redeclarations, and expansion of all names referring to declarations outside of the scope of the record to their fully qualified names.
@@ -856,6 +856,7 @@ An instance of the record is declared as output parameter using a name not appea
 In the modification, all input parameters are used to set the corresponding record variables.
 
 A record constructor can only be called if the referenced record class is found in the global scope, and thus cannot be modified.
+The record constructor can be used in vectorized form according to \cref{scalar-functions-applied-to-array-arguments}.
 
 \begin{nonnormative}
 This allows constructing an instance of a record, with an optional modification, at all places where a function call is allowed.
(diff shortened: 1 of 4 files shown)


chapters/functions.tex
```diff
@@ -856,6 +856,7 @@ \section{Record Constructor Functions}\label{record-constructor-functions}
 In the modification, all input parameters are used to set the corresponding record variables.
 
 A record constructor can only be called if the referenced record class is found in the global scope, and thus cannot be modified.
+The record constructor can be used in vectorized form according to \cref{scalar-functions-applied-to-array-arguments}.
```
qlambert-pro, line 859: We need to address the record constructor of operator record. Since nothing is said of the conflict it could create.

chapters/functions.tex
```diff
@@ -833,7 +833,7 @@ \section{Built-in Functions}\label{built-in-functions}
 
 \section{Record Constructor Functions}\label{record-constructor-functions}
 
-Whenever a record is defined, a record constructor function with the same name and in the same scope as the record class is implicitly defined according to the following rules:
+Whenever a scalar record is defined, a record constructor function with the same name and in the same scope as the record class is implicitly defined according to the following rules:
```
henrikt-ma, line 836: Just asking: Is there a risk of confusion when speaking of a _scalar record_?  Could it be interpreted as a record just a single member of "size one"?

henrikt-ma (commented):
I am leaving it for @qlambert-pro to provide a review with approval.

chapters/functions.tex
HansOlsson: I doubt that, especially as it doesn't work with the rest of the text.

chapters/functions.tex
HansOlsson: An operator record shouldn't have an automatically generated record constructor, as that would cause conflicts with the defined operator record constructors, so their shouldn't be any conflict.

HansOlsson:
Pinging @qlambert-pro

qlambert-pro (approved):
lgtm

PR #3831, opened 2026-02-10
New LTS Ubuntu variant.

### LINKED ISSUE #3830: Out of date CI docker image base
Looks like the docker image used for the CI is based on Ubuntu Bionic, which went out of support in 2023:

https://github.com/modelica/ModelicaSpecification/blob/a4a6eb286be54d3bd5fe3e29d323be923143e4ad/.CI/latexml/Dockerfile#L1

https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions

Maybe it would be a good idea to consider using Noble Numbat, 24.04 LTS?

HansOlsson:
Closes #3830

Does this suffice or is there a need to update more on the server?

### DIFF
--- a/.CI/Jenkinsfile
+++ b/.CI/Jenkinsfile
@@ -18,7 +18,7 @@ pipeline {
   stage('build') {
     agent {
       docker {
-        image 'modelicaspec/latexml:20260210'
+        image 'modelicaspec/latexml:20260212'
         label 'linux'
         alwaysPull true
       }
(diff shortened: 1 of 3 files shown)


henrikt-ma:
LaTeXML 0.8.8 is the version I use locally, and while I haven't noticed any issues I must admit that I use the PDF document less and less in my daily work.  (The main reason why I am using the HTML build instead is that I very often need to give document links to others.)

sjoelund (approved):

PR #3833, opened 2026-02-12
Small cleanup of rules for input/output prefixes

henrikt-ma:
This fixes some stylistic issues I noted when reviewing #3827.


HansOlsson (approved):
Looks good.
I noticed that one case is missing (as before): input/output for connector components in model/block components; they should be treated as input/output for the connectors.
Likely worth a separate PR.

PR #3834, opened 2026-02-12
Workaround for LaTeXML issue with the babel package

henrikt-ma:
This is a fix for the issue noted in https://github.com/modelica/ModelicaSpecification/pull/3831#issuecomment-3889794159.

### DIFF
--- a/preamble.tex
+++ b/preamble.tex
@@ -3,7 +3,11 @@
 % change the layout that can wait
 \hfuzz=50pt
 
-\usepackage[english]{babel}
+% Work around LaTeXML issue with the babel package:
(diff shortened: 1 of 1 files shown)


HansOlsson (approved):
Ok, the result looks ok, and it doesn't make sense to bring in an entire package for just that.

henrikt-ma:
> So babel was only used to get access to `\alsoname` used for `\seealso`?

Not sure, but providing `\alsoname` was sufficient to make the PDF build work without babel.  Maybe babel is also making sure that other terms are not defined based on some crazy locale, but this would probably take some time to discover…

PR #3838, opened 2026-02-19
Derivative in left-hand-side

### LINKED ISSUE #3764: Incorrect definition of statement within the Modelica grammar
I have recently been implementing a Modelica parser and ran into an issue with the grammar as a result of the definition of `statement` within the grammar.  The rule is defined as:

```
statement :
  ( component-reference ( ":=" expression | function-call-args )
    | "(" output-expression-list ")" ":="
      component-reference function-call-args
    | break
    | return
    | if-statement
    | for-statement
    | while-statement
    | when-statement
  )
  description
```

However, this then causes parsing to fail on statements such as:
`der(a) := expression;`

This is because `component-reference` does not allow `der(a)` on the left hand side of the assignment.

The problem can be fixed by replacing `component-reference` with `result-reference` in the statement definition. `result-reference` is defined within the grammar but does not appear in any of the rules and hence it is never used within the grammar.

HansOlsson:
As previously discussed.
Closes #3764

### DIFF
--- a/chapters/statements.tex
+++ b/chapters/statements.tex
@@ -90,6 +90,7 @@ The syntax of statements is as follows:
 \begin{lstlisting}[language=grammar]
 statement :
    ( component-reference ( ":=" expression | function-call-args )
+     | der "(" component-reference ")" ":=" expression
      | "(" output-expression-list ")" ":="
        component-reference function-call-args
      | break
@@ -106,11 +107,12 @@ statement :
 
 The syntax of \firstuse[assignment statement!simple]{simple assignment statement} is as follows:
 \begin{lstlisting}[language=grammar]
-component-reference ":=" expression
+( component-reference | der "(" component-reference ")" ) ":=" expression
 \end{lstlisting}
 
 The \lstinline[language=grammar]!expression! is evaluated.
-The resulting value is stored into the variable denoted by \lstinline[language=grammar]!component-reference!.
+The resulting value is stored into the variable denoted by \lstinline[language=grammar]!component-reference! (or as its derivative).
+An assignment to a derivative may only occur in a model or block.
 
 The \lstinline!expression! must not have higher variability than the assigned component, see \cref{variability-of-expressions}.
 
@@ -119,7 +121,7 @@ Assignment to array variables with subscripts is described in \cref{array-indexi
 \subsubsection{Assignments from Called Functions with Multiple Results}\label{assignments-from-called-functions-with-multiple-results}
 
 There is a special form of assignment statement that is used only when the right-hand side contains a call to a function with multiple results.
-The left-hand side contains a parenthesized, comma-separated list of variables receiving the results from the function call.
+The left-hand side contains a parenthesized, comma-separated list of variables (or derivatives) receiving the results from the function call.
 A function with $n$ results needs $m \leq n$ receiving variables on the left-hand side, and the variables are assigned from left to right.
 
 \begin{lstlisting}[language=modelica]
--- a/chapters/statements.tex
+++ b/chapters/statements.tex
@@ -91,10 +91,8 @@ The syntax of statements is given by the production \lstinline[language=grammar]
 \subsection{Simple Assignment Statements}\label{simple-assignment-statements}
 
 A \firstuse[assignment statement!simple]{simple assignment statement} is an assignment statement where the left-hand side of the \lstinline!:=! is a component reference or a derivative.
+(The restriction that \lstinline!der! is not allowed inside \lstinline!function! classes also applies to assignments.)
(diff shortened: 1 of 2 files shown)


chapters/syntax.tex
```diff
@@ -295,6 +295,7 @@ \subsection{Equations}\label{equations1}
 statement-or-procedure :
    component-reference ":=" expression
    | component-reference function-call-args
+   | der "(" component-reference ")" ":=" expression
```
HansOlsson, line 298: @HansOlsson Reorder to have assignments next to each other (function-call-args first)

chapters/statements.tex
```diff
@@ -90,6 +90,7 @@ \section{Statements}\label{statements}
 \begin{lstlisting}[language=grammar]
 statement :
    ( component-reference ( ":=" expression | function-call-args )
+     | der "(" component-reference ")" ":=" expression
      | "(" output-expression-list ")" ":="
        component-reference function-call-args
```
henrikt-ma: Align with how it is written in the grammar, or take the opportunity to replace outdated duplicated grammar with cross reference to `\cref{equations1}`.

More generally, I would actually be in favor of making each production rule in _syntax.tex_ its own listing, with a systematic way of creating a cross reference to a specific production rule with something like `\cref{production:statement}`.  With precise cross references like that, the need to duplicate the production in other places of the document is more or less eliminated.

chapters/statements.tex
```diff
@@ -106,11 +107,12 @@ \subsection{Simple Assignment Statements}\label{simple-assignment-statements}
 
 The syntax of \firstuse[assignment statement!simple]{simple assignment statement} is as follows:
 \begin{lstlisting}[language=grammar]
-component-reference ":=" expression
+( component-reference | der "(" component-reference ")" ) ":=" expression
```
henrikt-ma: Without "repeating" the grammar:
```
A \firstuse[assignment statement!simple]{simple assignment statement} is an assignment statement where the left-hand side of the \lstinline!:=! is a component reference or a derivative.
(The restriction that \lstinline!der! is not allowed inside \lstinline!function! classes also applies to assignments.)
The right-hand side of the \lstinline!:=! is evaluated, and the resulting value is stored into the left-hand side variable or derivative.
```

chapters/statements.tex
HansOlsson: I added the simple reference; a more precise cross-referencing would be good, but I think that should be a separate PR and prioritized.

chapters/statements.tex
HansOlsson: I copied that without the middle-line, and some update of the next paragraph (that already contained the restriction on derivatives.)

chapters/statements.tex
henrikt-ma: OK. Resolving.

chapters/statements.tex
henrikt-ma: Sure. The point I was trying to make was that we already have the rule forbidding `der` inside functions, so there should be no need for yet another normative rule saying that assignment to `der` is not allowed inside functions.  No big deal.

Edit: Also, I found it more natural to say this right here where the form of a _simple assignment statement_ is described, rather than as a semantic restriction given later.

henrikt-ma: Actually, let me restart this in a new conversation. Resolving.

chapters/statements.tex
```diff
@@ -86,31 +86,15 @@ \section{Statements}\label{statements}
...
-\begin{lstlisting}[language=grammar]
-component-reference ":=" expression
-\end{lstlisting}
+A \firstuse[assignment statement!simple]{simple assignment statement} is an assignment statement where the left-hand side of the \lstinline!:=! is a component reference or a derivative.
+The right-hand side of the \lstinline!:=! is evaluated, and the resulting value is stored into the left-hand side variable or derivative.
 
-The \lstinline[language=grammar]!expression! is evaluated.
-The resulting value is stored into the variable denoted by \lstinline[language=grammar]!component-reference!.
+The resulting value is stored into the variable denoted by \lstinline[language=grammar]!component-reference! (or as its derivative).
+An assignment to a derivative may only occur in a model or block.
```
henrikt-ma: We are repeating ourselves too much regarding the storage of the result.  Also, the paragraph break in the middle is strange.  This is avoided with this good old suggestion:

```suggestion
(The restriction that \lstinline!der! is not allowed inside \lstinline!function! classes also applies to assignments.)
The right-hand side of the \lstinline!:=! is evaluated, and the resulting value is stored into the left-hand side variable or derivative.
```

henrikt-ma (changes requested):
Need to address repeated explanation and strange paragraph break.

henrikt-ma (approved):
Looks good.

Issue #3811, opened 2026-01-19
Various places where parameter variability seems too high


henrikt-ma:
This issue collects a number of items questioning the use of parameter variability, when evaluable variability seems to make more sense.  (The issue was originally formulated just for `smooth`, but was generalized in order to not result in too many, too similar, issues.)

I haven't analyzed all of these carefully, but I find it better to err on the side of bringing up too many on the table than bringing up too few, and leave it for the group to determine which cases we shouldn't touch.

## `smooth`

I believe that the purpose of the `smooth` operator, https://specification.modelica.org/master/operators-and-expressions.html#modelica:smooth, is to allow tools to use the smoothness information during symbolic processing.  Hence, I believe it is a mistake that it is currently not required that the smoothness order is evaluable.

Strictly speaking, also a non-evaluable smoothness order could have been used for integration method order selection, but this appears far more complicated to implement than what I believe is the intention behind the design.

## `cat`

The dimension $k$ along which `cat` should operate, https://specification.modelica.org/master/arrays.html#array-concatenation.

## Component condition

The introduction to the section needs to use correct variability when talking about the component condition, or not talk about its variability at all: https://specification.modelica.org/master/class-predefined-types-and-declarations.html#component-declarations

Just saying this might be good enough:
> an optional `Boolean` condition

## Array dimensions example

From the array dimensions example in https://specification.modelica.org/master/class-predefined-types-and-declarations.html#component-declaration-static-semantics:
```
  parameter Integer n = 0;
  Real z[n*2] = cat(1, ones(n), zeros(n));// Parameter expressions are allowed
```

## `Connections.root`

Since connection handling happens during translation, non-evaluable variability in the non-normative remark seems wrong: https://specification.modelica.org/master/connectors-and-connections.html#modelica:Connections.root

Also note the incorrect grammar in _a parameter expressions_.

## Equations within `when`-equations

Maybe we already have an open issue about this, but in case we don't: https://specification.modelica.org/master/equations.html#equations-within-when-equations

This sounds wrong:
> unless all switching conditions of the `if`-equation are parameter expressions

## Array dimensions in functions as specialized class

From https://specification.modelica.org/master/functions.html#function-as-a-specialized-class:
> or given by constant or parameter expressions

## Subscripts inside `pre`

Restriction on the subscripts inside `pre(…)`: https://specification.modelica.org/master/operators-and-expressions.html#modelica:pre

## Restrictions for transitions

[...]

PR #3839, opened 2026-02-19
Further clarify parameter and evaluable for synchronous.

HansOlsson:
Clarification triggered by #3811 - but not part of original text.

Note that there's one special case that I haven't included, as I couldn't find a good text for it.

The special case is if the interval is zero for a non-evaluated rational clock - to me that should be handled as if it were a non-parameter expression that is zero, i.e., just an error, as it was intended to be inferred but there was nothing to infer it from.


chapters/synchronous.tex
```diff
@@ -277,7 +277,7 @@ \subsection{Argument Restrictions (Component Expression)}\label{argument-restric
 The named argument \lstinline!factor! of \lstinline!subSample! is restricted to be an evaluable expression.
 \begin{lstlisting}[language=modelica]
 Real u;
-parameter Real p=3;
+parameter Real p=3 annotation(Evaluate=true);
```
HansOlsson:
```suggestion
parameter Real p=3;
```

henrikt-ma (approved):
The change seems small and safe, so even though it would be nice with a review by an expert on this chapter, I am providing an approval in case no expert shows up.

PR #3841, opened 2026-02-19
Define when-clause clearly

### LINKED ISSUE #3840: Clarify meaning of when-clause
```
when expression then
  { some-equation ";" }
{ elsewhen expression then
  { some-equation ";" }
}
end when ";" 
```
and then adds 

- etc. etc.

casella:
Fixes #3840

### DIFF
--- a/chapters/equations.tex
+++ b/chapters/equations.tex
@@ -201,7 +201,7 @@ The second condition provides the generalization of \eqref{eq:dae-discrete-value
 
 \subsection{When-Equations}\label{when-equations}
 
-The \lstinline!when!-equations\index{when@\robustinline{when}!equation}\index{then@\robustinline{then}!when-equation@\robustinline{when}-equation}\index{elsewhen@\robustinline{elsewhen}!when-equation@\robustinline{when}-equation} have the following syntax:
+The \lstinline!when!-equations\index{when@\robustinline{when}!equation}\index{then@\robustinline{then}!when-equation@\robustinline{when}-equation}\index{elsewhen@\robustinline{elsewhen}!when-equation@\robustinline{when}-equation}, also known as \lstinline!when!-clauses, have the following syntax:
 \begin{lstlisting}[language=grammar]
 when expression then
   { some-equation ";" }


HansOlsson:
Good, but not complete.

The when-clause can be traced back to version 1.1 of the specification that in the grammar had when_clause_e (for when-equation) and when_clause_a (for when-algorithm). Thus (non-clocked) when-clause almost certainly is intended to cover  both (when applicable).

Section 8.5 even implicitly states this:
> All equations and _assignment statements_ within when-clauses and all assignment statements within
function classes are implicitly treated with noEvent, i.e., relations within the scope of these operators never induce state or time events...

HansOlsson (changes requested):
It should be clear that when-clauses also include when-statements in some way.

chapters/equations.tex
```diff
@@ -201,7 +201,7 @@ \subsection{If-Equations}\label{if-equations}
 
 \subsection{When-Equations}\label{when-equations}
 
-The \lstinline!when!-equations\index{when@\robustinline{when}!equation}\index{then@\robustinline{then}!when-equation@\robustinline{when}-equation}\index{elsewhen@\robustinline{elsewhen}!when-equation@\robustinline{when}-equation} have the following syntax:
+The \lstinline!when!-equations\index{when@\robustinline{when}!equation}\index{then@\robustinline{then}!when-equation@\robustinline{when}-equation}\index{elsewhen@\robustinline{elsewhen}!when-equation@\robustinline{when}-equation}, also known as \lstinline!when!-clauses, have the following syntax:
```
HansOlsson, line 204:
```suggestion
When-clauses\index{when@\robustinline{when}!clause} refer to both \lstinline!when!-equations and \lstinline!when!-statements, \cref{when-statements}.
The \lstinline!when!-equations\index{when@\robustinline{when}!equation}\index{then@\robustinline{then}!when-equation@\robustinline{when}-equation}\index{elsewhen@\robustinline{elsewhen}!when-equation@\robustinline{when}-equation} have the following syntax:
```
I know that it should preferably be rewritten, but at least this ensures that both when-equations and when-statements are mentioned.

HansOlsson: With rewritten I primarily mean that it should be placed somewhere else (in this section, or possible some other one), but I cannot easily do that as a comment.

PR #3845, opened 2026-02-27
Improve connection line annotation

### LINKED ISSUE #3832: Revisit deprecations for connection line Text annotations?
When working on the `Text` annotation for labels on connection lines, https://specification.modelica.org/master/annotations.html#S9.SS4.p3, we have encountered som surprising deprecations in the specification, and we would like to propose that the deprecations are revisited.

### Zero vertical extent

This is the main deprecation we would like to undo:
> Having a zero size for the extent is deprecated and is handled as if upper part is moved up an appropriate amount.

When `fontSize` is specified, _not_ specifying a height makes more sense than specifying it . Can anyone explain why this was deprecated?

To neither specify a (positive) `fontSize` or a positive vertical extent could be a an excellent way to leave the choice of size to the tool, enabling a consistent look of connection line labels across libraries.

Also, I can't see any deprecation of zero vertical extent for the "regular" `Text` annotation in https://specification.modelica.org/master/annotations.html#S9.SS5.SSS5.p1.  It would make sense if the two `Text` annotations didn't differ in this regard.

As in many other cases of deprecation, we have noticed that not even the MSL has moved away from the deprecated semantics, which we consider another good reason to undo the deprecation.  For example, `FullRobot` is full of these:
```
extent={{-6,3},{-6,3}})
```
In fact, all occurrences of `Text` for connection lines that I found with a quick'n dirty regexp search had 0 by 0 extent.

### Automatic horizontal alignment

I'd also like to bring up the deprecation of this default:
```
  TextAlignment horizontalAlignment =
    if index < 0 then TextAlignment.Right else TextAligment.Left "Note: different default";
```

While it is clear to me that the default is bad (it makes assumptions about how the connection line is oriented) and that this can be seen as a reason for deprecation, wouldn't it serve us better to undo the deprecation, remove the bad expression for the default, and say it's up to tools to choose a suitable alignment?

henrikt-ma:
Fixes #3832.

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -1267,19 +1267,20 @@ connect(a.x, b.x)
 \end{example}
 
 The optional \lstinline!Text!\index{Text@\robustinline{Text}!\robustinline{connect} annotation} primitive defines a text that will be written on the connection line.
-It has the following definition (it is not equal to the \lstinline!Text! primitive as part of graphics -- the differences are marked after \emph{Note} in the description-strings):
+It has the following definition (it is not equal to the \lstinline!Text! primitive as part of graphics -- the differences are marked with comments):
 \begin{lstlisting}[language=modelica]
 record Text
   extends GraphicItem;
   Extent extent;
-  String string "Note: different name";
+  Integer index; // Not present in other Text
+  String string; // Different name compared to other Text
   Real fontSize = 0 "unit pt";
   String fontName;
   TextStyle textStyle[:];
   Color textColor = Black;
   TextAlignment horizontalAlignment =
-    if index < 0 then TextAlignment.Right else TextAligment.Left "Note: different default";
-  Integer index "Note: new";
+    if index < 0 then TextAlignment.Right else TextAligment.Left;
+    // Different default compared to other Text
 end Text;
 \end{lstlisting}
 
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -1283,7 +1283,7 @@ record Text
 end Text;
 \end{lstlisting}
 
-The \lstinline!index! is one of the points of Line (numbered 1, 2, 3, \ldots{} where negative numbers count from the end, thus -1 indicate the last one).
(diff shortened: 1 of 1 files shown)


chapters/annotations.tex
```diff
@@ -1267,31 +1267,30 @@ \subsection{Connections}\label{connections1}
...
 The \lstinline!string! may use the special symbols \lstinline!"%first"! and \lstinline!"%second"! to indicate the connectors in the \lstinline!connect!-equation.
 
-The \lstinline!extent! and \lstinline!rotation! are relative to the \lstinline!origin! (default \lstinline!{0, 0}!) and the \lstinline!origin! is relative to the point on the \lstinline!Line!.
+The meanings of \lstinline!extent! and \lstinline!rotation! (inherited from \lstinline!GraphicItem!), as well as \lstinline!fontSize!, \lstinline!fontName! \lstinline!textStyle!, and \lstinline!textColor! are the same as for the other \lstinline!Text! annotation in \cref{text}.
 
-The \lstinline!textColor! attribute defines the color of the text.
-The text is drawn with transparent background and no border around the text (and without outline).
-The default value for \lstinline!horizontalAlignment! is deprecated.
-Having a zero size for the \lstinline!extent! is deprecated and is handled as if upper part is moved up an appropriate amount.
+When \lstinline!horizontalAlignment! is \lstinline!TextAlignment.Automatic!, the tool automatically determines the alignment to reduce collisions with other graphical items.
```
HansOlsson: Might be specific:
```suggestion
When \lstinline!horizontalAlignment! is \lstinline!TextAlignment.Automatic!, the tool automatically determines the alignment to reduce collisions with other graphical items, in particular the connection line itself and the components it is connected to.
```

DagBruck (approved):
Ok.

chapters/annotations.tex
DagBruck: Either is good enough. Or keep the original and just remove the word _other._

henrikt-ma: I did what @DagBruck suggested.  Can we then consider this resolved?

PR #3849, opened 2026-03-03
Initialization for sample

### LINKED ISSUE #3843: Clarification on events at initialization
From OpenModelica/OpenModelica#15023.

Consider this MWE

```modelica
model WhenNotInitialized
  Boolean clock = sample(0, 10.0);
  Integer count_1(start=0, fixed = true);
  Integer count_2(start=0, fixed = true);
  Integer count_3(start=0, fixed = true);
equation
  when time >= 0 then
      count_1 = pre(count_1) + 1;
  end when;
  
  when clock then
    count_2 = pre(count_2) + 1;
  end when;
   
  when {clock, initial()} then
    count_3 = pre(count_3) + 1;
  end when;
end WhenNotInitialized;
```
The question is: what are the expected values of count_1, count_2, and count_3 at time = 0.0? My and @AnHeuermann's understanding is 0, 0, 1. Here's the motivation:

### Case 1.

```modelica
  when time >= 0 then
      count_1 = pre(count_1) + 1;
  end when;
```
The when clause is not active during initialization, because you have to explicitly write `when initial()` to get that, see [Section 8.6](https://specification.modelica.org/master/equations.html#initialization-initial-equation-and-initial-algorithm). Once the simulation starts, time is already >= 0, so it doesn't _become_ >= 0, hence the when clause is never triggered. Expected result at time = 0: `count_1 = 0`.

### Case 2.
```modelica
  Boolean clock = sample(0, 10.0);
equation
  when clock then
    count_2 = pre(count_2) + 1;
  end when;
```
According to [MLS 3.7.5, Operator 3.23 sample](https://specification.modelica.org/master/operators-and-expressions.html#modelica:event-sample): 
> sample(start, interval) returns true and triggers time events at time instants 𝑠𝑡𝑎𝑟𝑡+i⋅𝑖𝑛𝑡𝑒𝑟𝑣𝑎𝑙 for i=0, 1⁢…, and is only true during the first event iteration at those times. At event iterations after the first one at each event and during continuous integration the operator always returns false.

During initialization , the variable `clock` is computed to true, since `sample()` returns true at time start = 0; the when clause is not activated, as there is no `when initial()`.

Then the simulation starts. All variables are computed again at time = StartTime = 0, but this time with fixed parameters and (discrete and continuous) states. sample() returns true again, so clock is again computed to be true. Hence, it does not _become_ true, so the when cause is not activated.

Expected result at time = 0: `count_2 = 0`.

### Case 3.

```modelica
  Boolean clock = sample(0, 10.0);
equation
  when {clock, initial()} then
    count_3 = pre(count_3) + 1;
  end when;
```
During initialization the when clock is active because of initial(), so count_3 = 1, clock = true. During the first step at time = 0 clock is still true, so the when clause is not triggered because it doesn't _become_ true. Expected result at time = 0: `count_3 = 1`.

### Summary

- Our interpretation: 0, 0, 1
- Dymola: 0, 1, 2
- OpenModelica: 0, 1, 2
- Other tools?

What should the correct results be?

HansOlsson:
Closes #3843
Adds the text from that ticket, and a bit more to make it even clearer.

### DIFF
--- a/chapters/operatorsandexpressions.tex
+++ b/chapters/operatorsandexpressions.tex
@@ -1426,6 +1426,7 @@ sample($\mathit{start}$, $\mathit{interval}$)
 \begin{semantics}
 Returns \lstinline!true! and triggers time events at time instants $\mathit{start} + i \cdot \mathit{interval}$ for $i = 0,\, 1\, \ldots$, and is only true during the first event iteration at those times.
 At event iterations after the first one at each event, during initialization, and during continuous integration the operator always returns \lstinline!false!.
+At the start of the simulation the first event iteration occurs after the initialization.
 The starting time $\mathit{start}$ and the sample interval $\mathit{interval}$ must be parameter expressions and need to be a subtype of \lstinline!Real! or \lstinline!Integer!.
 The sample interval $\mathit{interval}$ must be a positive number.
 The operator is not allowed inside \lstinline!function! classes.
--- a/chapters/operatorsandexpressions.tex
+++ b/chapters/operatorsandexpressions.tex
@@ -1425,7 +1425,7 @@ sample($\mathit{start}$, $\mathit{interval}$)
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
 Returns \lstinline!true! and triggers time events at time instants $\mathit{start} + i \cdot \mathit{interval}$ for $i = 0,\, 1\, \ldots$, and is only true during the first event iteration at those times.
-At event iterations after the first one at each event and during continuous integration the operator always returns \lstinline!false!.
+At event iterations after the first one at each event, during initialization, and during continuous integration the operator always returns \lstinline!false!.
 The starting time $\mathit{start}$ and the sample interval $\mathit{interval}$ must be parameter expressions and need to be a subtype of \lstinline!Real! or \lstinline!Integer!.
 The sample interval $\mathit{interval}$ must be a positive number.
 The operator is not allowed inside \lstinline!function! classes.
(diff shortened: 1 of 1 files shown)


chapters/operatorsandexpressions.tex
```diff
@@ -1425,7 +1425,8 @@ \subsection{Event-Related Operators with Function Syntax}\label{event-related-op
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
 Returns \lstinline!true! and triggers time events at time instants $\mathit{start} + i \cdot \mathit{interval}$ for $i = 0,\, 1\, \ldots$, and is only true during the first event iteration at those times.
-At event iterations after the first one at each event and during continuous integration the operator always returns \lstinline!false!.
+At event iterations after the first one at each event, during initialization, and during continuous integration the operator always returns \lstinline!false!.
+At the start of the simulation the first event iteration occurs after the initialization.
```
MarkusOlssonModelon: Following the logic from the issue the first line should be sufficient to resolve this.

> During initialization , the variable clock is computed to true, since sample() returns true at time start = 0; the when clause is not activated, as there is no when initial().
> 
> Then the simulation starts. [...] sample() returns true again, so clock is again computed to be true. Hence, it does not become true, so the when cause is not activated.

With the change that `sample()` always returns false during initialization, it now becomes true at the start of simulation, so the when-clause is active.

You could simply remove the second line, or extend it to clarify this, e.g:

```suggestion
At the start of the simulation the first event iteration occurs after the initialization even if the start argument is identical to the start time of the simulation.
```

Should possibly be non-normative.

MarkusOlssonModelon (changes requested):

MarkusOlssonModelon (approved):

PR #3850, opened 2026-03-04
And end-parenthesis in example

### LINKED ISSUE #3844: Clarify inverse of functions
First, the first box in chapter 12.7.3 Using the Inverse Annotation a bracket is missing:
`annotation ( inverse (uk = f2(. . ., y, . . .));`
`annotation ( inverse (uk = f2(. . ., y, . . .)));`

Second, to me it is not clear whether pairs of a function and its inverse should both have the inverse annotation, 
i.e. looking at Modelica.Math:
sin is the inverse of asin and asin is the inverse of sin
and so on ... (cos, tan, sinh, cosh, tanh. exp)

see also: [#4751](https://github.com/modelica/ModelicaStandardLibrary/issues/4751)

HansOlsson:
Closes #3844

### DIFF
--- a/chapters/functions.tex
+++ b/chapters/functions.tex
@@ -1536,7 +1536,7 @@ function $f_1$
   output $T_2$ $y$;
 algorithm
   $\ldots$
-  annotation(inverse($u_k$ = $f_2$($\ldots$, $y$, $\ldots$));
+  annotation(inverse($u_k$ = $f_2$($\ldots$, $y$, $\ldots$)));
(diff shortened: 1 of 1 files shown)


henrikt-ma (commented):
Adding the missing parenthesis looks like a good idea, but I'll leave it to @AHaumer to provide the review with approval if he thinks this is sufficient to close the issue.

AHaumer (approved):
perfect

PR #3853, opened 2026-03-06
States are always active for one tick.

### LINKED ISSUE #3842: Clarification about subsequent transition in StateMachines
In the minimal model below, the conditions of two subsequent transitions become true in the same clock tick.
According to the current Modelica Specification, it is not intuitively clear whether both transitions should fire within that same tick. Section 17.1 on immediate transitions reads as if this should happen (which also feels like the intuitive behavior). However, in Section 17.3.4, the “Semantics Summary,” I do not find any mechanism that would handle such a case. So probably the transitions should instead occur sequentially.
In any case, it would be helpful if the intended behavior were defined more explicitly.

```
model State_Machine
 Real sigIn=sample(time, Clock(0.05));
 inner output Real sigOut;
 block State0
  outer output Real sigOut;
 equation
   sigOut = 0;
 end State0;

 block State1
  outer output Real sigOut;
 equation
   sigOut = 1;
 end State1;

 block State2
  outer output Real sigOut;
 equation
   sigOut = 2;
 end State2;
 State0 state0;
 State1 state1;
 State2 state2;
equation
  initialState(state0);
  transition(state0, state1, sigIn > 0.5);
  transition(state1, state2, sigIn > 0.4);
end State_Machine;
```

HansOlsson:
Closes #3842

@gkurzbach Gerd, can you handle it or forward to your colleague, I don't have his github handle?

### DIFF
--- a/chapters/statemachines.tex
+++ b/chapters/statemachines.tex
@@ -17,8 +17,8 @@ The following properties are different to Lucid Synchrone 3.0:
 
   For this reason, the state machines in this chapter use \emph{immediate} (= the same as \emph{strong}) and \emph{delayed} transitions.
   Delayed transitions are \emph{immediate} transitions where the condition is automatically delayed with an implicit \lstinline!previous!.
-  
-  This implies that states are active for at least one tick.
+
+  This implies that states are active for at least one tick, even if an \emph{immediate} exit transition is active when entering the state.
 \item
   Parallel state machines can be explicitly synchronized with a language element (similarly to parallel branches in Sequential Function Charts).
   This often occurring operation can also be defined in Statecharts or in Lucid Synchrone state machines but only indirectly with appropriate conditions on transitions.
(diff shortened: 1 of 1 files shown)


chapters/statemachines.tex
```diff
@@ -17,6 +17,8 @@ \chapter{State Machines}\label{state-machines}
 
   For this reason, the state machines in this chapter use \emph{immediate} (= the same as \emph{strong}) and \emph{delayed} transitions.
   Delayed transitions are \emph{immediate} transitions where the condition is automatically delayed with an implicit \lstinline!previous!.
+  
+  This implies that states are active for at least one tick.
```
JohannesKrauseGitH: Technically this is sufficient I think, but maybe we can make the consequences little bit more explicit?

HansOlsson: good point, I tried to be a bit more explicit.

JohannesKrauseGitH (approved):

PR #3856, opened 2026-03-10
Define when-clause clearly

### LINKED ISSUE #3840: Clarify meaning of when-clause

henrikt-ma:
This is an alternative to the PR #3841.  I believe it is in line with what was said in the last language group meeting.

Fixes #3840.


HansOlsson (approved):
Looks good.

PR #3858, opened 2026-03-27
More detailed explanation of the two variants of using time.

### LINKED ISSUE #3855: What is the point of sampling 'time' with inferred clock?
Why is `sample` applied to `time` here?
https://github.com/modelica/ModelicaSpecification/blob/016cdbdaae84e67c9cd78a8926cac5a99b3c4ec0/chapters/synchronous.tex#L320-L324

It reminds me of how `time` is also sometimes sampled in the MSL, such as in [Modelica.Clocked.RealSignals.TimeBasedSources.Sine](
https://github.com/modelica/ModelicaStandardLibrary/blob/81e2c95fc66a7a43027a33cfe8f8e874e119f8ca/Modelica/Clocked/RealSignals/TimeBasedSources/Sine.mo#L15):
```
  simTime = sample(time);
```

What is the point of sampling `time` in these ways?  If it is pointless, should the specification and the MSL be simplified in order to not give the impression that it is needed?

HansOlsson:
Closes #3855

### DIFF
--- a/chapters/synchronous.tex
+++ b/chapters/synchronous.tex
@@ -985,11 +985,8 @@ A special case is the built-in variable \lstinline!time! (see \cref{built-in-var
 Each use of \lstinline!time! is conceptually included as a separate variable in this analysis, $\mathit{time}_i$ with $\text{\lstinline!der!}(\mathit{time}_{i}) = 1$.
 \begin{nonnormative}
 This means that  \lstinline!time! can be used in two different ways.
-
-The most intuitive is \lstinline!sample(time)! which works in any clocked partition, similarly as sampling any other continuous-time variable.
-
-Additionally, \lstinline!time! can also be directly used in different clocked partitions without any restrictions.
-However, every sub-partition directly referencing \lstinline!time! contains a call to \lstinline!der!, which makes it a discretized sub-partition \cref{discretized-partition}.
+The most intuitive is \lstinline!sample(time)!, which can be part of both discrete-time and discretized sub-partitions, similarly as sampling any other continuous-time variable.
+On the other hand, if \lstinline!time! is used without sampling it in a clocked partition, it brings a use of \lstinline!der! to its sub-partition, and hence the sub-partition will be a discretized one (compare \cref{discretized-partition}).
 \end{nonnormative}
 
 \subsection{Base-Partitioning}\label{base-clock-partitioning}
--- a/chapters/synchronous.tex
+++ b/chapters/synchronous.tex
@@ -984,8 +984,12 @@ A set of clock partitions is the \emph{connected components} (Wikipedia, \emph{C
 A special case is the built-in variable \lstinline!time! (see \cref{built-in-variable-time}).
 Each use of \lstinline!time! is conceptually included as a separate variable in this analysis, $\mathit{time}_i$ with $\text{\lstinline!der!}(\mathit{time}_{i}) = 1$.
 \begin{nonnormative}
-This means that \lstinline!time! can be used in different partitions without any restrictions.
-Additionally, it means that every sub-partition directly referencing \lstinline!time! contains a call to \lstinline!der!.
+This means that  \lstinline!time! can be used in two different ways.
+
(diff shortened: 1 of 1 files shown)


chapters/synchronous.tex
```diff
@@ -984,8 +984,12 @@ \subsection{Connected Components of the Equations and Variables Graph}\label{con
 A special case is the built-in variable \lstinline!time! (see \cref{built-in-variable-time}).
 Each use of \lstinline!time! is conceptually included as a separate variable in this analysis, $\mathit{time}_i$ with $\text{\lstinline!der!}(\mathit{time}_{i}) = 1$.
 \begin{nonnormative}
-This means that \lstinline!time! can be used in different partitions without any restrictions.
-Additionally, it means that every sub-partition directly referencing \lstinline!time! contains a call to \lstinline!der!.
+This means that  \lstinline!time! can be used in two different ways.
+
+The most intuitive is \lstinline!sample(time)! which works in any clocked partition, similarly as sampling any other continuous-time variable.
```
henrikt-ma: I get the impression that it should be emphasized that it can be part of a discrete-time sub-partition?

```suggestion
The most intuitive is \lstinline!sample(time)!, which can be part of both discrete-time and discretized sub-partitions, similarly as sampling any other continuous-time variable.
```

chapters/synchronous.tex
```diff
@@ -984,8 +984,12 @@ \subsection{Connected Components of the Equations and Variables Graph}\label{con
 A special case is the built-in variable \lstinline!time! (see \cref{built-in-variable-time}).
 Each use of \lstinline!time! is conceptually included as a separate variable in this analysis, $\mathit{time}_i$ with $\text{\lstinline!der!}(\mathit{time}_{i}) = 1$.
 \begin{nonnormative}
-This means that \lstinline!time! can be used in different partitions without any restrictions.
-Additionally, it means that every sub-partition directly referencing \lstinline!time! contains a call to \lstinline!der!.
+This means that  \lstinline!time! can be used in two different ways.
+
```
henrikt-ma: Instead of all these one or two sentence paragraphs, I think it will look better if we merge everything into a single non-normative paragraph on the theme of `time`.

```suggestion
```

chapters/synchronous.tex
```diff
@@ -984,8 +984,12 @@ \subsection{Connected Components of the Equations and Variables Graph}\label{con
...
+The most intuitive is \lstinline!sample(time)! which works in any clocked partition, similarly as sampling any other continuous-time variable.
+
```
henrikt-ma: As above:

```suggestion
```

chapters/synchronous.tex
```diff
@@ -984,8 +984,12 @@ \subsection{Connected Components of the Equations and Variables Graph}\label{con
...
+Additionally, \lstinline!time! can also be directly used in different clocked partitions without any restrictions.
+However, every sub-partition directly referencing \lstinline!time! contains a call to \lstinline!der!, which makes it a discretized sub-partition \cref{discretized-partition}.
```
henrikt-ma: I didn't see the value of saying _without any restrictions_.  Would this simplified one-sentence alternative work?
```suggestion
On the other hand, if \lstinline!time! is used without sampling it in a clocked partition, it brings a use of \lstinline!der! to its sub-partition, and hence the sub-partition must be a discretized one (compare \cref{discretized-partition}).
```

HansOlsson: Good, but I will change to "will be a discretized one". The reason is that "must be" to me implies that there's an error if it's not discretized, but it is not possible as "der" in itself makes it discretized.

HansOlsson: As follows:
```suggestion
On the other hand, if \lstinline!time! is used without sampling it in a clocked partition, it brings a use of \lstinline!der! to its sub-partition, and hence the sub-partition will be a discretized one (compare \cref{discretized-partition}).
```

henrikt-ma (approved):
Looks good to me, but feel free to await more positive reviews from experts on the synchronous language features.

PR #3859, opened 2026-03-27
Restrict input/output for outer.

### LINKED ISSUE #1658: Semantics of inner outer output
**Reported by thiele on 6 Feb 2015 13:46 UTC**
What are the semantics of `inner outer output`. Consider following model:
```modelica
model A "inner outer output semantics"
  inner Real i;
  B b;
  model B
    // inner outer Real i; // This is clearly defined, but below
    inner outer output Real i; // valid? semantics?
    C c;
    model C
      outer output Real i;
    equation 
        i = 3;
    end C;
  end B;
equation 
  i = 2;
end A;
```

OpenModelica simulates the model assigning `i:=2` to the "inner i" and `i:=3` to the "inner outer i". My Dymola doesn't simulate it claiming that there is one equation too many (it accepts the model after deleting either `i:=2` or `i:=3`).

Is there any normative text in the specification that defines the semantics of `inner outer output`?

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/1658

HansOlsson:
Closes #1658

### DIFF
--- a/chapters/scoping.tex
+++ b/chapters/scoping.tex
@@ -168,6 +168,7 @@ An element declared with the prefix \lstinline!outer!\indexinline{outer} referen
 
 Outer component declarations shall not have modifications (including binding equations).
 Outer class declarations should be defined using short-class definitions without modifications.
+Outer component declarations shall not have the prefix \lstinline!input! and may only use \lstinline!output! in state machines with special semantics, \cref{merging-variable-definitions}.
 However, see also \cref{simultaneous-inner-outer-declarations}.
 
 If the outer component declaration is a disabled conditional component (\cref{conditional-component-declaration}) it is also ignored for the automatic creation of inner component (neither causing it; nor influencing the type of it).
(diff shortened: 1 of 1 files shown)


chapters/scoping.tex
```diff
@@ -168,6 +168,7 @@ \section{Inner Declarations - Instance Hierarchy Name Lookup}\label{instance-hie
 
 Outer component declarations shall not have modifications (including binding equations).
 Outer class declarations should be defined using short-class definitions without modifications.
+Outer component declarations shall not have the prefix \lstinline!input! and may only use \lstinline!output! in state machines with special semantics, \cref{merging-variable-definitions}.
```
henrikt-ma: I think we tend to put in a word before the reference:

```suggestion
Outer component declarations shall not have the prefix \lstinline!input! and may only use \lstinline!output! in state machines with special semantics, see \cref{merging-variable-definitions}.
```

HansOlsson:
Language group: Some discussion about generalizing to non-statemachines, but proposal seems fine as is.

PR #3861, opened 2026-03-27
Fix remaining cases where evaluable variability should be required

### LINKED ISSUE #3811: Various places where parameter variability seems too high

henrikt-ma:
Fixes #3811.  (The non-synchronous part of the issue has already been fixed by #3839.)


HansOlsson (approved):
Ok. I noticed that https://github.com/modelica/ModelicaSpecification/blob/3352d86b7927e0d037bcff72d4ea9df3b2545ef9/chapters/classes.tex#L159
isn't clear with what "evaluable expressions" (previously parameter expressions) refer to. But that can be cleaned up later.

I _assume_ it means the array size n*2 is evaluable, and the binding expression doesn't matter.

PR #3862, opened 2026-03-28
Clarify role of when-clause conditions in event iteration

### LINKED ISSUE #3813: Event iteration convergence and reinit
There was some discussion about event iteration in #578, but I don't see that the current specification text makes the relation between `reinit` and event iteration clear.  What we have is this:
> \lstinline!reinit! does not break the single assignment rule, because \lstinline!reinit(x, expr)! in equations evaluates \lstinline!expr! to a value, then at the end of the current event iteration step it assigns this value to \lstinline!x! (this copying from values to reinitialized state(s) is done after all other evaluations of the model and before copying \lstinline!x! to \lstinline!pre(x)!).

and the criterion for convergence of the event iteration:
> Before the start of the integration, it must be guaranteed that for all variables \lstinline!v!, \lstinline!v = pre(v)!.

However, this doesn't mention that the `reinit` essentially invalidates the "current event iteration step", meaning that the event iteration convergence test is not meaningful right away.

Here is a model which illustrates the need:
```
model ReinitLadder
  Real x(start = 1, fixed = true);
equation
  der(x) = -1;
  when x < 0 then
    reinit(x, 0.2);
  elsewhen x > 0.1 then
    reinit(x, 0.4);
  elsewhen x > 0.3 then
    reinit(x, 0.6);
  elsewhen x > 0.5 then
    reinit(x, 0.8);
  elsewhen x > 0.7 then
    reinit(x, 1.0);
  end when;
  annotation(
    experiment(
      StopTime = 3.0
    )
  );
end ReinitLadder;
```

henrikt-ma:
Fixes #3813.

While I do think that this should be sufficient to consider #3813 fixed, it leaves me a bit uneasy that my intuition about the event iteration is not captured in any non-normative section.  That is, the intuition is that event iteration must be continued when some computed solution to the model equations becomes invalid due to some of the updates taking place after the evaluation of the model equations.  The intuition is also that once none of the computed solutions become invalidated, it should in principle be a no-op to evaluate the model equations one more time, so tools should err on the side of iterating one extra time rather than one time too little.

The two things taking place after evaluation of the model equations are the `pre(v) := v` updates and the execution of `reinit` statements.  Thus, a tool erring on the side of iterating one time too much could simply test whether there is any change to a `pre(v)` or a state updated by a `reinit`.  By noting that execution of a `reinit` implies that there is a `pre(v)` which will be updated from false to true, the condition may be simplified to the current rule in the specification.

However, the intuition about invalidated solutions to the model equations is a more intuitive starting point for optimizing the iteration, and in my opinion also a more useful way to abstractly understand what it is that the event iteration achieves:
- **Event iteration ensures that the model equations are solved consistently before transient analysis may proceed forward in time after an event.**

In view of the text proposed in this PR, it is particularly worth noting that the intuition does not suggest that a triggered `when`-clause should necessarily trigger continued iteration; the usual sorting of equations is sufficient, except for `reinit` which may change a state after it has been used.  Hence, the conceptual idea of introducing implicit variables for the `when`-clause conditions is neither helpful for tools that want to avoid unnecessary iterator nor for users who are trying to understand optimized tool behavior.

### DIFF
--- a/chapters/equations.tex
+++ b/chapters/equations.tex
@@ -806,15 +806,6 @@ In the case a parameter has both a binding equation and \lstinline!fixed = false
 
 Continuous-time \lstinline!Real! variables \lstinline!vc! have exactly one initialization value since the rules above assure that during initialization \lstinline!vc = pre(vc) = vc.startExpression! (if \lstinline!fixed = true!).
 
-Before the start of the integration, it must be guaranteed that for all variables \lstinline!v!, \lstinline!v = pre(v)!.
-If this is not the case for some variables \lstinline!vi!, \lstinline!pre(vi) := vi! must be set and an event iteration at the initial time must follow, so the model is re-evaluated, until this condition is fulfilled.
-In detail this means that during initialization initial equations and normal equations are solved with \lstinline!v! and \lstinline!pre(v)! as unknowns without any event iterations.
-Then only the normal equations are solved repeatedly (each time after \lstinline!v! is copied to \lstinline!pre(v)!) until \lstinline!v = pre(v)!.
-
-\begin{nonnormative}
-Tools may optimize initialization by not computing unnecessary \lstinline!pre(v)!, and only performing the event iteration if necessary.
-\end{nonnormative}
-
 A Modelica translator may first transform the continuous equations of a model, at least conceptually, to state space form.
 This may require to differentiate equations for index reduction, i.e., additional equations and, in some cases, additional unknown variables are introduced.
 This whole set of equations, together with the additional constraints defined above, should lead to an algebraic system of equations where the number of equations and the number of all variables (including \lstinline!der! and \lstinline!pre! variables) is equal.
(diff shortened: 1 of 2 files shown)


chapters/equations.tex
```diff
@@ -937,6 +935,22 @@ \section{Initialization, initial equation, and initial algorithm}\label{initiali
...
+  Real x(start = 0.0, fixed = true);
+equation
+  der(x) = 0;
+  when b then
+    reinit(x, 5);
+  end when;
+\end{lstlisting}
+After solving the initialization problem, both \lstinline!b! and \lstinline!pre(b)! are false.
+During transient analysis, \lstinline!initial()! is false, so \lstinline!b! changes from falso to true during the first round of the initial event iteration.
+This triggers the \lstinline!when!-clause, and \lstinline!x! gets reinitialized.
```
HansOlsson, line 951: The example work, but I'm not sure if we want to promote the use of "not initial()" etc.
What about a model like:
```
   Integer x(start=0), old_x;
   when {initial(),sample(1,1)} then
      x=pre(x)+1;
   end when;
   old_x=pre(x);
```
After solving the initialization  problem pre(x)=0, x=1, old_x=0 (and the when-clause is active)
During transient analysis an event loop is started leading to pre(x)=1, x=1, old_x=1 (the when-clause is no longer active).
Is that also clear?

henrikt-ma: I would say that `when not initial()` is a useful idiom, and hence actually nice to "promote" like this.

HansOlsson (approved):
Looks ok now.

MarkusOlssonModelon (approved):

PR #3863, opened 2026-03-30
Linkable grammar productions and lexical units

henrikt-ma:
This is a follow-up to https://github.com/modelica/ModelicaSpecification/pull/3838#discussion_r2924519521.

Opening in draft state since there are still things to do:
- Reduce/avoid duplicated grammar listings, relying on linked references to production rules instead.
- ~Fix the LaTeXML build.  The current solution was verified to work with cross references within the same generated HTML document, but it seems to break when crossing document borders.~

### DIFF
--- a/chapters/equations.tex
+++ b/chapters/equations.tex
@@ -131,22 +131,11 @@ The \lstinline!for!-equations/\lstinline!if!-equations are expanded.
 
 The same restrictions apply to \lstinline!Connections.branch!, \lstinline!Connections.root!, and \lstinline!Connections.potentialRoot!; which after expansion are handled according to \cref{equation-operators-for-overconstrained-connection-based-equation-systems1}.
 
+
 \subsection{If-Equations}\label{if-equations}
 
-The \lstinline!if!-equations\index{if@\robustinline{if}!equation}\index{then@\robustinline{then}!if-equation@\robustinline{if}-equation}\index{else@\robustinline{else}!if-equation@\robustinline{if}-equation}\index{elseif@\robustinline{elseif}!if-equation@\robustinline{if}-equation} have the following syntax:
-\begin{lstlisting}[language=grammar]
-if expression then
-  { some-equation ";" }
-{ elseif expression then
-  { some-equation ";" }
-}
-[ else
-  { some-equation ";" }
-]
-end if ";"
-\end{lstlisting}
-
-The \lstinline!expression! of an \lstinline!if!- or \lstinline!elseif!-clause must be a scalar \lstinline!Boolean! expression.
+The syntax of an \lstinline!if!-equations\index{if@\robustinline{if}!equation}\index{then@\robustinline{then}!if-equation@\robustinline{if}-equation}\index{else@\robustinline{else}!if-equation@\robustinline{if}-equation}\index{elseif@\robustinline{elseif}!if-equation@\robustinline{if}-equation} is given by \productionref{if-equation} in the grammar.
+Here, each \productionref{expression} in the grammar (representing the condition of an \lstinline!if!- or \lstinline!elseif!-clause) must be a scalar \lstinline!Boolean! expression.
 One \lstinline!if!-clause, and zero or more \lstinline!elseif!-clauses, and an optional \lstinline!else!-clause together form a list of branches.
 One or zero of the bodies of these \lstinline!if!-, \lstinline!elseif!- and \lstinline!else!-clauses is selected, by evaluating the conditions of the \lstinline!if!- and \lstinline!elseif!-clauses sequentially until a condition that evaluates to true is found.
 If none of the conditions evaluate to true the body of the \lstinline!else!-clause is selected (if an \lstinline!else!-clause exists, otherwise no body is selected).
@@ -170,6 +159,7 @@ Additional restrictions apply in combination with \lstinline!when!-equations, se
 If the first condition is violated, the single assignment rule would not hold, because the number of equations may change during simulation although the number of unknowns remains the same.
 The second condition provides the generalization of \eqref{eq:dae-discrete-valued} to \lstinline!if!-equations.
 \end{nonnormative}
+
 
 \subsection{When-Equations}\label{when-equations}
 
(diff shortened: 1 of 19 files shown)


chapters/equations.tex
```diff
@@ -77,29 +61,20 @@ \subsection{Simple Equality Equations}\label{simple-equality-equations}
...
-   IDENT [ in expression ]
-\end{lstlisting}
+A \lstinline!for!-equation may optionally use several iterators (\productionref{for-indices})\index{in@\robustinline{in}!for-equation@\robustinline{for}-equation}, see \cref{nested-for-loops-and-reduction-expressions-with-multiple-iterators} for more information.
 
-The following is one example of a prefix of a \lstinline!for!-equation:
+% FIXME: It is weird to give an example in the form of grammar like this:
+The following is an example of a prefix of a \lstinline!for!-equation:
 \begin{lstlisting}[language=grammar]
 for IDENT in expression loop
 \end{lstlisting}
```
henrikt-ma: I'd like to draw the attention to this grammar listing and the corresponding listing for the `for`-statement.  Can't we simply drop these sections?

HansOlsson: Looking at it, my first guess was that it was defining "prefix of a for-equation" as a term. 

But that term isn't used anywhere - so that cannot be the reason. Looking back I see that 2.0 called it a "for clause" and in 2.2 the "prefix of a for-equation" was introduced.

So, my revised explanation is that the fragment was originally called a "for clause", but then we recognized that it was incomplete, and called it a "prefix of a..." to make that clearer.

However, it doesn't seem clear and the fragment doesn't seem to bring any benefit, so I would agree with dropping this one - and similar ones. If the corresponding terms are actually used, we need to consider it in more detail.

henrikt-ma: Good.  Dropped.  Resolving.

chapters/functions.tex
```diff
@@ -1087,13 +1068,13 @@ \section{Derivatives and Inverses of Functions}\label{derivatives-and-inverses-o
 
 \begin{annotationdefinition}[smoothOrder]
 \begin{synopsis}[grammar]\begin{lstlisting}
-"smoothOrder" "=" UNSIGNED-NUMBER
+"smoothOrder" "=" unsigned-number
```
HansOlsson: Isn't this a mistake that should be corrected as follows:
```suggestion
"smoothOrder" "=" UNSIGNED-INTEGER
```

chapters/functions.tex
```diff
@@ -1087,13 +1068,13 @@ \section{Derivatives and Inverses of Functions}\label{derivatives-and-inverses-o
...
 "smoothOrder"
    "("
       "normallyConstant" "=" IDENT
       { "," "normallyConstant" "=" IDENT }
    ")"
-   "=" UNSIGNED-NUMBER
+   "=" unsigned-number
```
HansOlsson: Same
```suggestion
   "=" UNSIGNED-INTEGER
```

chapters/functions.tex
```diff
@@ -1134,7 +1115,7 @@ \section{Derivatives and Inverses of Functions}\label{derivatives-and-inverses-o
    "(" derivative-constraint { "," derivative-constraint } ")"
 
 derivative-constraint :
-   "order" = UNSIGNED-NUMBER
+   "order" = unsigned-number
```
HansOlsson: Dito
```suggestion
   "order" = UNSIGNED-INTEGER
```

chapters/packages.tex
```diff
@@ -227,25 +227,21 @@ \subsection{Directory Hierarchy Mapping}\label{mapping-a-package-class-hierarchy
...
 For a sub-entity of an enclosing structured entity, the \lstinline!within!-clause shall designate the class of the enclosing entity; and this class must exist and must not have been defined using a short class definition.
-See \cref{stored-definitions-containing-multiple-class-definitions} regarding the use of \lstinline!within!-clause when a \lstinline[language=grammar]!stored-definition! does not hold exactly one class definition.
+See \cref{stored-definitions-containing-multiple-class-definitions} regarding the use of \lstinline!within!-clause when a \productionref{stored-definition} does not hold exactly one class definition.
 
 \begin{example}
-The subpackage \lstinline!Rotational! declared within \lstinline!Modelica.Mechanics! has the fully qualified name \lstinline!Modelica.Mechanics.Rotational!, which is formed by concatenating the \lstinline[language=grammar]!packageprefixname! with the short name of the package.
+The subpackage \lstinline!Rotational! declared within \lstinline!Modelica.Mechanics! has the fully qualified name \lstinline!Modelica.Mechanics.Rotational!, which is formed by concatenating the \productionref{name} with the short name of the package.
```
HansOlsson: Hm... Is it fully clear that `name` refer name-production within `within`? I guess the "packageprefixname" was intended to clarify that.

chapters/statements.tex
```diff
@@ -5,26 +5,24 @@ \chapter{Statements and Algorithm Sections}\label{statements-and-algorithm-secti
...
 \section{Algorithm Sections}\label{algorithm-sections}
 
 An \firstuse[algorithm!section]{algorithm section} is a part of a class definition comprised of the keyword \lstinline!algorithm!\index{algorithm@\robustinline{algorithm}} followed by a sequence of statements.
-The formal syntax is as follows:
-\begin{lstlisting}[language=grammar]
-algorithm-section :
-   [ initial ] algorithm { statement ";" }
-\end{lstlisting}
-
+The syntax is given by \productionref{algorithm-section} in the grammar, and the different kinds of equatins that may occur within are given by \productionref{statement}.
```
HansOlsson: Spelling
```suggestion
The syntax is given by \productionref{algorithm-section} in the grammar, and the different kinds of equations that may occur within are given by \productionref{statement}.
```

chapters/packages.tex
henrikt-ma: That's a valid question.  Would this be better?
```suggestion
The subpackage \lstinline!Rotational! declared within \lstinline!Modelica.Mechanics! has the fully qualified name \lstinline!Modelica.Mechanics.Rotational!, which is formed by concatenating the \productionref{name} of the \productionref{within-clause} with the short name of the package.
```

(I see that having _name_ hyperlinked here could be considered misleading, but I'd like to keep it so for the sake of consistency.)

HansOlsson: Yes. that works.

HansOlsson (approved):
Looks good.

PR #3867, opened 2026-04-02
Non-normatively give rough explanation of 'tunable parameter'

### LINKED ISSUE #3860: Tunable parameters don't exist according to specification
I recently received a complaint from someone trying to understand what a tunable parameter is.  The person does have a background in FMI, and naturally turned to the Modelica specification for answers.  However, what one finds in the specification (https://specification.modelica.org/master/class-predefined-types-and-declarations.html#component-variability) is just this (my emphasis):
> A variable np declared with the parameter prefix, is called a non-evaluable parameter variable unless it is an evaluable parameter. It is also simply called a non-evaluable parameter. **It does not change during transient analysis**, with a value determined by the initialization problem. For further details, see [4.5.2](https://specification.modelica.org/master/class-predefined-types-and-declarations.html#parameters).

Also, the linked section about parameters does not leave any room for tunable parameters.

I suggest that we reformulate slightly with tunability in mind, being well aware that there is currently no consensus about how to characterize the set of tunable parameters, or what it even means to tune a tunable parameter.

For example, something along those lines:
> A variable np declared with the parameter prefix, is called a non-evaluable parameter variable unless it is an evaluable parameter. It is also simply called a non-evaluable parameter. It does not change during transient analysis, with a value determined by the initialization problem. For further details and tool-specific exceptions, see [4.5.2](https://specification.modelica.org/master/class-predefined-types-and-declarations.html#parameters).

Details to give in [4.5.2](https://specification.modelica.org/master/class-predefined-types-and-declarations.html#parameters), possibly as non-normative:
> A variant of non-evaluable parameter which is in use, but with tool-dependent meaning, is _tunable parameter_ (ADD TO DOCUMENT INDEX).  Despite the rule that a parameter (evaluable or not) does not change during transient analysis, the idea is that some parameters still would be _possible_ to change at events during transient analysis.  When a tool identifies such parameters, they are often referred to as _tunable_ parameters.  Similar to other parameter component variabilities, some of the tunable parameters may be directly assigned at the events, whereas other tunable parameters will only be updated as a consequence of the direct assignments. Note that the criterion of being possible to change at an event, as well as the related the semantics of making a change, are tool-dependent.

We might even give a small example, but I think we need to keep it really simple to avoid discussing tool variations:
```
  parameter Real tau(unit = "s") = 1.0;
equation
  tau * der(x) = -x;
```

If the existence of the tunable parameter concept is recognized by the specification, I also have a customer request to discuss.  The customer would like an annotation for asserting that a particular parameter must be tunable.  The use of the annotation is two-fold:
- Make it possible to detect that a parameter is not tunable already during translation, instead of when later failing to change the value.
- When detected during translation, the tool is in a much better position for providing the reason for non-tunability.

Instead of introducing this as a vendor-specific annotation, is `Tunable = true` something worth having in the standard?  Of course, it should be formulated as something only applying to tools with the non-standard support for tunability.  In my opinion, it would be a nice step in the direction of alignment with FMI.  Would we expect problems due to tool variations to be an issue in practice?

henrikt-ma:
Fixes #3860.

### DIFF
--- a/chapters/classes.tex
+++ b/chapters/classes.tex
@@ -590,6 +590,7 @@ When a tool identifies such a parameter, it is often referred to as a \emph{tuna
 Both evaluable and non-evaluable parameters can be tunable.
 Similar to other parameter component variabilities, some of the tunable parameters may be directly assigned at the events, whereas other tunable parameters will only be updated as a consequence of the direct assignments.
 Note that the criterion of being possible to change at an event, as well as the related semantics of making a change, are tool-dependent.
+Possible reasons for a parameter not being tunable include that changes would cause significant non-physical transients, require a complicated re-evaluation of parameter expressions, or make no difference after initialization.
 \end{nonnormative}
 
 \begin{nonnormative}
--- a/chapters/classes.tex
+++ b/chapters/classes.tex
@@ -589,7 +589,7 @@ Despite the rule that a parameter (evaluable or not) does not change during tran
 When a tool identifies such a parameter, it is often referred to as a \emph{tunable parameter}.
 Both evaluable and non-evaluable parameters can be tunable.
 Similar to other parameter component variabilities, some of the tunable parameters may be directly assigned at the events, whereas other tunable parameters will only be updated as a consequence of the direct assignments.
-Note that the criterion of being possible to change at an event, as well as the related the semantics of making a change, are tool-dependent.
+Note that the criterion of being possible to change at an event, as well as the related semantics of making a change, are tool-dependent.
 \end{nonnormative}
 
 \begin{nonnormative}
--- a/chapters/classes.tex
+++ b/chapters/classes.tex
@@ -589,8 +589,9 @@ Despite the rule that a parameter (evaluable or not) does not change during tran
 When a tool identifies such a parameter, it is often referred to as a \emph{tunable parameter}.
 Both evaluable and non-evaluable parameters can be tunable.
 Similar to other parameter component variabilities, some of the tunable parameters may be directly assigned at the events, whereas other tunable parameters will only be updated as a consequence of the direct assignments.
-Note that the criterion of being possible to change at an event, as well as the related semantics of making a change, are tool-dependent.
+Note that the criterion of being possible to change at an event, as well as the related semantics of making a change (deviating from the mathematical formalism given in \cref{modelica-dae-representation}), are tool-dependent.
 Possible reasons for a parameter not being tunable include that changes would cause significant non-physical transients, require a complicated re-evaluation of parameter expressions, or make no difference after initialization.
+Tunable parameters are also mentioned informally for the \lstinline!interaction! annotation, see \cref{modelica:interaction}.
 \end{nonnormative}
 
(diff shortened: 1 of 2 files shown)


chapters/classes.tex
```diff
@@ -583,6 +583,15 @@ \subsection{Parameters}\label{parameters}
...
 \end{nonnormative}
 
+\begin{nonnormative}
+A variant of parameter component variability which is in use, but with tool-dependent meaning, is \firstuse{tunable parameter}.
+Despite the rule that a parameter (evaluable or not) does not change during transient analysis, the idea is that some parameters still would be possible to change at events during transient analysis.
+When a tool identifies such a parameter, it is often referred to as a \emph{tunable parameter}.
+Both evaluable and non-evaluable parameters can be tunable.
+Similar to other parameter component variabilities, some of the tunable parameters may be directly assigned at the events, whereas other tunable parameters will only be updated as a consequence of the direct assignments.
+Note that the criterion of being possible to change at an event, as well as the related the semantics of making a change, are tool-dependent.
+\end{nonnormative}
```
HansOlsson: I would add a bit more as follows:
```suggestion
Note that the criterion of being possible to change at an event, as well as the related semantics of making a change, are tool-dependent.
Some reasons for not making all parameters tunable are that changes would cause significant non-physical transients, require a complicated re-evaluation of parameter expressions, or make no difference after initialization.
\end{nonnormative}
```

chapters/classes.tex
```diff
@@ -583,6 +583,15 @@ \subsection{Parameters}\label{parameters}
 Here, a good error message for the variability error can include the information that the reason for \lstinline!n! being a non-evaluable parameter is that it has a dependency on the non-evaluable parameter \lstinline!b!.
 \end{nonnormative}
 
+\begin{nonnormative}
+A variant of parameter component variability which is in use, but with tool-dependent meaning, is \firstuse{tunable parameter}.
+Despite the rule that a parameter (evaluable or not) does not change during transient analysis, the idea is that some parameters still would be possible to change at events during transient analysis.
```
HansOlsson, line 588: Language group: Should clarify that triggers an event.
Further discussion: Clarify that may require an event (FMI requires switching to event mode).

HansOlsson: For changing: see also https://specification.modelica.org/master/annotations.html#user-input

chapters/classes.tex
henrikt-ma: Done, almost as suggested.  Resolving, as I expect that we will take another look at this tomorrow.

HansOlsson (approved):
LGTM

henrikt-ma:
Based on the discussion at the language meeting, I have now removed the assumption that tunable parameters may only change at events.  If I remember correctly, we also decided to not say anything regarding the potential need for event handling when a tunable parameter changes, but to just leave that as one (unmentioned) aspect of the tool-dependent semantics.

henrikt-ma:
I forgot that this PR was already merged.  I'll open a new one.

PR #3871, opened 2026-04-13
Non-normatively mention result-reference from 'der'

### LINKED ISSUE #3857: Where is der(v, N) defined?
The [Modelica grammar](https://specification.modelica.org/master/modelica-concrete-syntax.html) has:

```
result-reference :
   component-reference
   | time
   | der "(" ( component-reference | time ) [ "," UNSIGNED-INTEGER ] ")"
```
which is also referred to in [Section 18.5.2.3](https://specification.modelica.org/master/annotations.html#plot-curves). 

However, the definition of [Operator 3.11 der](https://specification.modelica.org/master/operators-and-expressions.html#modelica:der) in Section 3.7.4 just says:

der(𝑒𝑥𝑝𝑟)
    The time derivative of 𝑒𝑥𝑝𝑟. If the expression 𝑒𝑥𝑝𝑟 is a scalar it needs to be a subtype of Real.

I guess this should be updated, explaining the meaning of the optional second integer input of the operator.

henrikt-ma:
Fixes #3857.

### DIFF
--- a/chapters/operatorsandexpressions.tex
+++ b/chapters/operatorsandexpressions.tex
@@ -813,8 +813,15 @@ An exception is when the operator \lstinline!reinit! (\Cref{modelica:reinit}) is
 The operator is not allowed inside \lstinline!function! classes.
 If $\mathit{expr}$ is an array, the operator is applied to all elements of the array.
 For non-scalar arguments the function is vectorized according to \cref{vectorized-calls-of-functions}.
+
 \begin{nonnormative}
 For \lstinline!Real! parameters and constants the result is a zero scalar or array of the same size as the variable.
+\end{nonnormative}
+
+\begin{nonnormative}
+There is also a two argument form of \lstinline!der! used in annotations, see \productionref{result-reference} in the grammar.
+This form is used to refer to higher order derivatives of a component as described in \cref{plot-curves}.
+It is not valid as a general expression outside annotations.
 \end{nonnormative}
 
 \begin{example}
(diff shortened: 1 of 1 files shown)


chapters/operatorsandexpressions.tex
```diff
@@ -813,10 +813,17 @@ \subsection{Derivative and Special Purpose Operators with Function Syntax}\label
...
 For non-scalar arguments the function is vectorized according to \cref{vectorized-calls-of-functions}.
+
 \begin{nonnormative}
 For \lstinline!Real! parameters and constants the result is a zero scalar or array of the same size as the variable.
 \end{nonnormative}
 
+\begin{nonnormative}
+There is also a two argument form of \lstinline!der! used in annotations, see \productionref{result-reference} in the grammar.
+This form is used to refer to higher order derivatives of a component as described in \cref{plot-curves}.
+It is not valid as a general expression outside annotations.
```
```suggestion
It is not valid as a general expression outside those annotations.
```
Since it's only valid for some specific annotations.

henrikt-ma: It is true that it is not being used by all annotations, but _those_ is a dangling reference here.  Would it be better to just drop _outside annotations_?

HansOlsson: Yeah, that would work as well.

henrikt-ma: Done.  Resolving.

HansOlsson (approved):
LGTM

PR #3872, opened 2026-04-15
Work around issue with cleveref and hyperref

henrikt-ma:
This is a workaround for the bug that a reference to appendix B looks like _chapter B_.


HansOlsson (approved):
Looks good.
I noticed that it stops at subsubappendix - and we don't have any smaller parts (yet) in the appendix.

PR #3874, opened 2026-04-16
Require positive delayTime

### LINKED ISSUE #3245: Disallow zero delay

henrikt-ma:
Fixes #3245.  This is a replacement for #3724, agreed upon today in the language group meeting.

In addition to the core fix for #3245, I also included one of the non-normative sentences proposed in #3724.


HansOlsson (approved):
LGTM. Will wait a bit with merging in case anyone really objects.

HansOlsson:
No other comments, so merging.

PR #3875, opened 2026-04-16
Consistently call it "Modelica Language Specification", and consistent subtitle for Tutorial

HansOlsson:
Change title to the simpler "Modelica Language Specification" and split off the rest ("A Unified Object-Oriented Language for Systems Modeling") as a subtitle.

Tutorial is also updated:
* Same subtitle instead of the earlier one (originally the subtitle stated that it was for "Physical Systems Modeling" but with clock semantics the "Physical" was dropped to indicate that it is more general).
* Registered trademark.

### DIFF
--- a/Tutorial.md
+++ b/Tutorial.md
@@ -1,4 +1,5 @@
-# Modelica™ - A Unified Object-Oriented Language for Physical Systems Modeling Tutorial
+# Modelica® Tutorial
+**A Unified Object-Oriented Language for Physical Systems Modeling**
 
 ## Abstract:
 
--- a/Tutorial.md
+++ b/Tutorial.md
@@ -1,5 +1,5 @@
 # Modelica® Tutorial
-**A Unified Object-Oriented Language for Physical Systems Modeling**
+**A Unified Object-Oriented Language for Systems Modeling**
 
 ## Abstract:
 
--- a/chapters/titlepage.tex
+++ b/chapters/titlepage.tex
@@ -12,15 +12,15 @@
 \vspace{1cm}
 
 \huge
-Modelica\textsuperscript{\textregistered} -- A Unified Object-Oriented Language for~Systems Modeling
+Modelica\textsuperscript{\textregistered} Language Specification
 
-Language Specification
+\Large
+A Unified Object-Oriented Language for~Systems Modeling
 
 Version \mlsversion
 
 \vspace{1cm}% This has no effect in the generated HTML.
 
-\Large
 \makeatletter
 \@date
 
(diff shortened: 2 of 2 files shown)


Tutorial.md
```diff
@@ -1,4 +1,5 @@
-# Modelica™ - A Unified Object-Oriented Language for Physical Systems Modeling Tutorial
+# Modelica® Tutorial
+**A Unified Object-Oriented Language for Physical Systems Modeling**
```
HansOlsson:
```suggestion
**A Unified Object-Oriented Language for Systems Modeling**
```
Same as for specification.

henrikt-ma:
Maybe the title and description of this PR should put emphasis on aligning the _title page_ with what is already used as "title" in other places:
- HTML page header and PDF metadata derived from: `\title{Modelica\textsuperscript{\textregistered} Language Specification version~\mlsversion}`
- PDF page header: `\rhead{Modelica Language Specification \mlsversion\\ \nouppercase{\rightmark} \vspace{1mm}}`

For me, the fact that we include the version number in the places above does not imply that we cannot say that the title is just _Modelica Language Specification_, and that we have chosen to show the combination of title and version in these important locations.  Now, on the title page we have a different way of displaying the title, the version, plus the additional subtitle.

HansOlsson:
> Maybe the title and description of this PR should put emphasis on aligning the _title page_ with what is already used as "title" in other places:
> 
> * HTML page header and PDF metadata derived from: `\title{Modelica\textsuperscript{\textregistered} Language Specification version~\mlsversion}`
> * PDF page header: `\rhead{Modelica Language Specification \mlsversion\\ \nouppercase{\rightmark} \vspace{1mm}}`
> 
> For me, the fact that we include the version number in the places above does not imply that we cannot say that the title is just _Modelica Language Specification_, and that we have chosen to show the combination of title and version in these important locations. Now, on the title page we have a different way of displaying the title, the version, plus the additional subtitle.

Should now be hopefully fully done. 
Normally I don't update the description, as I more see it as the first part of a conversation, and thus updating it breaks the conversation.

HansOlsson:
Note that this doesn't change the actual version to 3.7, so we can merge it earlier.

chapters/titlepage.tex
```diff
@@ -12,15 +12,15 @@
 \vspace{1cm}
 
 \huge
-Modelica\textsuperscript{\textregistered} -- A Unified Object-Oriented Language for~Systems Modeling
+Modelica\textsuperscript{\textregistered} Language Specification
 
-Language Specification
+\Large
+A Unified Object-Oriented Language for~Systems Modeling
```
henrikt-ma: I don't think there should be a non-breakable space here anymore:

```suggestion
A Unified Object-Oriented Language for Systems Modeling
```

henrikt-ma (approved):
Good, apart from the non-breakable space which has become out of place.

PR #3877, opened 2026-04-16
See if we can just remove the MathJax.

HansOlsson:
Basically we currently:
- Mathematics rely on MathJax to generate CHTML (you can then switch to SVG) unless you use some specific browsers like Firefox.

The replacement disables this and instead relies on MathML support in modern browsers, see https://caniuse.com/mathml
- Firefox (as before)
- Safari
- Chrome
- Microsoft Edge  

The alternative would be to improve the logic as most browsers now are MathML capable (and use SVG as a better default).

Closes #3868


henrikt-ma (approved):
To really see if this works I think we'll have to merge first, so that the change makes it into to HTML development build that everyone can easily access.

PR #3878, opened 2026-04-17
Tunable parameter relation to events

henrikt-ma:
This is a small follow-up to #3867, including the outcome of discussions at the language meeting which happened after #3867 was merged.

Quoting https://github.com/modelica/ModelicaSpecification/pull/3867#issuecomment-4266635370:
> Based on the discussion at the language meeting, I have now removed the assumption that tunable parameters may only change at events. If I remember correctly, we also decided to not say anything regarding the potential need for event handling when a tunable parameter changes, but to just leave that as one (unmentioned) aspect of the tool-dependent semantics.


HansOlsson (approved):
Ok. I still think that handling it without an event is a bad idea, but leave that for tools.

PR #3883, opened 2026-04-21
Clarify that transitions also use the connection-annotations.

HansOlsson:
This is getting more and more important the more they diverge.
Since the example uses `Text(string=...)` it is closer to this one (so similar handling of text alignment).

Technically it may be that it should be yet another one, since `index` isn't used.

### DIFF
--- a/chapters/statemachines.tex
+++ b/chapters/statemachines.tex
@@ -169,7 +169,8 @@ equation
   \label{fig:state-machine-layout}
 \end{figure}
 
-The annotation for graphics of \lstinline!transition! has the following structure: \lstinline!annotation(Line($\ldots$), Text($\ldots$))!; and for \lstinline!initialState()!: \emph{graphical-primitives}\lstinline!(Line($\ldots$))!; with \lstinline!Line! and \lstinline!Text! annotations for connections defined in \cref{connections1}.
+Similar to \lstinline!connect!-equations, a \lstinline!transition! may have both a \lstinline!Line! and a \lstinline!Text! annotation, see \cref{connections1}.
+An \lstinline!initialState! equation may use any graphical primitive (including \lstinline!Line!) as annotation, see \cref{graphical-primitives}.
 
 \begin{example}
 \begin{lstlisting}[language=modelica]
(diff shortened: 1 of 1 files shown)


henrikt-ma (commented):
This looks like just a nice clarification, but then I don't know what to do with your comment about `index` not being used?

HansOlsson:
> This looks like just a nice clarification, but then I don't know what to do with your comment about `index` not being used?

I think we can just ignore it for now.

chapters/statemachines.tex
```diff
@@ -169,7 +169,7 @@ \section{State Machine Graphics}\label{state-machine-graphics}
   \label{fig:state-machine-layout}
 \end{figure}
 
-The annotation for graphics of \lstinline!transition! has the following structure: \lstinline!annotation(Line($\ldots$), Text($\ldots$))!; and for \lstinline!initialState()!: \emph{graphical-primitives}\lstinline!(Line($\ldots$))!; with \lstinline!Line! and \lstinline!Text! annotations defined in \cref{annotations}.
+The annotation for graphics of \lstinline!transition! has the following structure: \lstinline!annotation(Line($\ldots$), Text($\ldots$))!; and for \lstinline!initialState()!: \emph{graphical-primitives}\lstinline!(Line($\ldots$))!; with \lstinline!Line! and \lstinline!Text! annotations for connections defined in \cref{connections1}.
```
henrikt-ma: The clarification aside, this is a horrible-looking paragraph.  This would look better:

```suggestion
Similar to \lstinline!connect!-equations, a \lstinline!transition! may have both a \lstinline!Line! and a \lstinline!Text! annotation, see \cref{connections1}.
An \lstinline!initialState! equation may use any graphical primitive (including \lstinline!Line!) as annotation, see \cref{graphical-primitives}.
```

It's a very bad design, though, because if we were to treat this systematically in the specification, we would need to make every graphical primitive its own annotation, saying that it is allowed for \lstinline!initialState! equations.  I hope we leave this in its broken state, and don't decide to treat every graphical primitive as a first-class annotation.

henrikt-ma (approved):
The situation is horrible, but it's not this PR's fault.

PR #3885, opened 2026-04-23
Clarify that only sub-directories containing package.mo are considered for conflict

### LINKED ISSUE #3882: Conflict between external resources and directory hierarchy mapping?
## Current state

In https://specification.modelica.org/master/packages.html#external-resources it is clear that _modelica:/P.A.B/Resources/foo.jpg_ shall be stored at _…/P/A/B/Resources/foo.jpg_.  There is also a nice explanation of a potential confusion when storing an entire sub-package in a single file (my emphasis):
> When Modelica packages are stored hierarchically in a file system (i.e., package A in a directory A containing package.mo) the resource modelica:/A/Resources/C.jpg should be stored in the file A/Resources/C.jpg, it is **not recommend** to use modelica:/A.B/C.jpg for referencing resources; it could be stored in the file A/B/C.jpg – **which is counter-intuitive if A.B is stored together with A**.

Note: counter-intuitive, not illegal.

So far so good.

However, https://specification.modelica.org/master/packages.html#directory-hierarchy-mapping gives the following rule for the directory structure on disk,
> Two sub-entities shall not define classes with identical names

along with the example:
> Example : A directory shall not contain both the sub-directory A and the file A.mo.

### Question 1

Applied to the example with _modelica:/A.B/C.jpg_ above, does the rule above mean that the following file system hierarchy is forbidden?
```
A
- package.mo (containing definition of B)
- B
  - C.jpg
```
or are we saved by the definition of `B` in _package.mo_ not qualifying as a "sub-entity"?

### Question 2

Are the rules making the Modelica URI _modelica:/P.A.B/Resources/foo.jpg_ invalid if the entire `A` is stored in _P/A.mo_?

## Possible resolution

Could we settle these questions by only defining a directory as a sub-entity if it contains a _package.mo_?  Then the example would have to be adjusted slightly, like so:
> Example : A directory shall not contain both the sub-directory A containing a package.mo, and the file A.mo.

HansOlsson:
Closes #3882

### DIFF
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -217,7 +217,7 @@ If directory \filename{A} contains the three files \filename{package.mo}, \filen
 \end{example}
 
 \begin{example}
-A directory shall not contain both the sub-directory \filename{A} and the file \filename{A.mo}.
+A directory shall not contain both the sub-directory \filename{A} (containing a \filename{package.mo}) and the file \filename{A.mo}.
 \end{example}
 
 In order to preserve the order of sub-entities it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -210,6 +210,7 @@ The node typically contains documentation and graphical information for a packag
 A directory may also contain one or more sub-entities (directories or files).
 The sub-entities are mapped as elements of the class defined by their enclosing structured entity.
 Two sub-entities shall not define classes with identical names.
+Sub-directories shall either contain a \filename{package.mo} and be part of package, or not contain any file with the extension \filename{mo}.
 
 \begin{example}
 If directory \filename{A} contains the three files \filename{package.mo}, \filename{B.mo} and \filename{C.mo}, the classes defined are \lstinline!A!,
@@ -218,7 +219,6 @@ If directory \filename{A} contains the three files \filename{package.mo}, \filen
 
 \begin{example}
 A directory shall not contain both the sub-directory \filename{A} (containing a \filename{package.mo}) and the file \filename{A.mo}.
-A sub-directory not containing a \filename{package.mo} shall not contain any file with the extension \filename{mo}.
 \end{example}
 
 In order to preserve the order of sub-entities it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -218,6 +218,7 @@ If directory \filename{A} contains the three files \filename{package.mo}, \filen
 
 \begin{example}
 A directory shall not contain both the sub-directory \filename{A} (containing a \filename{package.mo}) and the file \filename{A.mo}.
+A sub-directory not containing a \filename{package.mo} shall not contain any file with the extension \filename{mo}.
 \end{example}
 
 In order to preserve the order of sub-entities it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -210,7 +210,8 @@ The node typically contains documentation and graphical information for a packag
 A directory may also contain one or more sub-entities (directories or files).
 The sub-entities are mapped as elements of the class defined by their enclosing structured entity.
 Two sub-entities shall not define classes with identical names.
-Sub-directories shall either contain a \filename{package.mo} and be part of package, or not contain any file with the extension \filename{mo}.
+Immediate sub-directories that contain a \filename{package.mo} are part of the package.
+It is recommended that immediate sub-directories not containing a \filename{package.mo} do not contain any file with the extension \filename{mo}.
 
 \begin{example}
 If directory \filename{A} contains the three files \filename{package.mo}, \filename{B.mo} and \filename{C.mo}, the classes defined are \lstinline!A!,


chapters/packages.tex
```diff
@@ -217,7 +217,7 @@ \subsection{Directory Hierarchy Mapping}\label{mapping-a-package-class-hierarchy
 \end{example}
 
 \begin{example}
-A directory shall not contain both the sub-directory \filename{A} and the file \filename{A.mo}.
+A directory shall not contain both the sub-directory \filename{A} (containing a \filename{package.mo}) and the file \filename{A.mo}.
```
maltelenz, line 222: Do we want to be stricter, and say `(containing any .mo files)`?

HansOlsson: Good point, but I would rather clarify that in general if there are .mo files there should be a package.mo, like:
```suggestion
A directory shall not contain both the sub-directory \filename{A} (containing a \filename{package.mo}) and the file \filename{A.mo}.
A sub-directory not containing a \filename{package.mo} shall not contain any file with the extension \filename{mo}.
```

maltelenz: Fine with me.

maltelenz (approved):

chapters/packages.tex
```diff
@@ -217,7 +217,8 @@ \subsection{Directory Hierarchy Mapping}\label{mapping-a-package-class-hierarchy
...
+A sub-directory not containing a \filename{package.mo} shall not contain any file with the extension \filename{mo}.
```
henrikt-ma: This looks like a rule rather than an observation to go into an example.  I believe what we need is a normative rule which only speaks of _package.mo_, and then the example could be reformulated to not look like a new rule:

```suggestion
In a sub-directory not containing a \filename{package.mo}, files with the extension \filename{mo} do not constitute entities in the package/class hierarchy.
```

henrikt-ma:
I miss the normative change that makes it clear that a directory only qualifies as an entry if it contains _package.mo_ (as opposed to qualifying by only being a directory, and than having to fulfill the requirement to contain _package.mo_).

That is, replace this:
```
The nature of such an external entity falls into one of the following two groups:
\begin{itemize}
\item
  Directory in the file system.
\end{itemize}

\begin{itemize}
\item
  File in the file system.
\end{itemize}
```
by this:
```
The nature of such an external entity in the file system falls into one of the following two groups:
\begin{itemize}
\item
  A directory containing a file named \filename{package.mo}.
\item
  A file with filename extension \filename{mo}.
\end{itemize}
```

henrikt-ma (changes requested):
We need normative rules outside the example, and no rules should be given in the example.

chapters/packages.tex
HansOlsson: It should be a rule; so I'll just remove it from the example.

chapters/packages.tex
```diff
@@ -210,14 +210,15 @@ \subsection{Directory Hierarchy Mapping}\label{mapping-a-package-class-hierarchy
 A directory may also contain one or more sub-entities (directories or files).
 The sub-entities are mapped as elements of the class defined by their enclosing structured entity.
 Two sub-entities shall not define classes with identical names.
+Sub-directories shall either contain a \filename{package.mo} and be part of package, or not contain any file with the extension \filename{mo}.
```
HansOlsson: If we find this too impactful it could be downgraded to "should".

chapters/packages.tex
henrikt-ma: Good.  Resolving.

chapters/packages.tex
henrikt-ma: I think it would be better to start with the suggested change in https://github.com/modelica/ModelicaSpecification/pull/3885#issuecomment-4307565166.  With that, I don't think we need another normative statement here.

As an alternative, to this statement, we could expand the example with the following:
> A file with filename extension \filename{mo} located in a sub-directory without a \filename{package.mo} is not part of the package hierarchy.

HansOlsson: To me they are attempting to handle the problem in different ways.

Consider a file-structure:
```
A (directory)
  - package.mo
  - B (directory)
    -  C.mo
```
To me this is a bad idea, and this is likely the result of some mistake, as a directory should either contain a package.mo - or no mo-files at all.

The alternative to not make B part of the package hierarchy implies that it is ok, but just ignored.

henrikt-ma: The use case is that `A.B` is defined in _A/package.mo_, and that _C.mo_ is a resource accessed as _modelica:/A.B/C.mo_.  That is, we shouldn't put any filename extension constraints on the resources stored under _modelica:/A.B_.

HansOlsson: Having .mo files inside a Modelica package that aren't part of the package (due to missing package.mo) isn't a good idea in my opinion. I'm fine with having it as just as a recommendation, or possibly delaying that after 3.7.

For the latter the following is solution:
```suggestion
Sub-directories are only considered as sub-entities if they directly contain a \filename{package.mo}.
```

maltelenz: > Having .mo files inside a Modelica package that aren't part of the package (due to missing package.mo) isn't a good idea in my opinion.

I agree. Anything else is one rename or new class operation away from collisions appearing from out of nowhere, to be handled (or more likely not handled and confusion ensuing) by a combination of users and tools. We should just forbid this esoteric use case to make everybodys life easier and more predictable.

HansOlsson: Can you discuss with @henrikt-ma and reach a consensus?

henrikt-ma: The real solution here is [MCP-0037](https://github.com/modelica/ModelicaSpecification/tree/MCP/0037/RationaleMCP/0037) where all external resources are placed in a directory that cannot be mistaken for containing classes in the class hierarchy, so let's go with the filename extension restriction for now then.

henrikt-ma: I realized that this might be a good compromise:
```suggestion
Immediate sub-directories shall either contain a \filename{package.mo} and be part of package, or not contain any file with the extension \filename{mo}.
```

However, I still think all of this would be much better stated higher up in the text:
```
The nature of such an external entity in the file system falls into one of the following two groups:
\begin{itemize}
\item
  A directory containing a file named \filename{package.mo}.
\item
  A file with filename extension \filename{mo}.
\end{itemize}
To avoid confusion, immediate sub-directories not containing a \filename{package.mo} are not allowed to contain any file with filename extension \filename{mo}.
```

henrikt-ma: Actually, the "compromise" is needed to support the MSL, and I don't find this sort of use odd at all:
- https://github.com/modelica/ModelicaStandardLibrary/blob/master/Modelica/Resources/Images/Media/UsersGuide/Demo.mo

henrikt-ma: Considering that a normative ban would need to go through a deprecation period during which tools should give deprecation warnings, after which tools should take action to reject libraries with banned filenames, I also see a clear advantage of giving a recommendation instead, as this makes tools fully compliant without the need to do anything.  I would see no issue with tools having an (ideally user-controllable) warning with such a specification, so in the end it comes down to whether we make it a requirement or not for tools to take action.

HansOlsson: My primary concern was with Immediate sub-directories; both as that has the highest risk - and because it is the easiest to check. But, as previously stated, it can be down-graded to a recommendation.

```suggestion
Immediate sub-directories that contain a \filename{package.mo} are part of the package.
It is recommended that immediate sub-directories not containing a \filename{package.mo} do not contain any file with the extension \filename{mo}.
```

henrikt-ma: Good, then just say this upfront when defining the two kind of entities (again, I can't make a Github suggestion, since the modified text is not close enough to the currently modified lines):
```
The nature of such an external entity in the file system falls into one of the following two groups:
\begin{itemize}
\item
  A directory containing a file named \filename{package.mo}.
\item
  A file with filename extension \filename{mo}.
\end{itemize}

\begin{nonnormative}
It is recommended that immediate sub-directories not containing a \filename{package.mo} do not contain any file with the extension \filename{mo}.
\end{nonnormative}
```

HansOlsson: I accepted the suggestion above.

I don't think it makes sense to move the text earlier - as that section describes the top-level part (one file or an entire directory).

For a directory mapping the restriction applies to all relevant sub-directories of the directory mapping - so `A/B/D/E.mo` below is an issue (as `A/B/D` is an immediate sub-directory of `A/B` containing a mo file without package.mo)
```
A (directory)
  - package.mo
  - B (directory)
    -  package.mo
    - C.mo
    - D (directory)
      - E.mo
```
We might refine it, but I think it is good enough for the release.

henrikt-ma (approved):
Sure, I'll try to take it from here in a separate PR that is not urgent to merge.

maltelenz (approved):

PR #3886, opened 2026-04-23
List37 changes

HansOlsson:
A list of already accepted changes for Modelica 3.7.
(Work in progress.)

### DIFF
--- a/RationaleMCP/Changes/3_7/3610.md
+++ b/RationaleMCP/Changes/3_7/3610.md
@@ -2,7 +2,7 @@
 
 ## Main change
 
-Variability of functions with `GenerateEvents`.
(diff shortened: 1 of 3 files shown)


HansOlsson:
It would be good to have this reviewed and merged to be able to do proper announcements, and update references (even if incomplete).

henrikt-ma:
Side question: Why have these documents ended up under _RationaleMCP_?

henrikt-ma (approved):
Looks good, but I haven't had the time to verify that nothing has been missed.

HansOlsson:
> Side question: Why have these documents ended up under _RationaleMCP_?

There's some relation between new changes and MCPs - and I didn't want to introduce another top-level directory.
I understand that we could restructure these directories, but it may take time

PR #3887, opened 2026-04-24
Use gender agnostic language

phannebohm:
This is a common way in English to refer to people without implying any gender. I think it is convenient to read and it avoids accidentally referring to males only. What do you think?

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -717,7 +717,7 @@ For example, a tool is not expected to provide means to plot a variable with \ls
 If a variable is declared in a protected section, a tool might not include it in a simulation result.
 By setting \lstinline!HideResult = false!, the modeler would like to have the variable in the simulation result, even if in the protected section.
 
-\lstinline!HideResult! is for example used in the connectors of the \lstinline!Modelica.StateGraph! library to not show variables to the modeler that are of no interest to him and would confuse him.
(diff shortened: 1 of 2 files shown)


chapters/annotations.tex
```diff
@@ -717,7 +717,7 @@ \section{Simulations}\label{annotations-for-simulations}\label{simulations}
 If a variable is declared in a protected section, a tool might not include it in a simulation result.
 By setting \lstinline!HideResult = false!, the modeler would like to have the variable in the simulation result, even if in the protected section.
 
-\lstinline!HideResult! is for example used in the connectors of the \lstinline!Modelica.StateGraph! library to not show variables to the modeler that are of no interest to him and would confuse him.
+\lstinline!HideResult! is for example used in the connectors of the \lstinline!Modelica.StateGraph! library to not show variables to the modeler that are of no interest to them and would confuse them.
```
henrikt-ma, line 720: The idea is nice, but it looks odd to combine _the modeler_ with _them_.

qlambert-pro: I think it may sound odd to some, but this is the correct way of writing these sentences.
https://en.wikipedia.org/wiki/Singular_they#Usage

HansOlsson (approved):
Looks good.

PR #3888, opened 2026-04-24
Clarity standard type coercion for function calls with multiple outputs

henrikt-ma:
This PR clarifies that the following is valid:
```
function f
  output Integer y1 = 1;
  output Integer y2 = 2;
end f;

model MultiCoercion
  Real x1;
  Real x2;
equation
  (x1, x2) = f();
end MultiCoercion;
```

This works in System Modeler, but it would be good to know about the support in other tools before we add it to the specification.

### DIFF
--- a/chapters/interface.tex
+++ b/chapters/interface.tex
@@ -543,7 +543,7 @@ The type of the full expression (e.g., \lstinline!if x then A else B!) is also d
   If \lstinline!A! is an \lstinline!Integer! expression:
   \begin{itemize}
   \item
-    In equations receiving multiple outputs of a function call and assignment statements, and where \lstinline!A! is part of the left-hand-side and \lstinline!B! the corresponding part of the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression.
+    In simple equality equations with non-trivial \productionref{output-expression-list} and assignment statements, and where \lstinline!A! is part of the left-hand-side and \lstinline!B! the corresponding part of the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression.
   \item
     Otherwise \lstinline!B! must be a \lstinline!Real! or \lstinline!Integer! expression.
     For exponentiation and division the type of the full expression is \lstinline!Real! (even if both \lstinline!A! and \lstinline!B! are \lstinline!Integer!) see \cref{element-wise-exponentiation} and \cref{division-by-numeric-scalars}, for relational operators the type of the full expression is \lstinline!Boolean!.
--- a/chapters/interface.tex
+++ b/chapters/interface.tex
@@ -543,7 +543,7 @@ The type of the full expression (e.g., \lstinline!if x then A else B!) is also d
   If \lstinline!A! is an \lstinline!Integer! expression:
   \begin{itemize}
   \item
-    For assignment statements and equations receiving several outputs of a function call, and where \lstinline!A! is part of the left-hand-side and \lstinline!B! part of the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression.
+    In equations receiving several outputs of a function call and assignment statements, and where \lstinline!A! is part of the left-hand-side and \lstinline!B! part of the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression.
   \item
     Otherwise \lstinline!B! must be a \lstinline!Real! or \lstinline!Integer! expression.
     For exponentiation and division the type of the full expression is \lstinline!Real! (even if both \lstinline!A! and \lstinline!B! are \lstinline!Integer!) see \cref{element-wise-exponentiation} and \cref{division-by-numeric-scalars}, for relational operators the type of the full expression is \lstinline!Boolean!.
--- a/chapters/interface.tex
+++ b/chapters/interface.tex
@@ -543,7 +543,7 @@ The type of the full expression (e.g., \lstinline!if x then A else B!) is also d
   If \lstinline!A! is an \lstinline!Integer! expression:
   \begin{itemize}
   \item
-    In equations receiving several outputs of a function call and assignment statements, and where \lstinline!A! is part of the left-hand-side and \lstinline!B! the corresponding part of the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression.
+    In equations receiving multiple outputs of a function call and assignment statements, and where \lstinline!A! is part of the left-hand-side and \lstinline!B! the corresponding part of the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression.
   \item
     Otherwise \lstinline!B! must be a \lstinline!Real! or \lstinline!Integer! expression.
     For exponentiation and division the type of the full expression is \lstinline!Real! (even if both \lstinline!A! and \lstinline!B! are \lstinline!Integer!) see \cref{element-wise-exponentiation} and \cref{division-by-numeric-scalars}, for relational operators the type of the full expression is \lstinline!Boolean!.
--- a/chapters/interface.tex
+++ b/chapters/interface.tex
@@ -540,9 +540,15 @@ The type of the full expression (e.g., \lstinline!if x then A else B!) is also d
   If \lstinline!A! is a \lstinline!Real! expression then \lstinline!B! must be a \lstinline!Real! or \lstinline!Integer! expression.
   The type of the full expression is \lstinline!Real!, compare \cref{standard-type-coercion}, unless the operator is a relational operator (\cref{equality-relational-and-logical-operators}) where the type of the full expression is \lstinline!Boolean!.
 \item
-  If \lstinline!A! is an \lstinline!Integer! expression then \lstinline!B! must be a \lstinline!Real! or \lstinline!Integer! expression.
-  For exponentiation and division the type of the full expression is \lstinline!Real! (even if both \lstinline!A! and \lstinline!B! are \lstinline!Integer!) see \cref{element-wise-exponentiation} and \cref{division-by-numeric-scalars}, for relational operators the type of the full expression is \lstinline!Boolean!.
-  In other cases the type of the full expression is \lstinline!Real! or \lstinline!Integer! (same as \lstinline!B!), compare \cref{standard-type-coercion}.
+  If \lstinline!A! is an \lstinline!Integer! expression:
+  \begin{itemize}
+  \item
+    For assignment statements and equations receiving several outputs of a function call, and where \lstinline!A! is part of the left-hand-side and \lstinline!B! part of the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression.
+  \item
+    Otherwise \lstinline!B! must be a \lstinline!Real! or \lstinline!Integer! expression.
+    For exponentiation and division the type of the full expression is \lstinline!Real! (even if both \lstinline!A! and \lstinline!B! are \lstinline!Integer!) see \cref{element-wise-exponentiation} and \cref{division-by-numeric-scalars}, for relational operators the type of the full expression is \lstinline!Boolean!.
+    In other cases the type of the full expression is \lstinline!Real! or \lstinline!Integer! (same as \lstinline!B!), compare \cref{standard-type-coercion}.
+  \end{itemize}
 \item
   If \lstinline!A! is a \lstinline!Boolean! expression then \lstinline!B! must be a \lstinline!Boolean! expression and the type of the full expression is \lstinline!Boolean!.
 \item
--- a/chapters/interface.tex
+++ b/chapters/interface.tex
@@ -524,17 +524,20 @@ The type of the full expression (e.g., \lstinline!if x then A else B!) is also d
   In an array expression the two records may contain elements with different sizes, but apart from that they must be type compatible.
   That generates a heterogenous array of records, see \cref{arrays}.
   The type of the full expression is a record comprised of named elements that are type compatible with the corresponding named elements of both \lstinline!A! and \lstinline!B!.
-\item The rules for array expressions depend on the operation (the rules for binary operators are given in \cref{scalar-vector-matrix-and-array-operator-functions}
-and for array concatenation in \cref{array-concatenation}).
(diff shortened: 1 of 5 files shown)


perost (approved):

perost:
Works fine in OpenModelica. 8.3.1 already says that they must be type compatible, but I guess it doesn't hurt to clarify it.

henrikt-ma:
Should we also clarify that this is not allowed?
```
function f
  input Real u;
  output Real y1 = u;
  output Integer y2 = 2;
end f;

model TwoWayMultiCoercion
  Real u;
  Integer x1 = 1;
  Real x2;
equation
  (x1, x2) = f(u);
end TwoWayMultiCoercion;
```

That is, that standard type coercion does not take place on the left-hand side expressions of a multi-output equation or statement?

HansOlsson:
I think we should prohibit that extra variant as well, see text below.

I agree that tools could handle it, but users can also handle it by re-writing as follows:

```
function f
  input Real u;
  output Real y1 = u;
  output Integer y2 = 2;
end f;

model TwoWayMultiCoercion
  Real u;
  Integer i1=1;
  final Real x1 = i1;
  Real x2;
equation
  (x1, x2) = f(u);
end TwoWayMultiCoercion;
```

We seem to have multiple variants here, and quite a long list of references:
- https://specification.modelica.org/master/equations.html#simple-equality-equations "The types of the left-hand-side and the right-hand-side of an equation need to be compatible in the same way as two arguments of binary operators" referencing https://specification.modelica.org/master/interface-or-type-relationships.html#type-compatible-expressions
- https://specification.modelica.org/master/interface-or-type-relationships.html#type-compatible-expressions "If A is a Real expression then B must be a Real or Integer expression. The type of the full expression is Real, ..." "If A is an Integer expression then B must be a Real or Integer expression." referencing https://specification.modelica.org/master/arrays.html#standard-type-coercion
- https://specification.modelica.org/master/arrays.html#standard-type-coercion "Proposed change"

I would propose to move the new text to 6.7 https://specification.modelica.org/master/interface-or-type-relationships.html#type-compatible-expressions and state, specifically ""If A is a Real expression then B must be a Real or Integer expression. " and change "If A is an Integer expression then B must be a Real or Integer expression.", to say that for assignment and/or multi-returning functions if "A" is an integer expression then B must be an Integer expression" (the latter handles the second case.)

There is also:
- https://specification.modelica.org/master/functions.html#output-formal-parameters-of-functions "The type of each component reference in the list must agree with the type of the corresponding output component."
- https://specification.modelica.org/master/statements-and-algorithm-sections.html#simple-assignment-statements - Nothing specified

HansOlsson (changes requested):
Should clarify a bit more. But good idea.

henrikt-ma:
The way I understand the specification is that we need both:
- The typing rules that explain when an `Integer` is compatible with `Real`.
- The standard type coercion rule which gives the (obvious) semantics to the correctly typed programs.

It would have made more sense to be to explain this in terms of an implicit conversion from `Integer` to `Real`, as this would cover both typing and semantics at once, but now it is was it is so I think we need to add text in both places.

henrikt-ma:
I am confused by the symmetrical treatment of `A` and `B` in https://specification.modelica.org/master/interface-or-type-relationships.html#type-compatible-expressions; it doesn't look like it would apply to things like assignment statements.  I hope SystemModeler is correct to reject this, which seems similar to what we need for multi-output function calls?
```
model AssignmentCoercion
  Integer k = 1;
  Real r;
algorithm
  k := r;
end AssignmentCoercion;
```

perost:
> I am confused by the symmetrical treatment of `A` and `B` in https://specification.modelica.org/master/interface-or-type-relationships.html#type-compatible-expressions; it doesn't look like it would apply to things like assignment statements. I hope SystemModeler is correct to reject this, which seems similar to what we need for multi-output function calls?
> 
> ```
> model AssignmentCoercion
>   Integer k = 1;
>   Real r;
> algorithm
>   k := r;
> end AssignmentCoercion;
> ```

The specification doesn't actually seem to say that the rhs of an assignment need to be type compatible with the lhs in any way from what I can see.

In OpenModelica we use a variant of 6.7 for this that is not symmetrical, but rather takes only one expression that's checked against an expected type and is type cast if necessary. I expect that's what most tools do, but it doesn't seem to be supported by the specification.

henrikt-ma:
> In OpenModelica we use a variant of 6.7 for this that is not symmetrical, but rather takes only one expression that's checked against an expected type and is type cast if necessary. I expect that's what most tools do, but it doesn't seem to be supported by the specification.

That makes sense, I wish it would also be how it was done in the specification…

For those binary operators where the operands should be treated symmetrically I supposed it just requires that the directed type compatibility is applied in both directions, saying that one of the operands needs to be type compatible with the other.

Until we have a better starting point for the typing rules, I'll concentrate on the standard type coercion semantics part.

HansOlsson:
> I am confused by the symmetrical treatment of `A` and `B` in https://specification.modelica.org/master/interface-or-type-relationships.html#type-compatible-expressions; it doesn't look like it would apply to things like assignment statements. I hope SystemModeler is correct to reject this, which seems similar to what we need for multi-output function calls?

That's why I proposed replacing: 
* If A is an Integer expression then B must be a Real or Integer expression. For exponentiation and division the type of the full expression is Real (even if both A and B are Integer) see [section 10.6.7](https://specification.modelica.org/master/arrays.html#element-wise-exponentiation) and [section 10.6.5](https://specification.modelica.org/master/arrays.html#division-by-numeric-scalars), for relational operators the type of the full expression is Boolean. In other cases the type of the full expression is Real or Integer (same as B), compare [section 10.6.13](https://specification.modelica.org/master/arrays.html#standard-type-coercion).

by something like:
* If A is an Integer expression:
   * For a multi-returning function or an assignment (where A is part of the left-hand-side and B part of the right-hand-side) then B must be an Integer expression, and the type of the expression is Integer. 
   * Otherwise B must be a Real or Integer expression. For exponentiation and division the type of the full expression is Real (even if both A and B are Integer) see [section 10.6.7](https://specification.modelica.org/master/arrays.html#element-wise-exponentiation) and [section 10.6.5](https://specification.modelica.org/master/arrays.html#division-by-numeric-scalars), for relational operators the type of the full expression is Boolean. In other cases the type of the full expression is Real or Integer (same as B), compare [section 10.6.13](https://specification.modelica.org/master/arrays.html#standard-type-coercion).

henrikt-ma:
Also, what is this statement supposed to mean?  Should it constitute its own paragraph?
> There is no similar rule for sub-typing.

henrikt-ma:
> by something like:

Done, so now we can have a conversation attached to the new text if needed.

henrikt-ma:
What about the special case of just having one variable on the left-hand side?
```
model AssignmentCoercion
  function f
    input Real u;
    output Real y1 = u;
    output Real y2 = u;
  end f;

  Integer k = 1;
  Real r;
equation
  k     = f(r); // Clearly OK.
  (k)   = f(r); // Cannot be different from the above.
  (k, ) = f(r); // OK or not?!
end AssignmentCoercion;
```

Depending on what we want, the current rule may need to be reformulated to speak of presence of commas instead of several received outputs.

chapters/interface.tex
```diff
@@ -524,22 +524,31 @@ \section{Type Compatible Expressions}\label{type-compatible-expressions}
...
 \item
-  If \lstinline!A! is an \lstinline!Integer! expression then \lstinline!B! must be a \lstinline!Real! or \lstinline!Integer! expression.
-  For exponentiation and division the type of the full expression is \lstinline!Real! (even if both \lstinline!A! and \lstinline!B! are \lstinline!Integer!) see \cref{element-wise-exponentiation} and \cref{division-by-numeric-scalars}, for relational operators the type of the full expression is \lstinline!Boolean!.
-  In other cases the type of the full expression is \lstinline!Real! or \lstinline!Integer! (same as \lstinline!B!), compare \cref{standard-type-coercion}.
+  If \lstinline!A! is an \lstinline!Integer! expression:
+  \begin{itemize}
+  \item
+    For assignment statements and equations receiving several outputs of a function call, and where \lstinline!A! is part of the left-hand-side and \lstinline!B! part of the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression.
```
HansOlsson:
```suggestion
    For assignment statements and equations receiving several outputs of a function call, and where \lstinline!A! is part of the left-hand-side and \lstinline!B! part of the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression, and the type of the full expression is \lstinline!Integer!.
```

HansOlsson (approved):
Looks good now.
The suggestion is that important - but more added for completeness.

chapters/interface.tex
henrikt-ma: I omitted this intentionally, since I didn't think it made sense to speak of the type of the assignment or equation as a whole, as these are not expressions?

HansOlsson:
> What about the special case of just having one variable on the left-hand side?
> 
> ```
> model AssignmentCoercion
>   function f
>     input Real u;
>     output Real y1 = u;
>     output Real y2 = u;
>   end f;
> 
>   Integer k = 1;
>   Real r;
> equation
>   k     = f(r); // Clearly OK.
>   (k)   = f(r); // Cannot be different from the above.
>   (k, ) = f(r); // OK or not?!
> end AssignmentCoercion;
> ```
> 
> Depending on what we want, the current rule may need to be reformulated to speak of presence of commas instead of several received outputs.

We could call it non-trivial parenthesized output-expression-list. As I understand it, for an equation it is the presence of a comma that matters, but for an assignment it applies to all parenthesized output-expression-list.

We need to differentiate it, since it doesn't make sense to have different rules for `(k)=f(r) `and `(k+2)=f(r)` or `(k)=f(r)+2`.

chapters/interface.tex
perost: This doesn't say how the expressions on the lhs and rhs are related to each other, and "part of" is a bit ambiguous. Maybe something like this instead?
```suggestion
    For assignment statements and equations receiving several outputs of a function call, and where \lstinline!A! is one of the component references (or derivatives) on the left-hand-side and \lstinline!B! the corresponding function call result on the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression.
```

HansOlsson: Well, the type does matter for the type-split "Local equation size", even though it is already defined as using the lhs-variable type.

perost: I don't see any issue with @HansOlsson's suggestion, since it's talking about the resulting type of A and B and not the whole assignment or equation.

perost: Maybe it would be better to say "In assignment statements and equation", rather than "For ..."?

HansOlsson: Possible, and could also swap from:

> In assignment statements and equations receiving several outputs of a function call,

to:

> In equations receiving several outputs of a function call and assignment statements,

As it relates to all assignment statements, not only the ones with several outputs.

henrikt-ma: > In equations receiving several outputs of a function call and assignment statements,

This is now done.

henrikt-ma: > I don't see any issue with @HansOlsson's suggestion, since it's talking about the resulting type of A and B and not the whole assignment or equation.

Is it really?  I thought it was talking about the type of "whole expressions" such as `A + B`?  That is, that the type of `1 + 1.5` is `Real`.

perost: > > I don't see any issue with @HansOlsson's suggestion, since it's talking about the resulting type of A and B and not the whole assignment or equation.
> 
> Is it really? I thought it was talking about the type of "whole expressions" such as `A + B`? That is, that the type of `1 + 1.5` is `Real`.

I was thinking about it in more in terms of `(A1, A2, ...) := (B1, B2, ...)`, and what the "compatible type" of e.g. A1 and B1 would be. But if we forbid type coercion in this case, then I guess it doesn't really matter anymore.

HansOlsson: I don't think this change is needed. 

If needed it would require more changes since
* **The text has been updated.**
* It should be one output from the corresponding function call on the right-hand-side, not just the function call.
* "Part of" kind of also work for record members in the left-hand-side; even if they are not the component-reference on the lhs.

HansOlsson: Can you confirm @perost ?

chapters/interface.tex
```diff
@@ -524,22 +524,31 @@ \section{Type Compatible Expressions}\label{type-compatible-expressions}
...
 \item
-  If \lstinline!A! is an \lstinline!Integer! expression then \lstinline!B! must be a \lstinline!Real! or \lstinline!Integer! expression.
-  For exponentiation and division the type of the full expression is \lstinline!Real! (even if both \lstinline!A! and \lstinline!B! are \lstinline!Integer!) see \cref{element-wise-exponentiation} and \cref{division-by-numeric-scalars}, for relational operators the type of the full expression is \lstinline!Boolean!.
-  In other cases the type of the full expression is \lstinline!Real! or \lstinline!Integer! (same as \lstinline!B!), compare \cref{standard-type-coercion}.
+  If \lstinline!A! is an \lstinline!Integer! expression:
+  \begin{itemize}
+  \item
+    In equations receiving several outputs of a function call and assignment statements, and where \lstinline!A! is part of the left-hand-side and \lstinline!B! part of the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression.
```
perost: The current text still doesn't say how A and B are related, other than being on different sides of the equation/assignment. 8.3.1 and 11.2.1.1 specify how the lhs and rhs are related though, so just adding "the corresponding" here should be enough to fix that.
```suggestion
    In equations receiving several outputs of a function call and assignment statements, and where \lstinline!A! is part of the left-hand-side and \lstinline!B! the corresponding part of the right-hand-side, \lstinline!B! must be an \lstinline!Integer! expression.
```

henrikt-ma:
> We could call it non-trivial parenthesized output-expression-list. As I understand it, for an equation it is the presence of a comma that matters, but for an assignment it applies to all parenthesized output-expression-list.

Done.

HansOlsson (approved):

PR #3890, opened 2026-04-28
Improve presentation of file system mapping

henrikt-ma:
This is a follow-up to #3885 with general cleanup of the the presentation in _File System Mapping of Package/Class_.

Improvements include:
- More careful introduction and use of terminology (_structured entity_ and _nonstructured entity_).
- Aligned language in subsections to make similarities easier to spot.
- Add missing rule that _package.mo_ should only define a single class.
- Expand non-normative section on sub-directories without _package.mo_.

### DIFF
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -177,17 +177,15 @@ In this case the search succeeds and the package \lstinline!MyLib.Pack2! is load
 \section{File System Mapping of Package/Class}\label{mapping-package-class-structures-to-a-hierarchical-file-system}\label{file-system-mapping-of-package-class}
 
 A package/class hierarchy may be represented in the hierarchical structure of the operating system (the file system).
-For classes with version information see also \cref{mapping-of-versions-to-file-system}.
-The nature of such an external entity falls into one of the following two groups:
+Such an entity in the file system is either of:
 \begin{itemize}
 \item
-  Directory in the file system.
+  A \willintroduce{structured entity}: A directory containing a file named \filename{package.mo} (\cref{directory-hierarchy-mapping}).
+\item
+  A \willintroduce{nonstructured entity}: A file with filename extension \filename{mo} (\cref{single-file-mapping}).
 \end{itemize}
 
-\begin{itemize}
-\item
-  File in the file system.
-\end{itemize}
+For classes with version information, see also \cref{mapping-of-versions-to-file-system}.
 
 Each Modelica file in the file system is stored in UTF-8 format (defined by The Unicode Consortium; \url{https://unicode.org}).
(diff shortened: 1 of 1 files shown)


HansOlsson:
I believe we need to discuss this more - as the PR made me realize that it is currently a bit of a mess.

So, the issue is that a .mo file can be:
- A file (not called package.mo) containing a top-level package or class
- A package.mo file for the class corresponding to the directory name
- A .mo file inside a directory with a package.mo

To me both of the latter ones are part of "structured entity"; i.e, the entire directory from the top-down would be called a structured entity. (Note that we have package.mo both at the top and further down; with similar rules - although only the top-directory may include version.)

HansOlsson:
Language group: should ideally to re-work to avoid the terms structured and non-structured entity.

chapters/packages.tex
```diff
@@ -200,19 +198,24 @@ \section{File System Mapping of Package/Class}\label{mapping-package-class-struc
...
+When mapping a package or class hierarchy to a directory, the directory is denoted a \firstuse{structured entity}.
+The file content of the \filename{package.mo} inside the directory shall match \productionref{stored-definition} in the grammar, defining a single class with name matching the name of the entity.
 
 \begin{nonnormative}
-The node typically contains documentation and graphical information for a package, but may also contain additional elements of the class \lstinline!A!.
+The \filename{package.mo} typically contains documentation and graphical information for a package, but may also contain additional elements of the package/class.
 \end{nonnormative}
 
-A directory may also contain one or more sub-entities (directories or files).
+A directory may also contain one or more sub-entities (structured or nonstructured, compare \cref{file-system-mapping-of-package-class}).
```
HansOlsson, line 208: As noted I would only call single files "nonstructured entity", so not call these files "nonstructured entity".

HansOlsson (changes requested):
The nonstructured entity is a bit of a mess, and this PR doesn't resolve that.
To me it should be reserved for single-file-mapping, and we don't need to define structured and nonstructured entity.

chapters/packages.tex
HansOlsson: I tried to look hard at the current specification and it seems to deliberately avoid using the term "nonstructured entity" for files that are part of directories with package.mo:

Specifically,
- A directory is called "structured entity"
- A single file is a "nonstructured entity"
- The other cases are called "sub-entity of an enclosing structured entity"

henrikt-ma:
OK to consider this PR superseded by #3924 and close without merging?

HansOlsson:
> OK to consider this PR superseded by #3924 and close without merging?

Yes.

PR #3894, opened 2026-05-06
Last minute amendments for C mapping of records

henrikt-ma:
This PR tries to mitigate the consequences of the rushed merge of #3198.


chapters/functions.tex
```diff
@@ -2314,6 +2310,27 @@ \subsection{Annotations for External Functions}\label{annotations-for-external-l
 /*literal*/ constant String Include
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
+Allowed for \lstinline!external!-clauses and record class annotations (\cref{annotations-for-c-mapping-of-records}).
+The included code should be valid C89 code.
+
+When \lstinline!annotation(Include="insertedCode")! appears in a record class annotation, the \lstinline!"insertedCode"! shall provide a definition (not merely a declaration without members) of the C struct to which the Modelica record is mapped.
+The struct members may be aligned differently, may be in a different order, and additional padding or fields may exist.
```
henrikt-ma: Different alignment compared to what?

HansOlsson: I believe the padding part handles the alignment so skip that.

```suggestion
The struct members may be in a different order, and additional padding or fields may exist.
```

HansOlsson (approved):

chapters/functions.tex
HansOlsson: As indicated above, I believe the padding part handles the alignment so skip that.

chapters/functions.tex
```suggestion
input R r;
output R2 r2;
```

HansOlsson: I don't know why comments don't work here.

chapters/functions.tex
henrikt-ma: Done.  Resolving.

chapters/functions.tex
henrikt-ma: Yeah, the Github comment feature used to work a lot better.

I am guessing that this is the fix for the typo reported in https://github.com/modelica/ModelicaSpecification/pull/3198#issuecomment-4387642205.  Please confirm that this is fixed and resolve.

chapters/functions.tex
```diff
@@ -2418,7 +2436,54 @@ \subsection{Annotations for External Functions}\label{annotations-for-external-l
...
+\begin{annotationdefinition}[CStructName]
+\begin{synopsis}\begin{lstlisting}
+/*literal*/ constant String CStructName
+\end{lstlisting}\end{synopsis}
+\begin{semantics}
+Allowed for record class annotations.
+
+The annotation \lstinline!CStructName = "CName"! specifies that \lstinline[language=C]!CName! is the name of the struct in the C mapping of the Modelica record class.
+
+By default the struct member names are the same as the record component names, but a different name can be specified using the \lstinline!CStructName! annotation.
```
```suggestion
By default the struct member names are the same as the record component names, but a different name can be specified using the \lstinline!CName! annotation.
```

HansOlsson:
I noticed that the `CName` annotation was added as well.
I assume that is intentional.

chapters/functions.tex
henrikt-ma: Good catch!

henrikt-ma:
> I noticed that the `CName` annotation was added as well. I assume that is intentional.

Yes, #3198 seemed half-baked without it.

chapters/functions.tex
```diff
@@ -2314,6 +2310,27 @@ \subsection{Annotations for External Functions}\label{annotations-for-external-l
 /*literal*/ constant String Include
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
+Allowed for \lstinline!external!-clauses and record class annotations (\cref{annotations-for-c-mapping-of-records}).
+The included code should be valid C89 code.
+
+When \lstinline!annotation(Include="insertedCode")! appears in a record class annotation, the \lstinline!"insertedCode"! shall provide a definition (not merely a declaration without members) of the C struct to which the Modelica record is mapped.
+The struct members may be in a different order compared to the record components, and additional padding or fields may exist.
+When there is no \lstinline!CStructName! annotation (\cref{modelica:CStructName}), it is assumed that the name of the struct is the same as the Modelica record class name.
```
henrikt-ma: We should make it a deprecated behavior to not provide this annotation when the record type is "used" in an `external`-clause.  I'll come back with a suggestion later today.

Modelica-ZhipengChen (commented):
This PR introduce `CName` annotation. Therefore, it is recommended to change the "CName" in `CStructName = "CName"` to something else to avoid confusion.

henrikt-ma:
> This PR introduce `CName` annotation. Therefore, it is recommended to change the "CName" in `CStructName = "CName"` to something else to avoid confusion.

Fixed.  It is now using the snake case `"c_name"` as example instead.  I also considered simply `"name"`, but I like the idea of including a hint that this name is in the C world, not the Modelica world.

chapters/functions.tex
henrikt-ma: Done.  Resolving.

chapters/functions.tex
```diff
@@ -2418,7 +2439,55 @@ \subsection{Annotations for External Functions}\label{annotations-for-external-l
...
+{\lstinline!CName!} & Name of struct member in C & \Cref{modelica:CName}\\
+{\lstinline!Include!} & Include directives & \Cref{modelica:Include}\\
+{\lstinline!IncludeDirectory!} & Directories containing header files & \Cref{modelica:IncludeDirectory}\\
+\hline
+\end{tabular}
+\end{center}
+
+The \lstinline!Include! and \lstinline!IncludeDirectory! annotations are also used with similar semantics for external functions.
+
+\begin{annotationdefinition}[CStructName]
```
henrikt-ma: When looking for the correct C terminology, I realized that the annotation is used to specify the struct _tag_, while struct _name_ could ambiguously be interpreted as the name of a typedef for the type.

Shouldn't the annotation be named `CStructTag` instead of `CStructName`?

henrikt-ma:
Requesting new review after adding deprecation of not having `Include` annotation when the C mapping of the record is needed.

The choice between `CStructTag` and `CStructName` (and reuse of the generic `CName`, for that matter) could be a topic for a poll in tomorrow's monthly design meeting.

chapters/functions.tex
HansOlsson: That would work as well. I notice that "tag" has been used in C at least as far back as K&R 2nd edition.

HansOlsson: All support for CStructTag
Discussion about deprecation, seems ok as well.

HansOlsson (approved):
ok

PR #3895, opened 2026-05-06
Correct hyphenation of left-hand side

henrikt-ma:
When preparing a minor clarification for #3888, I noticed that _(left|right)-hand side_ was often hyphenated incorrectly.  This PR does the cleanup and provides the minor clarification (which is formulated in terms of a left-hand side).


HansOlsson (approved):
Ok, seems consistent with style guides.

PR #3898, opened 2026-05-08
Be explicit about the struct being tagged

henrikt-ma:
The rest of the text makes more sense if we make it explicit from the beginning that the struct (to which a record is mapped) is tagged.  Since this already follows from text in other places, there is no urgency to include this (completely trivial) change in 3.7, in case it would slow down the release process.


HansOlsson (approved):
LGTM

HansOlsson:
For some reason there's a conflict.

PR #3906, opened 2026-05-20
Clean up casing of "Modelica language specification"

### LINKED ISSUE #3904: Outdated references to cleared appendix D
The reference to appendix D seems outdated.

> The contributors to this and to previous versions of this document are listed in appendix D.

> The revision history is available in appendix D.

henrikt-ma:
This is a cleanup PR with three parts:
- Consistent casing (and wording) when referring to the specification.
- Some trailing white-space cleanup in related markdown files.
- Fixes #3904.


HansOlsson (commented):
I agree with most of the changes, but some need a bit of discussion.
As far as I can tell the changes in the document are not blocking for the 3.7 release (will have to investigate if we can include them on the maintenance branch as a sort of errata without requiring a new vote).

I agree that it is much better to directly state that members of MAP-Lang contributed instead of referring to an appendix for the list of contributors and in the appendix only state that members of MAP-Lang contributed.

Discussion part:
I agree that we should be fairly consistent for the "Modelica Specification" capitalization, but for title, h1-headings and `#`-heading I think there's a strong case that one should use title capitalization, i.e., write it as "Modelica Specification". But that should be a bit more consistent - I notice that we also use "Hybrid Models" based on title-capitalization.

beutlich (commented):
Hm, not sure actually. For me it's always been Modelica Language, Modelica Specification, Modelica Conference, Modelica Design Meeting, Modelica Standard Library etc.

Repo name is ModelicaSpecification.

About text also is with capitals

grafik

HansOlsson:
> Hm, not sure actually. For me it's always been Modelica Language, Modelica Specification, Modelica Conference, Modelica Design Meeting, Modelica Standard Library etc.
> 
> Repo name is ModelicaSpecification.

Yes, thinking more I think that makes sense, as I see there are two related points:
* Titles and headings should use [title capitalization](https://en.wikipedia.org/wiki/Title_case) we see that used fairly consistently in both the specification text and in the tutorial.
* It should be consistent, so if the title (due to the previous fact) is "Modelica Language Specification" (without quotes) one will refer to it in that way (without quotes) - and that may be spread to things that seem like titles even if they aren't.

(The second point differs from APA style; but those rules seems over-designed.)

HansOlsson (changes requested):
As indicated I believe it should use title case.

henrikt-ma:
I have not updated based on the following reasoning:
- It should now be _Modelica Language Specification_ in title case when referring to the specification document; the name (title) of the document is _Modelica Language Specification_.
- However, I'd say that _Modelica_ is the name of the language, so when referring to the language it is fine to just say _Modelica_ when it is clear from context that it is the language and not something else Modelica-related that is being referred to.  When one wants to clarify/emphasize that it the language one is referring to, the _language_ is not treated as part of the name: _Modelica language_.  Similar constructs include _Modelica semantics_, _Modelica syntax_, and _Modelica community_.
- The full name of MAP-Lang is _Modelica Association Project Modelica Language_, or _Modelica Language_ (full title case) when the MAP qualification is clear from context.
- Even though not affecting this PR, I would reason similar when it comes to the MSL: The full name of the library is the _Modelica Standard Library_.  However, when just using the top-level package name to refer to the library, it becomes the _Modelica library_, just like they do it at https://github.com/modelica/ModelicaStandardLibrary/tree/master.  For another example, the Buildings people also write _Buildings library_ at https://github.com/lbl-srg/modelica-buildings.

henrikt-ma:
> * The full name of MAP-Lang is _Modelica Association Project Modelica Language_, or _Modelica Language_ (full title case) when the MAP qualification is clear from context.

I realize that I made a mistake here.  It is true that _Modelica Language_ is the name of the project, but MAP-Lang reads out as _Modelica Assocation Project Language_…

Edit: This triggered related PR: https://github.com/modelica/MAP-Lang_ProjectRules/pull/10

RationaleMCP/GuidingPrinciples.MD
```diff
@@ -1,17 +1,14 @@
-Guiding Principles of the Modelica Language Development
---
+# Guiding Principles of the Modelica language development
```
HansOlsson:
```suggestion
# Guiding Principles of the Modelica Language Development
```
I view this as title case; so all major words should be capital.

chapters/preface.tex
```diff
@@ -28,5 +28,6 @@ \chapter*{Preface}\label{preface}
 The text directly under the chapter headings are non-normative introductions to the chapters.
 
 The Modelica language has been developed since 1996.
+Modelica~1, the first version of Modelica, was released in September 1997, and had the language specification as a short appendix to the rationale.
```
HansOlsson, line 31: Interesting; I was surprised by it just being 1 - not 1.0; but the document itself had just a 1.

HansOlsson (commented):
Just a minor issue remains I think

HansOlsson:
Looks good now, except that the CI fails due to the removed appendix as it tries to link A5.html to the non-existent file.

henrikt-ma:
> Looks good now, except that the CI fails due to the removed appendix as it tries to link A5.html to the non-existent file.

Did you figure out where the link to A5.html is coming from?

HansOlsson:
> > Looks good now, except that the CI fails due to the removed appendix as it tries to link A5.html to the non-existent file.
> 
> Did you figure out where the link to A5.html is coming from?

I now see that it is `.CI\Jenkinsfile` - but I can't recall why all those links are created.

HansOlsson:
Update: The reason for the links was to keep old references working, and was added in https://github.com/modelica/ModelicaSpecification/issues/2545#issuecomment-614007961

I think we could remove all of the links now.

henrikt-ma:
> I think we could remove all of the links now.

Done.

HansOlsson (approved):
Looks good now.

Issue #3908, opened 2026-05-21
Some minor issues in the Spec found by AI


gkurzbach:
1.  Tense error – Appendix D: "This appendix described the history" → describes
 2. Doubled word – §18.5.2.1: "among the plots in the the same Figure annotation" → one the
 3. Doubled word – §18.5.2.7: "This is due to to the current inability" → one to
 4. Annotation renumbering error – §18.6 and §18.7: Annotations 18.1–18.4 are already used in §18.5 (info, revisions, styleSheets, figures); §18.6 re-opens at Annotation 18.1 Evaluate and §18.7 re-opens at Annotation 18.2 experiment … 18.4 TestCase — all should be renumbered continuing from 18.5 (likely caused by insertion of styleSheets annotation #3409 which shifted the sequence without updating §18.6/18.7)
 5. Typo – §16.5.1 (subSample): "every factor-th activativation" → activativation should be activation
 6. Wrong value in example – §16.3 (Real interval clock): comment says "ticks at 0, 0.003, 0.007, …" but with start=0.002 and interval=1e-3 the second tick is 0.002, not 0.003 — copy-pasted from the Rational interval clock example
 7. Imprecise cross-reference – §16.4 Operator 16.4 (Event clock): text refers reader to §16.9 for interval() at first tick, but the relevant content is in §16.10
 8. Typo ×2 – §18.9.1.2 (Graphical Properties): "The interaction attribue specifies ineractive features" → attribue → attribute; ineractive → interactive
 9. Copy-paste error – §18.9 Diagram annotation description: "The Diagram annotation defines the icon view of the class" → should be diagram view
10. Typo – §12.9.1.3: "annnotations" (three ns) → annotations
11. Wrong member name in example – §12.9.1.3: C struct generated from record R with members a1, a2 shows a1, b2 in the C struct — b2 should be a2 (no CName annotation justifies the rename; likely a copy-paste error)
12. §4.8 Balanced Models: connector class exemption from balancing (#3356) not clearly stated in normative text (only appears in a note); should be elevated

HansOlsson:
> 1. Tense error – Appendix D: "This appendix described the history" → describes

Possibly, but I believe the tense was chosen for a reason, and for the future we might perform an "appendectomy" instead.

Basically appendix D **described** the history up to 3.6 (see https://specification.modelica.org/maint/3.6/modelica-revision-history.html ) - but it is now only left as a placeholder in the specification; so I think removing it might be simpler.

[...]

PR #3913, opened 2026-05-25
Try to fix annotation numbering as seen in #3908 item 4

HansOlsson:
This seems to correct item 4 in #3908 
It is also similar to the other definitions at the same place. I don't know why it was so different before.

If this isn't the correct solution there should be more comments.


HansOlsson:
> This seems to correct item 4 in #3908 It is also similar to the other definitions at the same place. I don't know why it was so different before.
> 
> If this isn't the correct solution there should be more comments.

Well, the new idea is both have this correction and the comments!
I understand that we don't want to document all LaTeX commands, but when there have been errors I think it is justified.

henrikt-ma (approved):
Looks reasonable.

PR #3916, opened 2026-06-08
Avoid speaking of 'instance hierarchy'

henrikt-ma:
Fixing a few places where the term _instance hierarchy_ was used, and where I believe the correct term to use is _instance tree_.

### DIFF
--- a/chapters/scoping.tex
+++ b/chapters/scoping.tex
@@ -162,7 +162,7 @@ The package-restriction ensures that global name lookup of component references 
 See \cref{lookup-of-imported-names}.
 
 
-\section{Inner Declarations - Instance Tree Name Lookup}\label{instance-hierarchy-name-lookup-of-inner-declarations}\label{inner-declarations-instance-hierarchy-name-lookup}
+\section{Inner Declarations - Instance Tree Name Lookup}\label{instance-hierarchy-name-lookup-of-inner-declarations}\label{inner-declarations-instance-tree-name-lookup}
 
 An element declared with the prefix \lstinline!outer!\indexinline{outer} references an element instance with the same name but using the prefix \lstinline!inner!\indexinline{inner}.
 The \indexinline{inner} element shall be a child of an ancestor of the \lstinline!outer! element in the instance tree, and the match nearest to the \lstinline!outer! element is selected.
--- a/chapters/scoping.tex
+++ b/chapters/scoping.tex
@@ -188,7 +188,7 @@ An \lstinline!outer! element reference in a simulation model requires that one c
 An \lstinline!outer! element component may be of a partial class (but the referenced \lstinline!inner! component must be of a non-partial class).
 
 \begin{nonnormative}
-\lstinline!inner!/\lstinline!outer! components may be used to model simple fields, where some physical quantities, such as gravity vector, environment temperature or environment pressure, are accessible from all components in a specific instance tree.
+\lstinline!inner!/\lstinline!outer! components may be used to model simple fields, where some physical quantities, such as gravity vector, environment temperature or environment pressure, are accessible from all components in a part of the instance tree.
 Inner components are accessible throughout the model, if they are not ``shadowed'' by a corresponding \lstinline!inner! declaration in a more deeply nested level of the instance tree.
 \end{nonnormative}
 
--- a/chapters/scoping.tex
+++ b/chapters/scoping.tex
@@ -164,7 +164,8 @@ See \cref{lookup-of-imported-names}.
 
 \section{Inner Declarations - Instance Tree Name Lookup}\label{instance-hierarchy-name-lookup-of-inner-declarations}\label{inner-declarations-instance-hierarchy-name-lookup}
 
-An element declared with the prefix \lstinline!outer!\indexinline{outer} references an element instance with the same name but using the prefix \lstinline!inner!\indexinline{inner} which is nearest in the enclosing instance tree of the \lstinline!outer! element declaration.
+An element declared with the prefix \lstinline!outer!\indexinline{outer} references an element instance with the same name but using the prefix \lstinline!inner!\indexinline{inner}.
+The \indexinline{inner} element shall be a child of an ancestor of the \lstinline!outer! element in the instance tree, and the match nearest to the \lstinline!outer! element is selected.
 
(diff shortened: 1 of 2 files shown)


perost:
Maybe it would be a good idea to update the labels too, such as `instance-hierarchy-name-lookup-of-inner-declarations`.

HansOlsson:
> Maybe it would be a good idea to update the labels too, such as `instance-hierarchy-name-lookup-of-inner-declarations`.

Yes, they should be consistent.

chapters/scoping.tex
```diff
@@ -162,9 +162,9 @@ \subsection{Lookup of Imported Names}\label{lookup-of-imported-names1}
 See \cref{lookup-of-imported-names}.
 
 
-\section{Inner Declarations - Instance Hierarchy Name Lookup}\label{instance-hierarchy-name-lookup-of-inner-declarations}\label{inner-declarations-instance-hierarchy-name-lookup}
+\section{Inner Declarations - Instance Tree Name Lookup}\label{instance-hierarchy-name-lookup-of-inner-declarations}\label{inner-declarations-instance-hierarchy-name-lookup}
 
-An element declared with the prefix \lstinline!outer!\indexinline{outer} references an element instance with the same name but using the prefix \lstinline!inner!\indexinline{inner} which is nearest in the enclosing instance hierarchy of the \lstinline!outer! element declaration.
+An element declared with the prefix \lstinline!outer!\indexinline{outer} references an element instance with the same name but using the prefix \lstinline!inner!\indexinline{inner} which is nearest in the enclosing instance tree of the \lstinline!outer! element declaration.
```
HansOlsson: I think it was problematic before, but the new change adds issues.

A model has only one "instance tree", so "nearest in the enclosing instance tree" seems to roughly mean  "nearest enclosing in the instance tree". However, _roughly_ is doing quite a lot of work, and something like "nearest enclosing"  is needed (but not explained) as it isn't "nearest" in the instance graph (as trees are graphs).

The intent (as we all know) is that you go up the tree (or hierarchy) step-by-step stop as soon as you find a "sibling" that is an inner element with the same name, so using "instance sub-tree" might fix some issues - but it really should be explained in more detail.

HansOlsson: So something like:
```suggestion
An element declared with the prefix \lstinline!outer!\indexinline{outer} references an element instance with the same name but using the prefix \lstinline!inner!\indexinline{inner} whose parent in the instance tree is the closest ancestor of the \lstinline!outer! element declaration, having a matching \indexinline{inner} element.
```

chapters/scoping.tex
```diff
@@ -188,8 +188,8 @@ \section{Inner Declarations - Instance Hierarchy Name Lookup}\label{instance-hie
 An \lstinline!outer! element component may be of a partial class (but the referenced \lstinline!inner! component must be of a non-partial class).
 
 \begin{nonnormative}
-\lstinline!inner!/\lstinline!outer! components may be used to model simple fields, where some physical quantities, such as gravity vector, environment temperature or environment pressure, are accessible from all components in a specific model hierarchy.
-Inner components are accessible throughout the model, if they are not ``shadowed'' by a corresponding \lstinline!inner! declaration in a more deeply nested level of the model hierarchy.
+\lstinline!inner!/\lstinline!outer! components may be used to model simple fields, where some physical quantities, such as gravity vector, environment temperature or environment pressure, are accessible from all components in a specific instance tree.
```
HansOlsson: "specific instance tree" make it sound as if a model may have multiple instance trees, but during translation the model only has one - so I would more say that it is accessible in a sub-tree or part of the tree, and "instance sub-tree" would be a new term, so prefer the latter:
```suggestion
\lstinline!inner!/\lstinline!outer! components may be used to model simple fields, where some physical quantities, such as gravity vector, environment temperature or environment pressure, are accessible from all components in a part of the instance tree.
```

HansOlsson: Or possibly something like "in the instance tree, or part of it".

chapters/scoping.tex
henrikt-ma: That sentence was a tough parse.  Trying to simplify:
```suggestion
An element declared with the prefix \lstinline!outer!\indexinline{outer} references an element instance with the same name but using the prefix \lstinline!inner!\indexinline{inner}.
The \indexinline{inner} element shall be a child of an ancestor of the \lstinline!outer! element in the instance tree, and the match nearest to the \lstinline!outer! element is selected.
```

chapters/scoping.tex
```diff
@@ -162,9 +162,10 @@ \subsection{Lookup of Imported Names}\label{lookup-of-imported-names1}
 See \cref{lookup-of-imported-names}.
 
 
-\section{Inner Declarations - Instance Hierarchy Name Lookup}\label{instance-hierarchy-name-lookup-of-inner-declarations}\label{inner-declarations-instance-hierarchy-name-lookup}
+\section{Inner Declarations - Instance Tree Name Lookup}\label{instance-hierarchy-name-lookup-of-inner-declarations}\label{inner-declarations-instance-hierarchy-name-lookup}
```
HansOlsson:
```suggestion
\section{Inner Declarations - Instance Tree Name Lookup}\label{instance-hierarchy-name-lookup-of-inner-declarations}\label{inner-declarations-instance-tree-name-lookup}
```
Consistent name.

I'm aware that the double-label isn't working as intended (only one is present in html), so adding a third one wouldn't help, but removing the first one would require a change on line 592 as well.

HansOlsson (approved):
Looks good except the label.

henrikt-ma:
The section label is now updated.  I have no problems with this change per se, but could we please give up once and for all on the idea of maintaining multiple labels for the same targets, and just accept that labels can change from one release of the specification to another?  If we want to make labels more long-lived, I think the key is to make them more concise instead of semi-automatically derived from section titles.

Of course it is outside the scope of this PR to make the change.  It should go into a separate PR which would be trivial to prepare with AI assistance.

chapters/scoping.tex
HansOlsson: Given that the multiple labels (as noted) don't really work currently, as there's only one (the last) in the HTML-document I think that would be good.

PR #3920, opened 2026-06-17
Add informal description of 'mixed system iteration'

### LINKED ISSUE #3909: Clarification on fixed point iterations for pre()
Consider the following model:

```
model min_ex
    Real x(start = 3);
    Real z(start = 0);
    Real t;
    Boolean b;
    discrete Real output_state(start = 0);
equation
    b = x < 2;
    x + t + z = 3;
    if b then
        t = 2*time;
    else
        t = time;
    end if;
    if x < 1.9 then
        z = 1;
    else
        z = 0;
    end if;
    when b then
        output_state = x;
    end when;
end min_ex;
```

- Initial solution for `x` is `3`, and decreases as time increases as per` t = time`.
- At `time = 1 (x = 2)`, there will be an event:
  - first: `b` becomes `true` => `t` becomes `2`, new solution: `x = 1`
  - second: `x < 1.9` becomes `false` => `z` becomes `1`, new solution: `x = 0`
  - Note: This example can be expanded to make this an `N` step process.
- The when construct is to capture the value of `x` with the change in `b` (which was also the intention of the modeller), but what is the expected value? 
  - `1` as the value of the iteration when `b` first becomes `true`?
  - `0` as the final value after the value for `x` has been fully resolved? (which was what the modeller expected)
  - Something else?
  - The modeler expected this value to be `0` in this reproducer
- The example can simplified by replacing the use of `b` with `x < 2`. This gives fewer dependencies, while effectively being the same model.

What should be the correct solution here and more importantly: **why?**

Disclaimer: OCT currently yields `output_state = 1` for this model. We suspect this is not the correct result, but it is not evident to us from the specification.

**Specification for pre()** 
[Chapter 3 Operators and Expressions ‣ Modelica® Language Specification version 3.7](https://specification.modelica.org/master/operators-and-expressions.html#modelica:pre)  states a lot of things, but it is still not quite clear:

- What does “mask event iteration“ mean to begin with?
- Is this type of example the non-normative text refers to with “global fix point iteration scheme, similarly to the event iteration“?
  - I.e., a fixed point iteration that resolves the chain of “if” evaluations required to compute “x“, before evaluating the when block for the first time.

Edit: fixed mistake pointed out in comment.

henrikt-ma:
Fixes #3909.

### DIFF
--- a/chapters/operatorsandexpressions.tex
+++ b/chapters/operatorsandexpressions.tex
@@ -1469,7 +1469,7 @@ The first value of \lstinline!pre($y$)! is determined in the initialization phas
 
 After the model equations have been evaluated at an event instant, another round of model equation evaluation is triggered as long as there is at least one variable \lstinline!v! appearing inside \lstinline!pre($\ldots$)! and such that \lstinline!pre(v) <> v!.
 Before each new round of model evaluation, each \lstinline!pre(v)! is assigned the value of \lstinline!v!.
-This evaluation sequence is called \emph{event iteration}.
+This evaluation sequence is called \firstuse{event iteration}\index{mixed system iteration|seealso{event iteration}} (compare \willintroduce{mixed system iteration} described below).
 When no more round of model equation evaluation is triggered, the processing of the event is completed and the transient analysis may proceed forward in time.
 
 For purposes of event iteration, each scalar expression in a \lstinline!when!-clause condition is treated as an implicit variable \lstinline!v!.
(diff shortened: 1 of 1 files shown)


MarkusOlssonModelon (approved):
Looks good.

HansOlsson (approved):
Looks good.

chapters/operatorsandexpressions.tex
```diff
@@ -1482,7 +1482,8 @@ \subsection{Event-Related Operators with Function Syntax}\label{event-related-op
 A consequence of the implicit \lstinline!when!-clause condition variables is that every time a \lstinline!when!-clause is triggered, there is a variable with \lstinline!v <> pre(v)!, meaning that another round of the iteration is triggered.
 
 The language allows mixed algebraic systems of equations where the unknown variables are of type \lstinline!Real!, \lstinline!Integer!, \lstinline!Boolean!, or an enumeration.
-These systems of equations can be solved by a global fix point iteration scheme, similarly to the event iteration, by fixing the \lstinline!Boolean!, \lstinline!Integer!, and/or enumeration unknowns during one iteration.
+A common -- but not standardized -- approach to solving these systems of equations is to apply a fix point iteration scheme, alternating between solving all the \lstinline!Real! and solving all the other variables.
```
```suggestion
A common -- but not standardized -- approach to solving these systems of equations is to apply a fix point iteration scheme, alternating between solving the \lstinline!Real! and the other variables.
```

HansOlsson: I think the "all" is a bit unneeded - but not important.

PR #3922, opened 2026-06-24
A class has elements, not members

henrikt-ma:
This is a cleanup of the use of the term _member_.

To start with, this PR only addresses the issue of calling the elements of a class its _members_ in a couple of places, when the established language to use is to call them _elements_.

Should we keep calling the elements of a record _members_?

### DIFF
--- a/chapters/classes.tex
+++ b/chapters/classes.tex
@@ -10,7 +10,7 @@ Declarations are the syntactic constructs needed to introduce classes and object
 
 \section{Access Control -- Public and Protected Elements}\label{access-control-public-and-protected-elements}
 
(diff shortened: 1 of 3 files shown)


HansOlsson:
> This is a cleanup of the use of the term _member_.
> 
> To start with, this PR only addresses the issue of calling the elements of a class its _members_ in a couple of places, when the established language to use is to call them _elements_.

Good.
 
> Should we keep calling the elements of a record _members_?

Yes, to me "record members" is different as I think of that as a component - not a class.
(Record classes containing class elements is a bit rare, but still exist.)

We could use another name to make that clearer such "record component elements", but it seems a bit clumsy.

HansOlsson:
There's also "named elements" (classes and components), but I don't think we need to stress that dot-notation can only access named elements.

HansOlsson (approved):
LGTM

Issue #3921, opened 2026-06-24
Can a non-package have a package.order?


henrikt-ma:
When working on #3890 I realized that the explanation of _package.order_ only speaks of the ordering of classes and constants, but there is no restriction on the class specialization of the class defined in _package.mo_.  (The only requirement I can see on the class defined in _package.mo_ is that it must have a long class definition in case there are additional files or directories providing more children to the class.)

Since it would be such a mess to define what _package.order_ means for a non-package, can't we just add the restriction that the class in _package.mo_ must be a package when _package.order_ is present?

PR #3923, opened 2026-06-24
Clarify that package.order does not affect visibility

henrikt-ma:
Addressing a small need for clarification encountered in relation to #3921.

### DIFF
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -226,8 +226,9 @@ In order to preserve the order of sub-entities it is advisable to create a file 
 If a \filename{package.order} is present when reading a structured entity the classes and constants are added in this order; if the contents does not exactly match the classes and constants in the package, the resulting order is tool specific and a warning may be given.
 Classes and constants that are stored in \filename{package.mo} are also present in \filename{package.order} but their relative order should be identical to the one in \filename{package.mo} (this ensures that the relative order between classes and constants stored in different ways is preserved).
 
-Only elements defined in \filename{package.mo} are affected by the visibility control (\cref{access-control-public-and-protected-elements}) in \filename{package.mo}; the child classes defined in other files or directories always have public visibility.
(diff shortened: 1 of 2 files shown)


maltelenz (approved):
Seems fine to me.

chapters/packages.tex
```diff
@@ -226,6 +226,9 @@ \subsection{Directory Hierarchy Mapping}\label{mapping-a-package-class-hierarchy
 If a \filename{package.order} is present when reading a structured entity the classes and constants are added in this order; if the contents does not exactly match the classes and constants in the package, the resulting order is tool specific and a warning may be given.
 Classes and constants that are stored in \filename{package.mo} are also present in \filename{package.order} but their relative order should be identical to the one in \filename{package.mo} (this ensures that the relative order between classes and constants stored in different ways is preserved).
 
+Only elements defined in \filename{package.mo} are affected by the visibility control (\cref{access-control-public-and-protected-elements}) in \filename{package.mo}; the child classes defined in other files or directories always have public visibility.
+In particular, reordering classes using \filename{package.order} has no impact on visibility.
```
HansOlsson: The text is fine in itself. However, I had to check what and where it was already present and found:

https://specification.modelica.org/master/class-predefined-types-and-declarations.html#access-control-public-and-protected-elements

> All elements defined under the heading protected are regarded as protected. All other elements (i.e.,
defined under the heading public, without headings or in a separate file) are public (i.e., not protected). 

So, perhaps we should make it clear that it was already defined (not exactly sure how to formulate):
```suggestion
According to \cref{access-control-public-and-protected-elements} only elements defined in \filename{package.mo} are affected by the visibility control in \filename{package.mo}; the child classes defined in other files or directories always have public visibility.
In particular, reordering classes using \filename{package.order} has no impact on visibility.
```

henrikt-ma: I have now reformulated to not make it sound as if the rule is given in both places.  OK?

henrikt-ma:
I have added some additional clarifications, and in particular what this means for the ability to map a class to a separate file or directory.

chapters/packages.tex
HansOlsson: Ok.

PR #3924, opened 2026-06-24
Avoid speaking of file system entities

henrikt-ma:
This PR improves the _File System Mapping of Package/Class_ section by avoiding introduction of the various "entity" terms.  Instead, the new presentation is focused around the well known concepts of files and directories, with the addition of _stored hierarchy_ for the part of the whole file system hierarchy that corresponds to a Modelica class.

This PR is meant to replace #3890, applying the idea that the language group agreed upon in https://github.com/modelica/ModelicaSpecification/pull/3890#issuecomment-4622853310.

### DIFF
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -176,18 +176,15 @@ In this case the search succeeds and the package \lstinline!MyLib.Pack2! is load
 
 \section{File System Mapping of Package/Class}\label{mapping-package-class-structures-to-a-hierarchical-file-system}\label{file-system-mapping-of-package-class}
 
-A package/class hierarchy may be represented in the hierarchical structure of the operating system (the file system).
-For classes with version information see also \cref{mapping-of-versions-to-file-system}.
-The nature of such an external entity falls into one of the following two groups:
+A sub-tree (at any level) of the class tree can be represented in one of two ways in a file system:
 \begin{itemize}
 \item
-  Directory in the file system.
+  A directory containing a file named \filename{package.mo} (\cref{directory-hierarchy-mapping}).
+\item
+  A file with filename extension \filename{mo}, other than \filename{package.mo} (\cref{single-file-mapping}).
 \end{itemize}
 
-\begin{itemize}
-\item
-  File in the file system.
-\end{itemize}
+For classes with version information see also \cref{mapping-of-versions-to-file-system}.
 
 Each Modelica file in the file system is stored in UTF-8 format (defined by The Unicode Consortium; \url{https://unicode.org}).
 A deprecated feature is that the file may start with the UTF-8 encoded BOM\index{BOM (byte order mark)}\index{byte order mark} (byte order mark; \lstinline!0xef 0xbb 0xbf!); this is treated as white-space in the grammar.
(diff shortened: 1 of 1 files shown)


chapters/packages.tex
```diff
@@ -200,48 +197,62 @@ \section{File System Mapping of Package/Class}\label{mapping-package-class-struc
...
+A file defining a non-top-level class shall begin with a \lstinline!within!-clause specifying the parent of the defined class in the class tree.
+When a file defining a class resides next to a \filename{package.mo} (where the parent in the class tree is defined), the class defined in \filename{package.mo} shall have the \productionref{long-class-specifier} form.
+
 See \cref{stored-definitions-containing-multiple-class-definitions} regarding the use of \lstinline!within!-clause when a \productionref{stored-definition} does not hold exactly one class definition.
 
+\begin{nonnormative}
+A non-empty \lstinline!within!-clause only states redundant information, completely determined by the file location in the stored hierarchy rooted in the directory where the \filename{package.mo} has no or an empty \lstinline!within!-clause.
+The redundant information is not only a convenience for tools and humans, it is also information needed to verify that \lstinline!Protection!-annotations (\Cref{modelica:Protection.access}) are not bypassed by moving an encrypted file to a different location in the class tree.
```
HansOlsson:
```suggestion
```
We have currently not defined whether encrypted files can be stored hierarchically, and I could see several issues with that (including the within-issues - but also other ones) that I don't think we should just add it.

henrikt-ma: Isn't this something SEMLA is designed to support?

HansOlsson: > Isn't this something SEMLA is designed to support?

I'll have to investigate that in more detail, and also how to move the standardization of SEMLA forward; but I think we should delay adding this line until it is clearer.

The general issue this discusses is that a user can move encrypted files that are part of a package and that they can still be read in some way.

That seems to imply that the directory is somewhat open for modification (including adding non-encrypted files if I understand SEMLA correctly), and that seems to imply several other potential attacks that we need to guard against.

henrikt-ma: The intention is the opposite: This text clarifies that thanks to the `within` clause requirement, encrypted files cannot be moved, but I realize that the statement could have been formulated differently to make this more clear.  Then, the enforcement of the `Protection`-annotations is just a consequence of not being able to move the classes in the (encrypted) file to a different place of the class tree.

Edit: The updated sentence now puts emphasis on locking the position in the class tree.

chapters/packages.tex
```diff
@@ -200,48 +197,62 @@ \section{File System Mapping of Package/Class}\label{mapping-package-class-struc
 
 \subsection{Directory Hierarchy Mapping}\label{mapping-a-package-class-hierarchy-into-a-directory-hierarchy-structured-entity}\label{directory-hierarchy-mapping}
 
-A directory shall contain a node, the file \filename{package.mo}.
-The node shall contain a \productionref{stored-definition} that defines a class \lstinline!A! with a name matching the name of the structured entity.
+A directory whose \filename{package.mo} contains a \productionref{stored-definition} with an absent or empty \lstinline!within!-clause (\cref{the-within-clause}) is the root of a \firstuse{stored hierarchy} in the file system.
+In a directory which is part of a stored hierarchy, the immediate sub-directories containing a \filename{package.mo} file are also part of the same stored hierarchy.
```
HansOlsson: To me "absent or empty within-clause" would suggest that a package.mo stored as e.g., `Modelica/Dummy/package.mo` without a within-clause would be a new structed entity (separate from Modelica). Additionally, it is not clear what to do if the package.mo doesn't contain a stored-definition.

I tried to correct it as follows - but I'm not fully satisfied.
```suggestion
A directory with \filename{package.mo} and whose parent directory does not have a \filename{package.mo} is the root of a \firstuse{stored hierarchy} in the file system
The \filename{package.mo} shall contain a \productionref{stored-definition} and have an an absent or empty \lstinline!within!-clause (\cref{the-within-clause}).
In a directory which is part of a stored hierarchy, the immediate sub-directories containing a \filename{package.mo} file are also part of the same stored hierarchy.
```

chapters/packages.tex
HansOlsson: Hm... I need to think more about this, I still think it opens up complicated possibilities regarding encrypted packages, and would be happier without this line for now.

henrikt-ma: I agree, and will remove it when I get to a computer.

henrikt-ma: So, now it no longer speaks of encrypted files.

chapters/packages.tex
henrikt-ma: I agree that that there still exist gaps, but I think it might be best to not try to close every single one of them, or the specification will become too difficult to read with no real benefit.  I think that different formulations of these definitions are often going to differ in whether some constellation of files is going to be considered a stored hierarchy with an error, or just something which isn't even considered a stored hierarchy in the first place.  For tools that try to handle errors gracefully, the difference is almost erased entirely.

With this in mind, I think that the alternative you propose looks like an alternative about as good as the replaced rule; both formulations lead to their own behavior in corner cases that I believe won't matter in the end.  According to the proposed rule, the Modelica specification would practically make it an error to have a file named _package.mo_ on your system unless it is a Modelica file, causing issues for GNU gettext files, for example.  If you still think the proposed change is a good idea, I'll apply it – it's no big deal.

HansOlsson: After additional thought, I still think it's a good idea, but I agree that it might be improved even further.
 
For the _compiled_ GNU gettext files I agree it is an issue, but hopefully most of them will not be named "package".

henrikt-ma: To make the specification as non-intrusive as possible, how about combining both requirements for being considered a stored-definition in the first place, so that a gettext _package.mo_ isn't an error according to the Modelica specification?
```suggestion
A directory fulfilling all of the following is the root of a \firstuse{stored hierarchy} in the file system:
\begin{itemize}
\item
  The parent directory does not contain a file named \filename{package.mo}.
\item
  The directory itself contains the file \filename{package.mo}.
\item
  The \filename{package.mo} consists of a \productionref{stored-definition} with an absent or empty \lstinline!within!-clause (\cref{the-within-clause}).
\end{itemize}
In a directory which is part of a stored hierarchy, the immediate sub-directories containing a \filename{package.mo} file are also part of the same stored hierarchy.
```

HansOlsson: Sure, that works.

HansOlsson (approved):

PR #3926, opened 2026-07-03
3.7 accepted by MA.

HansOlsson:
So removing "-rc1".
Also correcting a minor typo, stray `)`, that I noticed earlier.


henrikt-ma (approved):
OK, I guess, but I really think there is something missing in a PR like this when there is no sign of dropping an in-document "rc1" version suffix.

The clean way I suggest that this should be handled next time:
- The `v3.7-rc1` _tag_ forever points to a commit on the _maint/3.7_ branch, where the in-document version is _3.7-rc1_.
- During the voting period, the listed table of versions points to the build of this tag.
- After the release candidate is accepted, a new commit changing the in-document version to _3.7_ is pushed on the _maint/3.7_ branch.
- The new tip of the branch is tagged _v3.7_.
- In the table of versions the row for 3.7-rc1 is replaced by a row for 3.7.

Issue #2239, opened 2018-11-04
Is versionBuild allowed in uses annotation?


modelica-trac-importer:
**Reported by henrikt on 14 Mar 2018 08:41 UTC**
In 18.8.2 _Version Handling_, the `uses` annotation is defined to have the format:
```
uses( IDENT (version = VERSION-NUMBER) )
```

Later, in 18.8.4 _Version Date and Build Information_, there is an example with a model called `M2` where the `uses` annotation also has a `versionBuild` attribute.

Which section is wrong?

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/2239

modelica-trac-importer:
**Modified by beutlich on 14 Mar 2018 10:30 UTC**

modelica-trac-importer:
**Comment by hansolsson on 14 Mar 2018 10:43 UTC**
Replying to [ticket:2239 Henrik Tidefelt]:
> In 18.8.2 _Version Handling_, the `uses` annotation is defined to have the format:
> ```
> uses( IDENT (version = VERSION-NUMBER) )
> ```
> 
> Later, in 18.8.4 _Version Date and Build Information_, there is an example with a model called `M2` where the `uses` annotation also has a `versionBuild` attribute.
> 
> Which section is wrong?

18.8.4 seems correct and 18.8.2 incomplete and should be updated; based on #161 that specified `versionBuild` and that `versionBuild` or `dateModified` in rare cases can be part of `uses`-annotation.

modelica-trac-importer:
**Comment by hansolsson on 20 Mar 2018 13:45 UTC**
Language group:
Clear decision above.
But unclear who added versionBuild in uses-annotation for the specific model that triggered this (no tool wants to admit to adding this).

Issue #161, opened 2018-11-03
Annotations 'versionBuild' and 'versionDate' not documented but should be.


modelica-trac-importer:
**Reported by dietmarw on 14 Apr 2009 09:52 UTC**
I'm not sure if these are real official annotations at all but since they are now used throughout the libraries of the Modelica Association they should get documented or even specified. A proper specification would also avoid certain wrong version and date interpretations as happened in the past.

----

Migrated-From: https://trac.modelica.org/Modelica/ticket/161

modelica-trac-importer:
**Comment by dietmarw on 14 Apr 2009 12:47 UTC**
In case this is going to be put into the specifications I would advise to follow the currently used svn keyword substitution:
```
versionBuild="$Revision: 2437 $"
versionDate="$Date:: 2009-04-10 16:33:14 +0200 #$"
```

This pattern can automatically be generated upon checkout by using the [SVN Keyword substitution mechanism](http://svnbook.red-bean.com/en/1.5/svn.advanced.props.special.keywords.html) like this
```
versionBuild="$Revision$",
versionDate="$Date::                            $",
```
Instead of `Revision` one can also use `Rev` (this is currently used in the MSL but is not as self-explanatory and code should be easy to read right).

One thing that still looks a bit ugly though is that `versionDate` contains a double-colon '::' whereas 'versionBuild' does not. 

I would propose to use **either**:
 a. Specify that both fields contain the double-colon (i.e., the number of characters substituted upon svn-checkout is limited to the number of spaces between **`::`** and **`$`**). 
   * Example of automatic substitution by svn checkout for max 6-digit revision number and iso-date string without the local description in parenthesis):
    ```
versionBuild="$Revision::        $",
versionDate="$Date::                            $",
    ```
    produces
    ```
versionBuild="$Revision:: 2437   $"
versionDate="$Date:: 2009-04-10 16:33:14 +0200 #$"
    ```
 b. Specify that both fields contain only a single colon (i.e., the string field is automatically expanded depending on the string length).
    * Example of automatic substitution by svn checkout:
    ```
versionBuild="$Revision$",
versionDate="$Date$",
    ```
    produces
    ```
versionBuild="$Revision: 2437 $"
versionDate="$Date: 2009-04-10 16:33:14 +0200 (Fr, 10 Apr 2009) $"
    ```
Mind that in the later case the date expression within the parenthesis of `versionDate` depends on the locale language setting and is therefore not a useful to include. 

[...]

PR #3927, opened 2026-07-07
More consistently describe where an annotation is allowed

henrikt-ma:
This addresses most of a problem I noted when an LLM didn't dare drawing conclusions about where the `missingInnerMessage` annotation is allowed.

Currently **not** covered by this PR:
- How to formulate this for the `License` annotation (three different context, two different value types).

### DIFF
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -1706,7 +1706,7 @@ If not specified, the names of new components are tool-specific.
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
 Allowed for class annotations.
-Specifies a whitespace separated list of recommended type prefixes to include in the \productionref{type-prefix} part of a \productionref{component-clause1} generated when creating a component of the class.
+Specifies a white-space separated list of recommended type prefixes to include in the \productionref{type-prefix} part of a \productionref{component-clause1} generated when creating a component of the class.
 
 The following prefixes may be included in the \lstinline!defaultComponentPrefixes! string: \lstinline!inner!, \lstinline!outer!, \lstinline!replaceable!, \lstinline!constant!, \lstinline!parameter!, \lstinline!discrete!.
 The default is an empty string.
--- a/chapters/annotations.tex
+++ b/chapters/annotations.tex
@@ -678,7 +678,6 @@ The annotations listed below define how models can be checked, translated, and s
 \end{center}
 
 \begin{annotationdefinition}[experiment]
-
 \begin{synopsis}
 \begin{lstlisting}[language=modelica]
 record experiment
@@ -690,6 +689,7 @@ end experiment;
 \end{lstlisting}
 \end{synopsis}
 \begin{semantics}
+Allowed for model and block class annotations.
 The \lstinline!experiment! annotation defines the start time (\lstinline!StartTime!) in {[}s{]}, the stop time (\lstinline!StopTime!) in {[}s{]}, the suitable time resolution for the result grid (\lstinline!Interval!) in {[}s{]}, and the relative integration tolerance (\lstinline!Tolerance!) for simulation experiments to be carried out with the model or block at hand.
 When \lstinline!Interval! or \lstinline!Tolerance! is not provided, the tool is responsible for applying appropriate defaults.
 
@@ -708,6 +708,8 @@ Tools can allow users to override these settings without modifying the model.
 /*literal*/ constant Boolean HideResult;
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
+Allowed for component declarations.
+
 \lstinline!HideResult = true! defines that the model developer proposes to not show the simulation results of the corresponding component.
 
 \lstinline!HideResult = false! defines that the developer proposes to show the corresponding component.
@@ -729,6 +731,7 @@ record TestCase
 end TestCase;
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
+Allowed for model and block class annotations.
 If \lstinline!shouldPass! is \lstinline!false! it indicates that the translation or the simulation of the model should fail.
 If a tools checks a package where classes have \lstinline!shouldPass = false! they should not generate errors, and checking may even be skipped.
 On the other hand, models with \lstinline!shouldPass = false! may be useful for creation of negative tests in tool-specific ways.
@@ -1660,6 +1663,7 @@ The annotations listed below define properties for use in graphical user interfa
 /*literal*/ constant String preferredView;
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
+Allowed for class annotations.
 The \lstinline{preferredView} annotation defines the default view when selecting the class.
 The value \lstinline!"info"! means class documentation (``information''), \lstinline!"diagram"! means diagram view, \lstinline!"icon"! means icon view, and \lstinline!"text"! means Modelica source code (``text'').
 If not specified the default view is tool specific.
@@ -1671,7 +1675,8 @@ If not specified the default view is tool specific.
 /*literal*/ constant Boolean DocumentationClass;
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
-Only allowed as class annotation on any kind of class and only having effect when \lstinline!true!, meaning that this class and all classes within it are treated as having the annotation \lstinline!preferredView = "info"!.
+Allowed for class annotations.
+Only having effect when \lstinline!true!, meaning that this class and all classes within it are treated as having the annotation \lstinline!preferredView = "info"!.
 If the annotation \lstinline!preferredView! is explicitly set for a class, it has precedence over a \lstinline!DocumentationClass! annotation.
 
 \begin{nonnormative}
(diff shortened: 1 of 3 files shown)


chapters/annotations.tex
```diff
@@ -2454,8 +2465,7 @@ \subsection{Versions in the File System}\label{mapping-of-versions-to-file-syste
 \subsection{Version Date and Build Information}\label{version-date-and-build-information}
 
 This section describes annotations that a top-level class can have to specify information associated to the version number.
-
-The \lstinline!versionBuild! and \lstinline!dateModified! annotations can also be specified in the \lstinline!uses! annotation (together with the version number).
+Some of the annotations can also be specified in the \lstinline!uses! annotation.
```
HansOlsson, line 2468:
```suggestion
Some of the annotations can also be specified in the \lstinline!uses! annotation, as stated below.
```
Or something like that.

Compared to the previous text it seemed a bit vague (which ones?)
I understand that since it is given for the specific annotations it is better to _only_ give it there - but make that clear for users.

chapters/functions.tex
```diff
@@ -1144,7 +1144,8 @@ \section{Derivatives and Inverses of Functions}\label{derivatives-and-inverses-o
    IDENT "=" type-specifier function-call-args"
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
-A function with one output formal parameter may have one or more {\lstinline!inverse!} annotations to define inverses of this function.
+Allowed for function class annotations when the function has exactly one output formal parameter.
```
HansOlsson, line 1147: A side-note: At some point we might want to generalize inverses to multiple outputs - but not now. As I understand it one issue is the syntax.

chapters/inheritance.tex
```diff
@@ -1317,7 +1317,7 @@ \section{Selective Model Extension}\label{selective-model-extension}\index{desel
     for i in 2:10 loop
       connect( // This comment does not impact syntactic equivalence.
         a[i],
-        b[2*i] /* Without whitespace in the indexing expression. */ );
+        b[2*i] /* Without white-space in the indexing expression. */ );
```
HansOlsson, line 1320: I was unsure if that was needed, and searched for "white-space vs whitespace" which gave:
- https://en.wikipedia.org/wiki/Whitespace_character
- https://github.com/orgs/mdn/discussions/382

And the CoPilot summary:

> Use "whitespace" as a single word in computing contexts and "white space" as two words in typography or general design.
> 
> Usage in Computing
> In programming and computer science, "whitespace" refers to characters that create blank space in text, such as spaces, tabs, and line breaks. These characters affect text layout but are not visible themselves, and the term is typically written as a single word when used attributively, for example, "whitespace character" or "remove whitespace from a string" 
> Standards like the HTML and CSS specifications consistently use "whitespace" to refer to these characters

(Later note: That summary wasn't fully correct, HTML talks about "whitespace", but CSS 2.1 discusses "white space" while the property is called "white-space". CSS 1 also had the property "white-space" - but discussed "whitespace". It could even be that hyphen in the property-name is to make it easier to discuss whitespace as a concept vs. the property white-space.)

Thus I think we might as well keep "whitespace".
```suggestion
        b[2*i] /* Without whitespace in the indexing expression. */ );
```

chapters/annotations.tex
```diff
@@ -1699,7 +1705,8 @@ \section{Graphical User Interface}\label{annotations-for-the-graphical-user-inte
 /*literal*/ constant String defaultComponentPrefixes;
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
-The class annotation \lstinline!defaultComponentPrefixes! gives a whitespace separated list of recommended type prefixes to include in the \productionref{type-prefix} part of a \productionref{component-clause1} generated when creating a component of the class.
+Allowed for class annotations.
+Specifies a white-space separated list of recommended type prefixes to include in the \productionref{type-prefix} part of a \productionref{component-clause1} generated when creating a component of the class.
```
HansOlsson, line 1709:
```suggestion
Specifies a whitespace separated list of recommended type prefixes to include in the \productionref{type-prefix} part of a \productionref{component-clause1} generated when creating a component of the class.
```
So, no change here either.

chapters/inheritance.tex
henrikt-ma: Sounds good, but then we have a bunch of other places where this should be aligned.

chapters/functions.tex
henrikt-ma: Sure, I see no problem with returning to this possible question of generalization in the future.  Please resolve if there is nothing more to be said for now.

chapters/annotations.tex
henrikt-ma: The entire paragraph introducing this subsection reads like this:
>This section describes annotations that a top-level class can have to specify information associated to the version number.
>Some of the annotations can also be specified in the \lstinline!uses! annotation.

For me, it goes without saying that it talks about what is stated below, and re-stating this would be ugly.

HansOlsson: I agree that it talks about what is given below. **But** it isn't obvious that the specific annotations that can be on a uses-annotations are specified below for the specific annotations, and e.g., not listed under the uses-annotation.

Just something like the following may be enough to indicate that the information is included here, and not somewhere else. Something like the following may make that clearer without restating anything:
```suggestion
Including whether they can also be specified in the \lstinline!uses! annotation.
```

henrikt-ma: But it _is_ also stated for the `uses`-annotation itself, so we don't need this introductory paragraph to explain in detail where to find the information.

I think we need to remember that this is just the informal opening paragraph to the section, the consistent way of describing where each annotation may be used is the real solution to the problem.  Would it be better to open the section with a paragraph which doesn't speak at all of where the annotations may be used?

HansOlsson: My first reaction was that it _may_ work; and is preferable to a vague "some of ...". The important part is figuring out why it was added, and see whether it is still relevant.

As far as I can see the relevant issues are:
https://github.com/modelica/ModelicaSpecification/issues/2239
https://github.com/modelica/ModelicaSpecification/issues/161

It doesn't seem to indicate that it was added due to some clear need for that in itself, **but only for what follows**.
So, yes, we can remove it:

**But**, there is a clear need for what follows, and thus line 2461/2471 should clearly be updated as it currently says:
> It is recommended that tools do not automatically store \lstinline!versionBuild! and \lstinline!dateModified! in the \lstinline!uses! annotation.

which is odd to add if we haven't stated that any annotations can occur there at all (and also an odd way to introduce the specific ones).

So, if we don't mention them at the start I believe it is best to move&duplicate that line to versionBuild and dateModified.

Additionally:

Line 2175/2186 
> See \cref{version-date-and-build-information} regarding the use of \lstinline!versionBuild! and \lstinline!dateModified!.

I believe that should be replaced by two separate references, as the changes above would move the text it was referencing.

chapters/annotations.tex
```diff
@@ -2170,6 +2180,7 @@ \subsection{Version Handling}\label{version-handling}
   ")"
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
+Allowed for class annotations of top-level classes.
 \lstinline!uses($\mathit{otherPackage}$(version = $\mathit{otherPackageVersion}$))! defines that classes within this top-level class use the $\mathit{otherPackageVersion}$ of classes within the top-level class $\mathit{otherPackage}$.
 
 See \cref{version-date-and-build-information} regarding the use of \lstinline!versionBuild! and \lstinline!dateModified!.
```
HansOlsson, line 2186:
```suggestion
See \cref{versionBuild} and \cref{dateModified} regarding the use of \lstinline!versionBuild! and  \lstinline!dateModified!, respectively.
```
As indicated in https://github.com/modelica/ModelicaSpecification/pull/3927/changes#r3914643884

chapters/annotations.tex
HansOlsson, line 2472:
```suggestion
```
As suggested in https://github.com/modelica/ModelicaSpecification/pull/3927/changes#r3914643884

HansOlsson, line 2503:
```suggestion
In special cases, the \lstinline!uses!-clause of a model may specify \lstinline!versionBuild! and/or \lstinline!dateModified!.
In such a case the tool is expected to give a warning if there is a mismatch between library and model.
It is recommended that tools do not automatically store \lstinline!versionBuild! and \lstinline!dateModified! in the \lstinline!uses! annotation.
```
Based on suggestion in https://github.com/modelica/ModelicaSpecification/pull/3927/changes#r3914643884

chapters/annotations.tex
```diff
@@ -2497,6 +2509,7 @@ \subsection{Version Date and Build Information}\label{version-date-and-build-inf
 /*literal*/ constant String dateModified
 \end{lstlisting}\end{synopsis}
 \begin{semantics}
+Allowed for class annotations of top-level classes, and in the \lstinline!uses! annotation (together with the version number).
```
HansOlsson, line 2512:
```suggestion
Allowed for class annotations of top-level classes, and in the \lstinline!uses! annotation (together with the version number).
For the latter case see \cref{versionBuild}.
```

HansOlsson (changes requested):
As noted the dateModified and versionBuild changes are incomplete without updates in other places.

PR #3928, opened 2026-07-15
Define that package.order requires a package

### LINKED ISSUE #3921: Can a non-package have a package.order?

HansOlsson:
and that only packages should be stored as package.mo.
Closes #3921 

Note: I generalized it to require that even package.mo should (so, there might be exceptions) be a package. We might revisit that, but I think it is a good general rule.
I agree that having package.order for a non-package would be really problematic - as there are non-constants, equations etc that must be ordered.

Having package.mo for a non-package would to me only make sense for a top-level class where you want to store resources as if it were a package, and in that case having sub-classes (and especially sub-classes stored externally) doesn't really make sense.

### DIFF
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -238,7 +238,7 @@ A directory shall not contain both the sub-directory \filename{A} (containing a 
 \end{example}
 
 In order to preserve the order of classes and constants in package, it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
-If a \filename{package.order} is present when reading a directory, the classes and constants are added in this order; if the contents does not exactly match the classes and constants in the package, the resulting order is tool specific and a warning may be given.
+If a \filename{package.order} is present when reading a directory for a package, the classes and constants are added in this order; if the contents does not exactly match the classes and constants in the package, the resulting order is tool specific and a warning may be given.
 Classes and constants that are stored in \filename{package.mo} are also present in \filename{package.order} but their relative order should be identical to the one in \filename{package.mo} (this ensures that the relative order between classes and constants stored in different ways is preserved).
 
 As defined in \cref{access-control-public-and-protected-elements}, the effect of the visibility headings \lstinline!public! and \lstinline!protected! appearing in \filename{package.mo} only affect visibility of elements defined in that file; the child classes defined in other files or directories always have public visibility.
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -237,7 +237,7 @@ All children of the class defined in \filename{package.mo} (including those defi
 A directory shall not contain both the sub-directory \filename{A} (containing a \filename{package.mo}) and the file \filename{A.mo}.
 \end{example}
 
-In order to preserve the order of classes and constants in package, it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
+In order to preserve the order of classes and constants in a package, it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
 If a \filename{package.order} is present when reading a directory for a package, the classes and constants are added in this order; if the contents does not exactly match the classes and constants in the package, the resulting order is tool specific and a warning may be given.
 Classes and constants that are stored in \filename{package.mo} are also present in \filename{package.order} but their relative order should be identical to the one in \filename{package.mo} (this ensures that the relative order between classes and constants stored in different ways is preserved).
 
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -216,6 +216,7 @@ Any other directory in the stored hierarchy defines a child of the class defined
 By construction, a directory (e.g., the directory \filename{P}) which is part of a stored hierarchy will contain the file \filename{package.mo}.
 This file shall contain a \productionref{stored-definition}, defining a single class (possibly with nested classes inside).
 The name of the class (here, \lstinline!P!) shall match the name of the directory.
+The class should be a package.
 The definition in \filename{package.mo} defines all parts of the sub-tree corresponding to the directory, except zero or more direct child classes (defined by other files or sub-directories).
 
 \begin{nonnormative}
@@ -236,7 +237,7 @@ All children of the class defined in \filename{package.mo} (including those defi
 A directory shall not contain both the sub-directory \filename{A} (containing a \filename{package.mo}) and the file \filename{A.mo}.
 \end{example}
 
-In order to preserve the order of classes and constants, it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
+In order to preserve the order of classes and constants in package, it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
 If a \filename{package.order} is present when reading a directory, the classes and constants are added in this order; if the contents does not exactly match the classes and constants in the package, the resulting order is tool specific and a warning may be given.
 Classes and constants that are stored in \filename{package.mo} are also present in \filename{package.order} but their relative order should be identical to the one in \filename{package.mo} (this ensures that the relative order between classes and constants stored in different ways is preserved).
 
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -216,11 +216,11 @@ Any other directory in the stored hierarchy defines a child of the class defined
 By construction, a directory (e.g., the directory \filename{P}) which is part of a stored hierarchy will contain the file \filename{package.mo}.
 This file shall contain a \productionref{stored-definition}, defining a single class (possibly with nested classes inside).
 The name of the class (here, \lstinline!P!) shall match the name of the directory.
-The class should be a package.
 The definition in \filename{package.mo} defines all parts of the sub-tree corresponding to the directory, except zero or more direct child classes (defined by other files or sub-directories).
 
 \begin{nonnormative}
 The \filename{package.mo} typically contains documentation and graphical information for a package, but may also contain additional elements of the class \lstinline!P!.
+Typically, the class defined in \filename{package.mo} will be a package, but there are also situations where classes of other specializations may benefit from being stored as a directory with a \filename{package.mo} inside.
 \end{nonnormative}
 
 Each other file with filename extension \filename{mo} in a directory of a stored hierarchy is also part of the same stored hierarchy.
--- a/chapters/packages.tex
+++ b/chapters/packages.tex
@@ -286,28 +286,32 @@ It follows that any \lstinline!within!-clause (\cref{the-within-clause}) in such
 \section{External Resources}\label{external-resources}
 
 Examples of references to external resources include links and images in HTML documentation, and images in the \lstinline!Bitmap! annotation (see \cref{bitmap}).
-Absolute URIs should be used, for example \filename{file:///} and the URI scheme \filename{modelica:/} which can be used to retrieve resources associated with a package.
+Absolute URIs should be used, for example \filename{file:///} and the URI scheme \filename{modelica:/} which can be used to retrieve resources associated with a class.
 According to the URI specification scheme names are case-insensitive, but the lower-case form should be used, that is \filename{Modelica:/} is allowed but \filename{modelica:/} is the recommended form.
 
-The Modelica-scheme\index{Modelica!URI scheme}\index{URI!Modelica} has the ability to reference a hierarchical structure of resources associated with packages.
+\begin{nonnormative}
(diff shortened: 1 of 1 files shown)


chapters/packages.tex
```diff
@@ -236,8 +237,8 @@ \subsection{Directory Hierarchy Mapping}\label{mapping-a-package-class-hierarchy
 A directory shall not contain both the sub-directory \filename{A} (containing a \filename{package.mo}) and the file \filename{A.mo}.
 \end{example}
 
-In order to preserve the order of classes and constants, it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
-If a \filename{package.order} is present when reading a directory, the classes and constants are added in this order; if the contents does not exactly match the classes and constants in the package, the resulting order is tool specific and a warning may be given.
+In order to preserve the order of classes and constants in package, it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
```
maltelenz:
```suggestion
In order to preserve the order of classes and constants in a package, it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
```

henrikt-ma:
> Having package.mo for a non-package would to me only make sense for a top-level class where you want to store resources as if it were a package, and in that case having sub-classes (and especially sub-classes stored externally) doesn't really make sense.

I can see that it would be an inconsistency to allow _package.mo_ for a class where it is not allowed to have a corresponding _package.order_.  However, I think there are good use cases also for storing non-packages in a directory with one or more .mo files inside:
- When a function is only intended to be used by a particular model, I like to place the function within the model's definition, and if the function definition is big it can be convenient to define it in a separate file.
- When creating stand-alone examples, it is often convenient to make the top-level class the model to be simulated, with all helper classes defined inside.  In this case, the model acts as a kind of package with a predefined model to simulate, and not allowing it to be stored across multiple files would feel like an unnecessary restriction.
- ~I find it very useful to have external resources stored along with the class using them, and external resources are absolutely not something primarily of interest for packages.  That is, if _img.png_ belongs to `P.A`, I prefer to reference it as _modelica:/P.A/img.png_, but this means that _P/A_ is a directory, and then I think _P/A/package.mo_ is the more natural place for the class definition.~

Edit: @maltelenz just pointed out that https://specification.modelica.org/master/packages.html#external-resources speaks of _**package** name_, so I am striking out the item about regarding external resources attached to non-packages.

chapters/packages.tex
```diff
@@ -216,6 +216,7 @@ \subsection{Directory Hierarchy Mapping}\label{mapping-a-package-class-hierarchy
 By construction, a directory (e.g., the directory \filename{P}) which is part of a stored hierarchy will contain the file \filename{package.mo}.
 This file shall contain a \productionref{stored-definition}, defining a single class (possibly with nested classes inside).
 The name of the class (here, \lstinline!P!) shall match the name of the directory.
+The class should be a package.
```
henrikt-ma: As long as there are good use cases also for non-packages, I don't think we should use strong recommendations like this.  A weaker non-normative comment would be more suitable, but if it is only a weak recommendation I am not sure it belongs in the specification at all.

chapters/packages.tex
henrikt-ma, line 224: Alternative to the above:
```suggestion
Typically, the class defined in \filename{package.mo} will be a package, but there are also situations where classes of other specializations may benefit from being stored as a directory with a \filename{package.mo} inside.
\end{nonnormative}
```

chapters/packages.tex
henrikt-ma: A non-normative replacement has now been suggested below.

```suggestion
```

chapters/packages.tex
```diff
@@ -236,8 +237,8 @@ \subsection{Directory Hierarchy Mapping}\label{mapping-a-package-class-hierarchy
...
 \end{example}
 
-In order to preserve the order of classes and constants, it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
-If a \filename{package.order} is present when reading a directory, the classes and constants are added in this order; if the contents does not exactly match the classes and constants in the package, the resulting order is tool specific and a warning may be given.
+In order to preserve the order of classes and constants in package, it is advisable to create a file \filename{package.order}\index{package.order@\filename{package.order}} where each line contains the name of one class or constant (using its Modelica \lstinline!IDENT! form).
+If a \filename{package.order} is present when reading a directory for a package, the classes and constants are added in this order; if the contents does not exactly match the classes and constants in the package, the resulting order is tool specific and a warning may be given.
```
henrikt-ma, line 241: Don't we want to be more restrictive and actually ban the presence of _package.order_ for non-packages?

henrikt-ma (changes requested):
More discussion is needed regarding the status of _package.mo_ for non-packages.

chapters/packages.tex
gkurzbach: I would agree here.

chapters/packages.tex
gkurzbach: If we allow a directory structure for non package classes, probably also a package.order file is needed for that.

henrikt-ma: The current support for package constants is of course easy to generalize to non-constant components, but the whole point of this PR is that the _package.order_ idea is too difficult to apply to classes with equations.

HansOlsson: To me the relevant case other case is storing a top-level non-package in package.mo to have the Resources-directory as though it were a package; so we only want the directory structure for the Resources-directory not for storing local classes.

I would even say that there is no need to store any potential sub-classes separately for that case, and it's even a bit unusual to have a non-short local class inside a model; and if it is a short class it cannot be stored in a separate file.

henrikt-ma: > To me the relevant case other case is storing a top-level non-package in package.mo to have the Resources-directory as though it were a package; so we only want the directory structure for the Resources-directory not for storing local classes.

This relates to the item I stroke out in https://github.com/modelica/ModelicaSpecification/pull/3928#issuecomment-5279267109 above since the external resource text very clearly speaks of _packages_, not classes defined in a _package.mo_.  I this something we should revisit as part of this PR?

HansOlsson: Yes, I'll add the minimal part.

HansOlsson: Now added

chapters/packages.tex
```diff
@@ -285,28 +286,32 @@ \section{Stored Definitions Containing Multiple Class Definitions}\label{stored-
...
 These should be rewritten as \filename{modelica:/Modelica/Resources/C.jpg} and \filename{modelica:/Modelica.Blocks}.
 \end{nonnormative}
 
-When Modelica packages are stored hierarchically in a file system (i.e., package \lstinline!A! in a directory \filename{A} containing \filename{package.mo}) the resource
+When Modelica classes are stored hierarchically in a file system (i.e., package \lstinline!A! in a directory \filename{A} containing \filename{package.mo}) the resource
 \filename{modelica:/A/Resources/C.jpg} should be stored in the file \filename{A/Resources/C.jpg}, it is not recommend to use \filename{modelica:/A.B/C.jpg} for referencing resources; it could be stored in the file \filename{A/B/C.jpg} -- which is counter-intuitive if \lstinline!A.B! is stored together with \lstinline!A!.
-When Modelica packages are stored in other formats a similar mapping should be defined, such that a resource with a given path should be unique for any package name that precedes it.
+When Modelica classes are stored in other formats a similar mapping should be defined, such that a resource with a given path should be unique for any package name that precedes it.
```
henrikt-ma: Shouldn't this also be changed?


```suggestion
When Modelica classes are stored in other formats a similar mapping should be defined, such that a resource with a given path should be unique for any class name that precedes it.
```

HansOlsson: Yes.

HansOlsson:
Language meeting:

Goal:
- Allow package.mo for non-package class (for resources)
- Forbid package.order for non-package (as it defines location of child-classes that are stored in other files; and that is rare for a non-package).
- At least no need for a package.order if there's only package.mo; cannot be used to modify text-view of a model.
- Markus: Only allow sub-mo-files if non-package?

Agreement?: Most in favor, but Gerd thinks that package.order could work; so allow

Off-line poll->Hans

HansOlsson:
Background:
- Packages stored as `package.mo` may (and are advised) to have `package.order` to give classes stored in separate files/directories a sensible order.
- There is some reasons to have a (top-level) model stored in a `package.mo` (in a directory with a name matching the model) - since that allows resources to be stored in a sensible way in the directory.
- The current text for `package.order` doesn't work well for non-packages (it assumes there are only classes and constants; we already have the issue with extends-clauses #3438, it is more common here, and additionally we also have variables, equations, algorithms, and for functions external-clauses. Note that annotations aren't an issue (anymore) as they should be last.)

Poll for variants among MAP-Lang members:
- Storing sub-classes in separate files and/or directories (together with `package.mo`) is only allowed for packages. Using `package.order` to order them is only allowed for packages. :heart:
- Storing sub-classes in separate files and/or directories (together with `package.mo`) is only allowed for packages. Having a `package.order` is allowed, but it has no practical effect for non-packages. :rocket:
- Storing sub-classes in separate files and/or directories is allowed also for non-packages stored as `package.mo`. For the description of `package.order` change constants to components, and in case of ambiguity equations/algorithms are after sub-classes. 🎉 
- Storing sub-classes in separate files and/or directories is allowed also for non-packages stored as `package.mo`. The package.order only includes child classes and package constants (not constant components declared in non-packages), and is only intended for ordering items in a tree view of the class tree. (Added) 👀 

Select one (or possibly more) reaction(s) from "Add or Remove Reaction" (below to the left).
Poll ends September 18th.

henrikt-ma:
The poll above does not include the following liberal and backward compatible alternative discussed at the meeting:
- Storing sub-classes in separate files and/or directories is allowed also for non-packages stored as `package.mo`. The `package.order` only covers child classes and **package** constants (not constant components declared in non-packages), and is only intended for ordering items in a tree view of the class tree.

HansOlsson:
> The poll above does not include the following liberal and backward compatible alternative discussed at the meeting:
> 
> * Storing sub-classes in separate files and/or directories is allowed also for non-packages stored as `package.mo`. The `package.order` only covers child classes and **package** constants (not constant components declared in non-packages), and is only intended for ordering items in a tree view of the class tree.

I must have missed that idea, but added it. 

However, note that:
- The normal `package.order` for packages allows a consistent textual view of the entire package - that is not possible for this case, and in this case the order between sub-classes and other elements will vary between tools.
- Unless the child classes are encapsulated they are not accessible from the outside; and even if encapsulated it is not clear whether tools will show them in the class view.

henrikt-ma:
With ❤️ and 🚀, imagine that `FMU_A` and `FMU_B` are imported FMUs.  When organized in a package, resources _can_ be placed next to the wrapper models:
```
- MyLib
  - package.mo
  - ImportedFMUs
    - package.mo
    - FMU_A.mo
    - FMU_A
      - Resources
        - modelDescription.xml
        - binaries
        - documentation
        - LICENSE.md
        - …
    - FMU_B.mo
    - FMU_B
      - Resources
        - …
```

However, the same pattern doesn't work when an FMU is imported as a top-level class, because then the directory corresponding to, say, _FMU_A_ would be the top-level directory of the package, meaning that _FMU_A.mo_ will sitt side by with with _Resources_:
```
FMU_A
- FMU_A.mo
- Resources
  - …
```
Hence, it gets a bit messy to move an imported FMU from the original top-level location into a package like `ImportedFMUs`, because the shape in the file system needs to change.

With 🎉 and 👀, the same pattern can be used both at the top-level and inside packages.  Top-level:
```
FMU_A
- package.mo
- Resources
  - modelDescription.xml
  - …
```
Inside package:
```
ImportedFMUs
- FMU_A
  - package.mo
  - Resources
    - ...
```

HansOlsson:
Was there a misunderstanding of "Storing sub-classes in separate files and/or directories (together with package.mo) is only allowed for packages."? The parenthesis can be skipped and the idea was:
- A non-package class can be stored in package.mo (even if not a good idea).
- However, sub-classes cannot be stored separately for a non-class.

Added: I know that there were a discussion on restricting non-package `package.mo` to the top-level, originally I favored that - but there was some push-back, and since non-package `package.mo` is already a bit odd I think it was concluded to not add that restriction.

henrikt-ma:
> Was there a misunderstanding of "Storing sub-classes in separate files and/or directories (together with package.mo) is only allowed for packages."? The parenthesis can be skipped and the idea was:

Don't know if anybody misunderstood, but skipping it sounds like a good idea.

> * A non-package class can be stored in package.mo (even if not a good idea).

I think the imported FMU example shows clearly that it can be a good idea.

PR #3935, opened 2026-09-08
Introduce 'nullCombination' and 'affineCombination'

henrikt-ma:
Driven by #3748, this PR introduces two new built-in functions which we are quite far from being possible to express as user-defined functions.  By having these in the language, the discussion about how to infer `absoluteValue` can avoid getting stuck on the issues that these two functions solve cleanly.

The concepts captured by these functions are well established in mathematics, and represented in several other software environments where the distinction between absolute and relative quantities is of importance.  Adding them to the language as a preparation for #3748 is similar to how we have recently added language features to prepare for good unit checking rules, such as semantics for raising to an integer power, rational exponents of units, the `nthRoot`, and unitful literals.

The examples demonstrate how the functions can be used with (non-coherent) absolute Celsius temperatures, but in principle they are equally important when working with absolute quantities of any unit, coherent or not.

When combined with future unit semantics, these functions will provide a powerful combination of features I didn't see in any of the other softwares in my survey:
- Static (translation-time) analysis of absolute vs relative.
- Possibility to consider weight variability when checking validity of the sum of the weights.
- Interplay with units in two different but meaningful ways for the two functions.

### Possible extension: homothety

The PR currently does *not* include a related operation in the same family, namely homothety, which could be defined as a syntactic sugar:
```
homothety(k, p, center = c) := affineCombination({1 - k, k}, {c, p})
```
The form of an operator where the center of the transformation is required to be given as a named argument would clarify intent when expressing things like "today the temperature is twice as warm as yesterday":
```
tToday = homothety(2, tYesterday, center = T_ice)
```
(I refuse to use the ridiculously named `Constants.T_zero`; the `T_ice` is just an alternative name which makes the expression above look more meaningful.)

In case the name `homothety` would be rejected because it sounds too similar to `homotopy`, alternative name candidates include `dilation` and `scale`.  With these less specific names, it could be argued that the operator should also allow scaling a relative quantity, and then the presence of `center` would mark the difference between operating on relative or absolute quantities.

### DIFF
--- a/chapters/operatorsandexpressions.tex
+++ b/chapters/operatorsandexpressions.tex
@@ -947,6 +947,8 @@ The operators listed below include the derivative operator and special purpose o
 {\lstinline!der($\mathit{expr}$)!} & Time derivative & \Cref{modelica:der} \\
 {\lstinline!cardinality($c$)!} & Number of occurrences in {\lstinline!connect!}-equations & \Cref{modelica:cardinality} \\
 {\lstinline!homotopy($\mathit{actual}$, $\mathit{simplified}$)!} & Homotopy initialization & \Cref{modelica:homotopy} \\
+{\lstinline!nullCombination($\mathit{weights}$, $\mathit{items}$)!} & Null combination & \Cref{modelica:nullCombination} \\
+{\lstinline!affineCombination($\mathit{weights}$, $\mathit{items}$)!} & Affine combination & \Cref{modelica:affineCombination} \\
 {\lstinline!semiLinear($x$, $k^{+}$, $k^{-}$)!} & Sign-dependent slope & \Cref{modelica:semiLinear} \\
 {\lstinline!inStream($v$)!} & Stream variable flow into component & \Cref{modelica:inStream} \\
 {\lstinline!actualStream($v$)!} & Actual value of stream variable & \Cref{modelica:actualStream} \\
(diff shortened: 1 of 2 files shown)


HansOlsson:
We need to discuss this at the next meeting.

An alternative, which I believe MapleSoft had implemented at least for affine combination, would be to support this without any new functions. E.g., stating that `sum(w.*x)/sum(w)` and/or `sum(w*x)/sum(w)` is an affine combination of the vector `x` provided the same variable/parameter/constant `w` is used in both places.

Thus, one question is how users view having a new special function `affineCombination` (both in terms of reading code and writing it) compared to being required to write such calls in special ways (assuming we don't handle possible all cases).

Another is how complicated it is to detect such special cases for tools. Note that finding those special cases is important even if we introduce `affineCombination` in order to find cases where users should have used it.

In many (almost all?) cases the affine combination will be a convex combination; should that be noted?

For media there are a number convex combinations where we for species concentrations have two different formulations, either `sum(X)=1` or `cat(1,X,{1-sum(X))` . I don't know if we need to consider something special for either of these cases.

Looking at `Modelica.Media` I expected to find uses, but the first somewhat related case I found was `Modelica.Media.Interfaces.PartialMixtureMedium.moleToMassFractions` which isn't an affine combination - but the weights for it computed as: `moleFractions.*MMX/sum(moleFractions.*MMX)`, which seems like a variant to also consider.

However, Media also have more normal cases like:
- `Modelica.Media.IdealGases.Common.MixtureGasNasa.h_TX` 
- `Modelica.Media.IdealGases.Common.MixtureGasNasa.specificHeatCapacityCp` 

Extra:
I focus on affine combinations as I don't know how common null combinations are in practice, except the trivial ones. Clearly it should work to write `a-b` instead of `nullCombination({1,-1},{a,b})`.

Note that we be should be that using `affineCombination` (or similarly) doesn't require/make that the physics correspond to an affine (sub-)space.

henrikt-ma:
> An alternative, which I believe MapleSoft had implemented at least for affine combination, would be to support this without any new functions. E.g., stating that `sum(w.*x)/sum(w)` and/or `sum(w*x)/sum(w)` is an affine combination of the vector `x` provided the same variable/parameter/constant `w` is used in both places.
> 
> Thus, one question is how users view having a new special function `affineCombination` (both in terms of reading code and writing it) compared to being required to write such calls in special ways (assuming we don't handle possible all cases).

I am familiar with their design.  Their system builds upon a runtime tally of the weights in a linear combination of absolute temperatures, and when the weights happen to add up to zero, they treat the linear combination as a null combination, and when they add up to one they treat the linear combination as an affine combination.  When the weights add up to neither they paint the sum in red to indicate that it is meaningless work in progress or simply some sort of error in the program.  It is a fragile solution, and that it relies on runtime tracking of the sum of weights makes it impossible to analyze the "`absoluteValue` semantics" at translation time.  That they at least made an effort to address this need is nice to see, but their design is not one that Modelica should follow.

henrikt-ma:
> Another is how complicated it is to detect such special cases for tools. Note that finding those special cases is important even if we introduce `affineCombination` in order to find cases where users should have used it.

Checking `absoluteValue` consistency similar to unit consistency will point at problematic places where `affineCombination` is needed to express an absolute quantity (temperature).  Checking `absoluteValue` is less conclusive in nature than checking the units, but in models making adequate use of temperature types with `absoluteValue` annotations defined, I expect the problematic places to be detected.

> In many (almost all?) cases the affine combination will be a convex combination; should that be noted?

The `homothety` extension mentioned in the opening comment is an example of a non-convex combination, but I wouldn't be against adding a note about the relation to convex combinations.

henrikt-ma:
> For media there are a number convex combinations where we for species concentrations have two different formulations, either `sum(X)=1` or `cat(1,X,{1-sum(X))` . I don't know if we need to consider something special for either of these cases.

The `cat(1,X,{1-sum(X))` idea was considered as an alternative to the proposed normalization, but rejected for at least two reaons:
- Lack of symmetry with the items being combined: `affineCombination({0.2, 0.4}, {a, b, c})`
- Weights must have unit `"1"` in order for the `1-sum(X)` to make sense, while the normalization supports common cases of weights with units such as mass or mass flow.

henrikt-ma:
> Looking at `Modelica.Media` I expected to find uses, but the first somewhat related case I found was `Modelica.Media.Interfaces.PartialMixtureMedium.moleToMassFractions` which isn't an affine combination - but the weights for it computed as: `moleFractions.*MMX/sum(moleFractions.*MMX)`, which seems like a variant to also consider.

This looks like a different operation, expressed in a bad way.  Shouldn't it simply be written like so?
```
output Modelica.Units.SI.MassFraction X[size(moleFractions, 1)] =
  (moleFractions .* MMX) / (moleFractions * MMX)
  "Mass fractions of gas mixture";
```

Edit: The more relevant thing I should have said regarding this function is that it operates on quantities which are always measured on ratio scales (units where 0 means absence of the quantity). Hence, these computations don't really put the `absoluteValue` inference problem to the test, as a failure to infer is not a threat to unit inference and unit conversion.

henrikt-ma:
> Extra: I focus on affine combinations as I don't know how common null combinations are in practice, except the trivial ones. Clearly it should work to write `a-b` instead of `nullCombination({1,-1},{a,b})`.

Of course a difference should be written as `a - b`.

PR #3937, opened 2026-09-09
Slice improvements

### LINKED ISSUE #3932: Variability of slicing
Updated example from: https://github.com/modelica/ModelicaSpecification/issues/2240#issuecomment-3612687891

Note that these examples are a lot more esoteric - and thus it doesn't make sense to add special rules for them:
```modelica
model M
 record A
   constant Real x[:];
 end A;
 parameter Integer n=2;
 parameter A a[n](each x={1,2});
 constant Integer j1=size(a.x, 1); // ?
 constant Integer j2=size(a.x, 2); // ?
 constant Real ax[:,:]=a.x;

  record B
    Real x[:];
  end B;
  B b[n](x=[(1:n)*time]);
  parameter Integer j3=size(b.x,2); //?
end M;
```
The important questions are:
- What is the variability of `a.x`? It seems that since `a` is declared with a parameter size then a slice `a.x` should also have parameter variability, right? And thus `ax` as well as `j1` and `j2` should be illegal, right?
- For `size` there is an exception for variables in models, but these are slices - so I assume it does not apply and thus `j3` is also illegal, right?

HansOlsson:
Define variability for slicing as discussed.
Also add "slice" to index, as I realized it was missing when trying to reference to it.

Closes #3932


chapters/operatorsandexpressions.tex
```diff
@@ -1555,6 +1555,9 @@ \section{Variability of Expressions}\label{variability-of-expressions}
   When determining whether an equation can contribute to solving for a variable \lstinline!v! (for instance, when applying the perfect matching rule, see \cref{synchronous-data-flow-principle-and-single-assignment-rule}), the equation can only be considered contributing if the resulting solution would be at most as variable as \lstinline!v!.
 \end{itemize}
 
+The variability of \lstinline!a[I].x! is the highest variability of \lstinline!I! and \lstinline!x!, note that \lstinline!x! cannot higher variability than \lstinline!a!, due to \cref{variability-of-structured-entities}.
```
perost: Missing a "have" between "cannot higher". Also, "note that ..." seems like it should be the start of a new sentence to me.

HansOlsson: Ok, updating.
```suggestion
The variability of \lstinline!a[I].x! is the highest variability of \lstinline!I! and \lstinline!x!.
Note that \lstinline!x! cannot have higher variability than \lstinline!a!, due to \cref{variability-of-structured-entities}.
```

PR #3938, opened 2026-09-15
Secondary y axis in plots

henrikt-ma:
Triggered by @HansOlsson who suggested that something like this should be standardized in https://github.com/modelica/ModelicaStandardLibrary/pull/4811#issuecomment-5620033185.

Notable design choices of the current proposal:
- The axes are referred to as _left_ and _right_ – not _primary_ and _secondary_.  This gives the figure author a better idea of what the result will look like (it goes without saying on which side of the data window that each y axis should be displayed).
- The `y` and `yRight` form a closed set of y axes to configure.  Alternative designs include allowing `y` to be an array of axes, where each `Axis` gets to specify its `name` and/or `side` (or `placement`).  The alternatives were deemed too complex to be suitable for standardization.
- The `yAxis` only allows the two values `YAxisSelection.Left` and `YAxisSelection.Right`.  A tool which wants to support even more y axes can do so using a vendor annotation for the additional axes specifications, and use `__NameOfVendor_yAxis` instead of `yAxis` to get around the `yAxis` value restriction.
- Default behavior allows a tool to decide whether to use one or two axes, and the presence of a right y axis is triggered by a curve which requests to be associated with it.

Edit: Updated on 2026-09-16 to reflect current state as of ce39a80.


chapters/annotations.tex
```diff
@@ -278,10 +278,14 @@ \subsubsection{Figure and Plot Properties}\label{figure-plot-properties}
   Curve[:] curves "Plot curves";
   /*literal*/ constant Axis x "X axis properties";
   /*literal*/ constant Axis y "Y axis properties";
+  /*literal*/ constant Axis yRight "Right y axis properties";
 end Plot;
 \end{lstlisting}
 
-A \lstinline!Plot! can contain several \lstinline!curves!, see \cref{plot-curves}, that all share a common \lstinline!x! and \lstinline!y! axis with properties described in \cref{axis-properties}.
+A \lstinline!Plot! can contain several \lstinline!curves!, see \cref{plot-curves}, that all share a common \lstinline!x! axis with properties described in \cref{axis-properties}.
+There can be either one or two $y$ axes, with the proerties of the left and right axes defined by \lstinline!y! and \lstinline!yRight!, respectively.
```
```suggestion
There can be either one or two $y$ axes, with the properties of the left and right axes defined by \lstinline!y! and \lstinline!yRight!, respectively.
```

HansOlsson:
I agree that the general idea makes sense, but we should discuss it at the next monthly meeting before merging it.

DagBruck:
That is very much in line with our thoughts on the topic. Two things to consider:
- Maybe left/right axis is too explicit. We could also use the term _alternate axis_, i.e. the axis attribute would be `yAlt` instead of `yRight`.
- You use an enumeration `YAxisSelection`  to select axis. I think just a boolean flag `altY` (default false) is good enough.

chapters/annotations.tex
```diff
@@ -402,6 +406,7 @@ \subsubsection{Plot Curves}\label{plot-curves}
   expression x = time "X coordinate values";
   expression y "Y coordinate values";
   /*literal*/ constant String legend "Legend";
+  /*literal*/ constant YAxisSelection yAxis "Associated y axis";
```
DagBruck, line 409:
```suggestion
  /*literal*/ constant YAxisSelection yAxis=YAxisSelection.Left "Associated y axis";
```

DagBruck (changes requested):
I'm in favor with the minor comment that the default axis should be left.

Also, as commented previously one can consider if "alternate" axis is a more neutral term, and if a Boolean flag is good enough for axis selection (overkill to define an enumeration, espcially if there is a default).

chapters/annotations.tex
henrikt-ma: No, this would prevent a tool from automatically placing two variables on separate axes when their units are not conversion-compatible.  I am not saying that all tools should resolve this situation in that way, but the specification should at least allow it.

henrikt-ma:
> * Maybe left/right axis is too explicit. We could also use the term _alternate axis_, i.e. the axis attribute would be `yAlt` instead of `yRight`.

Initially I also had the idea that it should be less explicit (I called then _primary_ and _secondary_ then), but then I changed my mind when imagining that I as a figure author wouldn't have control over which axis is on the left and which axis is on the right.

With another layer of indirection we could decouple `YAxisSelection` from the choice of side:
```
type YAxisSelection = enumeration(Primary, Secondary);

record Plot
  YAxis y;
  YAxis ySecondary;
  …
end Plot;

YAxisPlacement = enumeration(Left, Right);

record YAxis
  extends Axis;
  YAxisPlacement placement;
end YAxis;
```
However, this would require a lot of explicit settings for a figure author who wants to be in control of the placement.

DagBruck:
> Initially I also had the idea that it should be less explicit (I called then _primary_ and _secondary_ then), but then I changed my mind when imagining that I as a figure author wouldn't have control over which axis is on the left and which axis is on the right.

Yes, that is very reasonable point.

> With another layer of indirection we could decouple `YAxisSelection` from the choice of side:
> However, this would require a lot of explicit settings for a figure author who wants to be in control of the placement.

Agreed. And I argued in favor of even more simplification by having a boolean attribute

`Boolean rightY = false "Use right axis for this curve";`

chapters/annotations.tex
DagBruck: Ok.

DagBruck (approved):

